Issue #280, opened 2026-04-04
[Feature]: Set log level from Python


marph91:
### Problem Statement

Currently, the log is flooded by messages like:

```
[DEBUG] Parsing content stream for text extraction
[TRACE] Detected document script: Latin (sampled 2 characters)
[TRACE] Word boundary detection: 0 boundaries in 2 characters
```

It would be nice to have the possibility to specify the log level or disable logging from Python.

### Proposed Solution

Adding a helper method, like `pdf_oxide.set_log_level("warning")` or `pdf_oxide.disable_logging()`.

### Feature Area

PDF Parsing

### Priority

Nice to have

### Use Case

I'm developing a note converter (https://github.com/marph91/jimmy) and would like to use pdf-oxide for conversion from PDF to Markdown.

### Example Usage (Optional)

```rust

```

### Checklist

- [x] I have searched existing issues to make sure this is not a duplicate
- [x] I have provided a clear description of the feature
- [ ] I am willing to contribute to the implementation (optional)

PR #285, opened 2026-04-05
release/v0.3.21: fix log level in Python (#283) + multi-arch wheels (#284)

### LINKED ISSUE #283: [Bug]: Log level is only partially respected
### Bug Description

This is a follow-up to https://github.com/yfedoseev/pdf_oxide/issues/280. There are still debug messages visible, even if the log level is set to error, like:

```
[DEBUG] Parsing content stream for text extraction
[TRACE] Detected document script: Latin (sampled 2 characters)
[TRACE] Word boundary detection: 0 boundaries in 2 characters
```

### Steps to Reproduce

1. Use https://sample-files.com/downloads/documents/pdf/image-doc.pdf
2. Run the following script:

```python
import logging
from pdf_oxide import PdfDocument

logging.basicConfig(level=logging.ERROR)

doc = PdfDocument("image-doc.pdf")
for page in range(doc.page_count()):
    doc.extract_text(page)
```

### Expected Behavior

Debug messages in the log

### Actual Behavior

Only error messages in the log

_No response_

### Environment

- **OS**: Ubuntu 24.04
- **Rust version**: -
- **pdf_oxide version**: 0.3.20
- **Python version** 3.14.0

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

_No response_

### Checklist

- [ ] I am using the latest version of pdf_oxide

### LINKED ISSUE #284: [Feature]: Add Linux aarch64 wheel and sdist to PyPI releases
### Problem Statement

We’d like to use `pdf_oxide` in our project, but our CI builds Docker images on Linux aarch64 (ARM64) runners. Currently there are only published wheels for:

- manylinux_2_34_x86_64
- macosx_10_12_x86_64
- macosx_11_0_arm64
- win_amd64

### Proposed Solution

- A `manylinux` aarch64 wheel: ARM64 Linux is increasingly common for CI runners (GitHub Actions, AWS Graviton, etc.)
- Publish an sdist (source distribution) so users with a Rust toolchain build from source on any unsupported platform as a fallback

### Alternatives Considered

_No response_

### Feature Area

Other

### Priority

Critical

### Use Case

It's a blocker preventing us from experimenting with this exciting library

### Example Usage (Optional)

```rust

```

### Additional Context

_No response_

### Checklist

- [ ] I am willing to contribute to the implementation (optional)

yfedoseev:
## Summary

- **Fixes #283** — Python log level was only partially respected. Three root causes fixed:
  1. The `extract_log_*!` macros in `src/pipeline/logging.rs` were calling `eprintln!` directly instead of going through the `log` crate, so `pyo3_log` / `logging.basicConfig(level=...)` / `pdf_oxide.set_log_level(...)` never saw them. Messages like `[DEBUG] Parsing content stream for text extraction` and `[TRACE] Detected document script: Latin` leaked through at ERROR level. The macros now forward to `log::debug!` / `log::trace!` / `log::info!` / `log::warn!` / `log::error!` and are properly gated by the log crate's max level filter. Each macro supports two call forms for backward compatibility: a literal-format arm (`extract_log_info!("page {}", n)`) and an expr-only fallback (`extract_log_info!(some_string_var)` → `log::info!("{}", some_string_var)`).
  2. `pyo3_log` caches Python logger effective levels per target on first use, and that cache survived `set_log_level` calls. Once any DEBUG message had been emitted by a target like `pdf_oxide.xref`, lowering the level later only worked for new targets. Fixed by holding the `ResetHandle` returned by `pyo3_log::try_init()` in a `OnceLock` and calling `handle.reset()` from `set_log_level`, `disable_logging`, and `setup_logging`.
  3. Added `get_log_level()` Python function (mirrors `set_log_level`) to allow callers to snapshot and restore the Rust-side log level filter. Exported in `python/pdf_oxide/__init__.py` and `__all__`.
  - Reported by @marph91 as a follow-up to #280.
- **Closes #284** — Rewrote `.github/workflows/python.yml` to use `PyO3/maturin-action@v1`, expanding the wheel matrix:
  - Linux: `manylinux_2_28_x86_64`, `manylinux_2_28_aarch64`, `musllinux_1_2_x86_64`, `musllinux_1_2_aarch64`
  - macOS: `x86_64`, `aarch64`
  - Windows: `x64`, `aarch64` (new)
  - sdist (new)
  - Lowered the manylinux glibc floor from `2_34` to `2_28` (covers RHEL 8, Debian 11, Ubuntu 20.04, Amazon Linux 2023). Reported by @jhhayashi.
- Bumped version to 0.3.21 across Rust lib, CLI, MCP, WASM, and PyPI packages (plus lockfiles).

## Test plan

- [x] `cargo build --features logging --lib` — clean
- [x] `cargo test --lib --features python` — passes (4396 tests)
- [x] `cargo clippy --features python --lib -- -D warnings` — clean
- [x] Reproduced issue #283 exactly with `sample-files.com/downloads/documents/pdf/image-doc.pdf`:
  - [x] `logging.basicConfig(level=logging.ERROR)` → **zero** DEBUG/TRACE output (previously leaked)
  - [x] `logging.basicConfig(level=logging.DEBUG)` → messages flow through Python `logging` module (correct `DEBUG:pdf_oxide.xref:...` format, proving the `pyo3_log` bridge is engaged)
  - [x] `pdf_oxide.set_log_level('error')` → correctly gates at the Rust filter level, including on targets already seen by pyo3_log
- [x] `test_log_level_issue_283_regression` added to `tests/test_python.py`: asserts DEBUG records flow at DEBUG level (sanity check) and no below-ERROR records leak after `set_log_level('error')`; uses `get_log_level()` to restore pre-test state without hard-coding a level
- [ ] CI: verify all new wheel build jobs succeed (manylinux aarch64, musllinux x64/aarch64, win_arm64, sdist)
- [ ] Post-merge: smoke-test an aarch64 wheel in a Graviton / Apple-Silicon Docker container

### DIFF
--- a/python/pdf_oxide/__init__.py
+++ b/python/pdf_oxide/__init__.py
@@ -70,6 +70,7 @@ from .pdf_oxide import (
     # Extraction
     TextSpan,
     disable_logging,
+    get_log_level,
     set_log_level,
     setup_logging,
 )
@@ -105,6 +106,7 @@ __all__ = [
     # Logging
     "setup_logging",
     "set_log_level",
+    "get_log_level",
     "disable_logging",
 ]
 __version__ = VERSION
--- a/src/python.rs
+++ b/src/python.rs
@@ -3510,9 +3510,10 @@ fn init_pyo3_log_handle() {
     PYO3_LOG_RESET_HANDLE.get_or_init(|| {
         pyo3_log::try_init().unwrap_or_else(|_| {
             // Another logger was already installed (e.g. by an embedding
-            // host). Grab a ResetHandle anyway by installing a fresh
-            // Logger configuration — `reset_handle()` is available on any
-            // Logger instance without requiring global installation.
+            // host). In that case, do not replace the global logger;
+            // instead, create a standalone default `Logger` value and take
+            // its `ResetHandle`. `reset_handle()` is available on any
+            // `Logger` instance and does not itself perform installation.
             pyo3_log::Logger::default().reset_handle()
         })
     });
@@ -3586,6 +3587,25 @@ fn disable_logging() {
     reset_pyo3_log_cache();
 }
 
+/// Return the current Rust-side log level filter as a lowercase string.
+///
+/// One of: `"off"`, `"error"`, `"warn"`, `"info"`, `"debug"`, `"trace"`.
+///
+/// This mirrors the gate controlled by `set_log_level` and is useful for
+/// tests or context managers that want to save/restore the log level
+/// around a block without hard-coding a "default".
+#[pyfunction]
+fn get_log_level() -> &'static str {
+    match log::max_level() {
+        log::LevelFilter::Off => "off",
+        log::LevelFilter::Error => "error",
+        log::LevelFilter::Warn => "warn",
+        log::LevelFilter::Info => "info",
+        log::LevelFilter::Debug => "debug",
+        log::LevelFilter::Trace => "trace",
+    }
+}
+
 #[pymodule]
 fn pdf_oxide(m: &Bound<'_, PyModule>) -> PyResult<()> {
     // Bridge Rust `log` to Python `logging` (silent by default, user
@@ -3596,6 +3616,7 @@ fn pdf_oxide(m: &Bound<'_, PyModule>) -> PyResult<()> {
     init_pyo3_log_handle();
     m.add_function(wrap_pyfunction!(setup_logging, m)?)?;
     m.add_function(wrap_pyfunction!(set_log_level, m)?)?;
+    m.add_function(wrap_pyfunction!(get_log_level, m)?)?;
     m.add_function(wrap_pyfunction!(disable_logging, m)?)?;
     m.add_class::<PyPdfDocument>()?;
     m.add_class::<PyPdf>()?;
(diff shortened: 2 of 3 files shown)


yfedoseev:
@copilot — addressed all three comments from the second review in ad8442f:

1. **`tests/test_python.py:1361`** (Python 3.8 compat): You're right that `list[logging.LogRecord]` is 3.9+. The annotation happens to be in function scope (so CPython doesn't evaluate it at runtime in 3.8, which is why CI passed), but it still trips `typing.get_type_hints()` and static checkers. Dropped the annotation entirely — it's a test fixture, not load-bearing.

2. **`tests/test_python.py:1435`** (finally leaks global state): Good catch. Exposed a new `get_log_level()` Python function that mirrors `set_log_level`, added it to `__all__` / `__init__.py`, and used it in the test to snapshot the pre-test Rust log level and restore it exactly in the `finally` block. No more hard-coded "info". Bonus: users now have a symmetric get/set pair.

3. **`src/python.rs:3515`** (misleading comment): Applied your suggested replacement verbatim. The comment now correctly describes what the fallback does (take a `ResetHandle` from a standalone `Logger` without installing it globally).

Local verification: 63/63 Python tests pass, clippy `-D warnings` clean across `python`, `logging`, and default feature configs.

PR #286, opened 2026-04-06
Add Linux aarch64, musl, Windows ARM64 wheels and sdist to Python release

jhhayashi:
The build-python-wheels job was missing several platforms that
build-binaries already supports. This adds:

- Linux ARM64 glibc (aarch64-unknown-linux-gnu)
- Linux x86_64 musl (x86_64-unknown-linux-musl)
- Linux ARM64 musl (aarch64-unknown-linux-musl)
- Windows ARM64 (aarch64-pc-windows-msvc)
- Source distribution (sdist) for platforms without prebuilt wheels

Also adds the necessary cross-compilation tooling (gcc-aarch64-linux-gnu,
musl-tools) and linker configuration for ARM64 Linux builds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,6 +149,7 @@ jobs:
         shell: bash
         env:
           CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: ${{ matrix.target == 'aarch64-unknown-linux-gnu' && 'aarch64-linux-gnu-gcc' || '' }}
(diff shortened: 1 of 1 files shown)


monchin:
@jhhayashi  There's a potential issue that `rylai` doesn't release linux musl and windows arm64, which might lead release workflow on these platforms to crash. If we need to support these platforms, perhaps you need to genarate .pyi file by rylai in other platforms and upload it on git.

And I'll try to publish `rylai` on linux musl and windows arm64 ASAP either

jhhayashi:
Wow, impressive catch @monchin; thanks for the awesome `rylai` lib!

And @yfedoseev thanks for the help getting this across the finish line

jhhayashi:
Are you able to generate the wheels for the existing 0.3.21 version or would this require a new version release?

PR #369, opened 2026-04-16
fix: bound all internal caches + fix path extraction OOM on chart-heavy PDFs

andrewjradcliffe:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Fixes #
#354 

## Changes Made

The object_cache, 5 font caches, xobject_spans_cache, form_xobject_images_cache, and global CMAP_CACHE all grew without bound during multi-page extraction — particularly severe on TeX-produced PDFs with per-page font subsets. All caches now use FIFO eviction:

- object_cache: 64 MB byte-size cap (BoundedObjectCache)
- font caches: 256–512 entry caps (BoundedEntryCache)
- xobject span/image caches: 1024 entry caps
- global CMAP_CACHE: 1024 entry cap (reuses BoundedEntryCache)

PathExtractor::process_form_xobject_paths used stack-based cycle detection but no global dedup, causing combinatorial explosion on pages with shared nested Form XObjects (e.g., chart/plot pages). A 609 KB arXiv PDF consumed 200 MB/s during extract_paths on page 11. Added processed_xobjects HashSet (matching TextExtractor's existing pattern) to skip already-processed XObjects.

Also:
- MutexExt trait with lock_or_recover() replaces 72 .lock().unwrap() calls across document.rs and cmap.rs for poison resistance
- Python bindings: .unwrap() → .ok_or_else(PyRuntimeError) on 9 sites
- document_editor: log::warn on silent .ok() during save
- Markdown converter: inter-group spacing for style transitions
- Exposed clear_cmap_cache() / cmap_cache_size() public API

## Testing

- [x] I have run `cargo fmt`

- [ ] Python tests pass

## Documentation

## Checklist

### DIFF
--- a/src/converters/markdown.rs
+++ b/src/converters/markdown.rs
@@ -74,6 +74,24 @@ lazy_static! {
 )]
 pub struct MarkdownConverter;
 
+/// Returns true if a geometric gap between two adjacent blocks indicates
+/// a word boundary, meaning a space should be inserted between them.
+fn needs_inter_block_space(
+    prev: &crate::layout::TextBlock,
+    next: &crate::layout::TextBlock,
+    spacing_config: &SpacingConfig,
+) -> bool {
+    let gap = next.bbox.left() - prev.bbox.right();
+    let char_size = prev.bbox.width.max(prev.bbox.height);
+    let threshold = spacing_config.word_margin * char_size;
+    if gap <= threshold {
+        return false;
+    }
+    let prev_ends_space = prev.text.chars().last().is_some_and(|c| c.is_whitespace());
+    let next_starts_space = next.text.chars().next().is_some_and(|c| c.is_whitespace());
+    !prev_ends_space && !next_starts_space
+}
+
 #[allow(deprecated)]
 impl MarkdownConverter {
     /// Create a new Markdown converter.
@@ -389,6 +407,7 @@ impl MarkdownConverter {
             //
             // Group consecutive blocks with same bold/italic status to avoid splitting
             // natural phrases like "Chinese stock market" into "**Chinese stock** market"
+            let spacing_config = SpacingConfig::default();
             let mut i = 0;
             while i < line_indices.len() {
                 let idx = line_indices[i];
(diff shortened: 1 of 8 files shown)


yfedoseev:
Thanks @andrewjradcliffe — this is a real fix and the problem classes are well-targeted. Three observations with spec / prior-art grounding:

**Why this bug exists at all.**
- ISO 32000-1:2008 §9.6.4 mandates unique 6-letter tags per font subset, so per-page subsetting (common on typeset / academic PDFs) legitimately creates N × font_count distinct Font objects. Unbounded caches will accumulate indefinitely on those inputs.
- §8.10.1 explicitly anticipates reader-side caching of Form XObject rendering results — which is why the `xobject_spans_cache` / `form_xobject_images_cache` exist in the first place — but correctness depends on the graphics state at invocation (the CTM-key fix that shipped in v0.3.31).
- §8.10 doesn't bound nested Form XObject depth or forbid shared children via multiple parents. The cross-reference DAG you're deduplicating with `processed_xobjects` is spec-compliant and must be handled.

**Things I'd change before merge:**

1. **Rebase on current main.** v0.3.31 just merged, touching `paths.rs` (commit `1ca8834`, Form XObject resource-scope isolation) and `document.rs` (B1 MediaBox post-filter, B3 first-seen-any tracking, B4 XY-cut routing). The rebase will have conflicts in both files — please confirm `processed_xobjects` is additive to the scope-swap fix, not redundant.

2. **LRU on the font caches, not FIFO.** `object_cache` is correctly FIFO — its access pattern is insert-once-read-once along page traversal, so recency isn't a useful signal. But the font caches exist specifically to serve cross-page reuse: a 1000-page document with 5 common fonts + per-page subsets would see FIFO evict the 5 common fonts first (inserted earliest), then thrash. Move-to-back in `BoundedEntryCache::get` gives LRU cheaply.

3. **Explain the test-expectation changes.** `tests/test_markdown_extraction_quality.rs` has 12+/12- — is that all from the style-group spacing fix in section D, or something else? Please call it out in the PR body so reviewers can verify each assertion flip is intentional.

**Smaller, non-blocking:**
- `BoundedEntryCache::insert` over-evicts when re-inserting an existing key (evicts an unrelated entry before noticing the key is a duplicate). Minor cache-efficiency issue.
- `BoundedObjectCache::estimate_size` doesn't count HashMap bucket overhead / Arc headers / allocator padding. Real memory can be 1.5–2× the cap — worth documenting as a guardrail rather than a hard ceiling.
- `BoundedObjectCache` and `BoundedEntryCache` could be unified into one generic `BoundedCache<K, V, M>` with a capacity-measure trait. Cleanup, not blocking.

**What's solid and should ship as-is:** the path-extraction `processed_xobjects` dedup (closes a legitimate gap in the DAG-traversal area the parsing community has documented for years), the `MutexExt::lock_or_recover` pattern (poisoning recovery for non-invariant caches is the right default), and the Python `.unwrap()` → `ok_or_else(PyRuntimeError)` conversion (Python bindings should never panic).

andrewjradcliffe:
Thanks for the quick review @yfedoseev. Interesting background on the root cause.

Addressed all the comments except:
> `BoundedObjectCache` and `BoundedEntryCache` could be unified into one generic `BoundedCache<K, V, M>` with a capacity-measure trait. Cleanup, not blocking.

My reasoning was: the two structs are local, small, and their differences are immediately visible. The trait would be used by just two types in one file. Hence, seems like a premature abstraction, but I can impl it if you think it's worthwhile.

yfedoseev:
Closing manually — the changes from this PR were committed directly to main and shipped in v0.3.33. GitHub couldn't auto-close it because the head branch was set to `main` instead of a feature branch. Thank you @andrewjradcliffe!

Issue #211, opened 2026-03-05
[Bug]: Wrong output and inconsistencies in `extract_words` and `extract_text_lines` output


ankursri494:
### Bug Description

The `extract_words` and `extract_text_lines` output are not correct.

### Steps to Reproduce

Run the test code on the PDFs referenced and inspect the result. PdfPlumber's output is included in the code, for easy comparison of the expected result.

### Expected Behavior

Words and text-lines should be extracted properly in correct reading order for the attached PDFs: **Top to Bottom, Right to Left**, analogous to how `PdfPlumber` returns it.

### Actual Behavior

1. Pdf **pdf_structure.pdf**: Words and text-lines are extracted starting from the bottom of the page, and the word order within lines is incorrect.
3. Pdf **2023-06-20-PV.pdf**:  Both functions return some garbled text. Not all words / lines are getting extracted.
4. Pdf **150109DSP-Milw-505-90D.pdf**:  Both functions again return garbled text output.

### PDF Information (if applicable)

[pdf_structure.pdf](https://github.com/user-attachments/files/25760871/pdf_structure.pdf)
[2023-06-20-PV.pdf](https://github.com/user-attachments/files/25760872/2023-06-20-PV.pdf)
[150109DSP-Milw-505-90D.pdf](https://github.com/user-attachments/files/25760873/150109DSP-Milw-505-90D.pdf)

### Environment

- **OS**: (Ubuntu 22.04)
- **Python version** (if using Python bindings): (3.12)
- **pdf_oxide version**: (0.3.14)
- **pdfplumber version**: (0.11.7)

### Code Sample

```rust
import pdfplumber
import pdf_oxide

pdf1 = pdfplumber.open("<pdf-path>")
pdf2 = pdf_oxide.PdfDocument("<pdf-path>")

print([i["text"] for i in pdf1.pages[0].extract_text_lines()])
print([i.text for i in pdf2.extract_text_lines(0)])
```

### Error Messages / Logs

```shell

```

### Additional Context

The PDF samples attached are picked up from PdfPlumber's test cases itself:  https://github.com/jsvine/pdfplumber/tree/stable/tests/pdfs

### Checklist

- [x] I have searched existing issues to make sure this is not a duplicate
- [x] I have provided all required information above
- [x] I am using the latest version of pdf_oxide

yfedoseev:
@ankursri494 you were right to flag this — closing without confirming the fix actually held was a mistake.

[...]

PR #413, opened 2026-04-26
fix(layout): restore left-to-right order for same-line spans after row-aware sort

RolandWArnold:
`row_aware_span_cmp` sorts spans by quantized Y band (descending) to keep jittery rows stable. For mixed-baseline glyphs (e.g. superscripts or subscripts), this places higher-Y spans ahead of their base, producing outputs like `"th8"` instead of `"8th"`.

Add `PdfDocument::reorder_same_line_runs(&mut [TextSpan])` and invoke it after the row-aware sort + `reorder_rowspan_labels` pass. The helper finds maximal same-line runs using `same_line_threshold` and re-sorts each run by X (with `sequence` as a stable tie-break).

This preserves the existing sort semantics and only reorders spans within a visual line; no spans are merged or dropped.

Adds regression coverage for:
- superscript ordering (`"8th"`)
- subscript between baseline letters (`"H2O"` with spacing to avoid span merge)
- multi-span same-line runs across distinct Y bands (`"abc"`)

Existing same-line and multi-line behaviour unchanged.

See [superscript.pdf](https://github.com/user-attachments/files/26949823/superscript.pdf):

- Before: `th8`
- After: `8th`

This correction of intra-line ordering is the future fix noted in #394's description.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [x] Tests
- [ ] CI/CD changes

## Related Issues

Relates to #456

## Changes Made

- Added `PdfDocument::reorder_same_line_runs`, which finds maximal same-line runs after row-aware sorting and re-sorts each run by X position, using `sequence` as a stable tie-break.
- Invoked `reorder_same_line_runs` after the existing row-aware sort and `reorder_rowspan_labels` pass, so the change is limited to intra-line ordering after the current row/rowspan ordering logic has run.
- Added regression tests covering superscript ordering (`8th`), subscript ordering (`H2O`), multi-span same-line runs across distinct Y bands (`abc`).

## Testing

- [x] I ran the targeted regression tests for this fix locally and they pass.
- [x] I have run `cargo fmt`

Tests run locally

- [x] `cargo test --test test_superscript_line_grouping`

- [ ] Python tests pass

## Documentation

## Checklist

image

Before: `th8`  
After: `8th`

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -120,6 +120,16 @@ const DEFAULT_XOBJECT_CACHE_MAX_ENTRIES: usize = 1024;
 /// this pairwise heuristic.
 const FORWARD_GAP_K: f32 = 1.25;
 
+/// Fraction of the larger font size used as the Y tolerance for treating
+/// two spans as belonging to the same visual line.
+const SAME_LINE_THRESHOLD_FONT_SIZE_FACTOR: f32 = 0.5;
+
+/// Maximum allowed inter-span X gap inside a candidate same-line reorder run.
+/// If the candidate's tentative X-order contains a larger gap, the run is
+/// probably a disjoint footer/header/field layout rather than a local
+/// mixed-baseline repair.
+const SAME_LINE_REORDER_MAX_GAP_FACTOR: f32 = 3.0;
+
 // Re-export BoundedEntryCache from cache module for local use and backward compatibility
 pub(crate) use crate::cache::BoundedEntryCache;
 
@@ -4265,6 +4275,11 @@ impl PdfDocument {
                 // vertically centred across several dense-column data rows)
                 // to sort at the top of their row block.
                 Self::reorder_rowspan_labels(&mut spans);
+
+                // Restore intra-line reading order after the row-aware band sort.
+                // Off-baseline glyphs (e.g. superscripts/subscripts) can land in
+                // adjacent bands and be emitted out of X order; fix that per line.
+                Self::reorder_same_line_runs(&mut spans);
             }
 
             // OCR fallback for scanned PDFs
@@ -5268,7 +5283,7 @@ impl PdfDocument {
     /// (for example superscripts and subscripts) are not split by a fixed
     /// absolute tolerance.
     fn same_line_threshold(prev: &TextSpan, current: &TextSpan) -> f32 {
-        prev.font_size.max(current.font_size).max(1.0) * 0.5
+        prev.font_size.max(current.font_size).max(1.0) * SAME_LINE_THRESHOLD_FONT_SIZE_FACTOR
     }
 
     /// Returns `true` if `inner` is contained within `outer`,
@@ -5285,6 +5300,104 @@ impl PdfDocument {
             && inner.right() <= outer.right() + eps
             && inner.top() >= outer.top() - eps
             && inner.bottom() <= outer.bottom() + eps
+    }
+
+    /// Returns `true` if a tentative left-to-right X-ordering of `run`
+    /// contains a horizontal gap exceeding
+    /// `SAME_LINE_REORDER_MAX_GAP_FACTOR * max(font_size)` between any
+    /// two consecutive spans. Used by [`reorder_same_line_runs`] to
+    /// reject candidate runs that are vertically close but horizontally
+    /// disjoint (e.g. tightly-set footer/header rows split across the
+    /// page).
+    ///
+    /// The slice is not mutated; the X-order is computed on a local
+    /// copy of `(left_x, right_x, font_size)` triples.
+    fn run_has_large_x_gap(run: &[TextSpan]) -> bool {
+        if run.len() < 2 {
+            return false;
+        }
+
+        let mut edges: Vec<(f32, f32, f32)> = run
+            .iter()
+            .map(|s| (s.bbox.x, s.bbox.x + s.bbox.width, s.font_size))
+            .collect();
+
+        edges.sort_by(|a, b| crate::utils::safe_float_cmp(a.0, b.0));
+
+        for pair in edges.windows(2) {
+            let prev = pair[0];
+            let cur = pair[1];
+
+            let gap = cur.0 - prev.1;
+            if gap <= 0.0 {
+                continue;
+            }
+
+            let max_fs = prev.2.max(cur.2).max(1.0);
+            if gap > SAME_LINE_REORDER_MAX_GAP_FACTOR * max_fs {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    /// Re-sort same-line spans by X after row-aware band sorting.
+    ///
+    /// Row-aware sorting can place off-baseline glyphs such as superscripts or
+    /// subscripts in adjacent Y bands before their base glyphs. This helper finds
+    /// candidate runs with the existing same-line threshold, then tentatively views
+    /// each candidate in X order. If that tentative X order contains a large gap,
+    /// the candidate is treated as disjoint footer/header/field content and is
+    /// left in the existing row-aware order.
+    ///
+    /// At the slice level no spans are merged or dropped; successful candidates are
+    /// only permuted. Downstream text assembly may then emit the reordered spans
+    /// into one visual line, which is the user-observable effect.
+    fn reorder_same_line_runs(spans: &mut [TextSpan]) {
+        let mut i = 0;
+
+        while i < spans.len() {
+            let mut j = i + 1;
+
+            while j < spans.len() {
+                let anchor = &spans[i];
+                let prev = &spans[j - 1];
+                let cur = &spans[j];
+
+                let to_prev = (cur.bbox.y - prev.bbox.y).abs();
+                let to_anchor = (cur.bbox.y - anchor.bbox.y).abs();
+
+                let tol_prev = Self::same_line_threshold(prev, cur);
+                let tol_anchor = Self::same_line_threshold(anchor, cur);
+
+                if to_prev > tol_prev || to_anchor > tol_anchor {
+                    break;
+                }
+
+                j += 1;
+            }
+
+            if j - i > 1 {
+                if Self::run_has_large_x_gap(&spans[i..j]) {
+                    // Candidate spans are vertically close, but not horizontally
+                    // contiguous. Do not X-sort them into a fake line; preserve
+                    // the row-aware order established before this helper.
+                    i = j;
+                    continue;
+                }
+
+                spans[i..j].sort_by(|a, b| {
+                    let cmp = crate::utils::safe_float_cmp(a.bbox.x, b.bbox.x);
+                    if cmp != std::cmp::Ordering::Equal {
+                        return cmp;
+                    }
+                    a.sequence.cmp(&b.sequence)
+                });
+            }
+
+            i = j;
+        }
     }
 
     /// # Returns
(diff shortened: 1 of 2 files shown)


RolandWArnold:
Rebased onto current main and marked ready for review. The earlier bot review was against the pre-rebase state. Thanks!

yfedoseev (changes requested):
Tested locally cherry-picked onto `release/v0.3.43` (which has the #457 reading-order rewire merged) and ran the conversion harness against an 86-PDF stratified corpus (academic, mixed, forms, government, newspapers, theses, plus the three #211 fixtures). 888 sampled `(pdf, page, method)` comparisons against the 0.3.42 baseline.

## What works

The fix delivers on the superscript / subscript case it's targeting. Verified on the academic fixtures: math expressions with sub/super-scripts now extract with subscripts attached to their base symbols. Concrete example from arxiv 1712.09005 page 0:

```
0.3.42:  'exp (−kxi − xjk /2σi )'
         'exp (−kx − x k /2σ )'      ← jittery: subscripts dropped from one occurrence
PR #413: 'i − xjk /2σi )'
         'k6=i exp (−kxi − xkk /2σi )'  ← all subscripts attached
```

So the directional intent (`reorder_same_line_runs` to absorb superscript jitter) is correct and the regression test for `"8th"` covers the bug.

## What needs to change

The same-line threshold is too loose. With `same_line_threshold(prev, current) = max(prev.font_size, current.font_size, 1.0) * 0.5`, the helper groups rows whose y-delta is up to half a font-size apart. That's exactly the inter-line spacing of tightly-set running headers/footers, so they get pulled into a single "line" and concatenated.

Concrete regression — CFR Title 7 page 0 footer:

```
0.3.42:  'VerDate Sep<11>2014 09:05 Feb 23, 2024 Jkt 262012 PO 00000 Frm 00001 ...'
         'skersey on DSK4WB1RN3PROD with CFR'                       ← two distinct lines

PR #413: 'skersey on DSK4WB1RN3PROD with CFRVerDate Sep<11>2014 09:05 Feb 23, 2024 ...'
                                              ↑ glued together, no whitespace
```

This pattern recurs across the corpus: of the 61 PDFs whose output changed under PR #413, **58 (95%) show a net line-count *decrease*** — i.e., the helper is merging lines that the baseline (correctly) keeps separate. Only 1 PDF gets a useful split. So the change is net-negative on prose-heavy / form / government documents while fixing the academic-math subset.

The 6 CFR titles in the corpus all show this exact `CFRVerDate` glue. IRS forms, Berkeley theses, and a couple of arxiv preprints show similar adjacent-line concatenations.

## Suggested directions

Two non-mutually-exclusive angles. Either alone should resolve the regression while keeping the math fix:

1. **Tighten the same-line threshold for the cross-pair check.** `0.5 * font_size` is fine for grouping characters within a run, but the per-pair distance check inside `reorder_same_line_runs` should be much stricter — closer to `0.25 * font_size` or `≤ font_size * 0.15`. Math superscripts sit ~30–40% of font size above the baseline, so 0.4 is the practical ceiling; below that picks up the sub/super case but excludes adjacent-row merges.

2. **Require x-overlap or x-adjacency between same-line candidates.** Superscripts share the parent token's x-range (they overhang it). Adjacent text rows from a running footer typically don't share x-ranges with the metadata row above (different x-origins, different widths). A check like "`cur.bbox.x` and `anchor.bbox.right` differ by no more than `2 * font_size`" would keep super/subscript runs together while rejecting cross-line glue.

Either approach should be measurable against the same 86-PDF corpus by re-running the conversion harness — a clean change keeps the academic math improvements while collapsing the line-merge count from 58 back toward 0.

The regression-test additions in `tests/test_superscript_line_grouping.rs` are good but only cover the positive direction. Adding a negative case along the lines of "two text lines at y-delta ~0.45 × font_size with disjoint x-ranges must remain separate after `reorder_same_line_runs`" would have caught the CFR pattern locally.

## Code-review notes (independent of the regression)

- Visibility: `pub(crate) fn reorder_same_line_runs` is fine; for symmetry with `reorder_rowspan_labels` the fn could just be `fn` (private to the impl block) since no external module needs to call it.
- The doc comment on the public-ish helper claims "no spans are merged or dropped." That's accurate at the slice level but the sentence is worth qualifying — *visible output* spans get re-ordered into a single emitted line by downstream `extract_text_with_options`, which is the user-observable effect.

Net: the fix is real and worth landing, but the threshold needs to be stricter (or x-overlap-gated) before it can ship. Happy to re-run the harness once it's revised.

RolandWArnold:
Thanks for the detailed corpus run and diagnosis, that was very helpful, and I've revised the PR using the x-adjacency direction rather than making the Y threshold narrower.

The helper still greedily finds Y-near candidate runs with the existing `same_line_threshold`, but it doesn't unconditionally X-sort those candidates anymore: before mutating the slice, it now tentatively views the candidate in left-to-right X order and rejects the reorder if that X-order contains a large horizontal gap (`> 3.0 * max(font_size)` for the adjacent pair) (in that case the candidate is left in the existing row-aware order, so disjoint footer/header rows do not get sorted into a fake same-line sequence).

This aims to catch the CFR-style case you reported, where things are vertically close enough to be picked up by the Y-greedy candidate step, but sufficiently horizontally disjoint that it shouldn't be treated as one local mixed-baseline run.

I kept `same_line_threshold` itself unchanged because it is used outside this helper as part of the broader same-line/spacing behavior. The new validation is local to `reorder_same_line_runs`.

I also addressed the review notes:

- changed `reorder_same_line_runs` from `pub(crate) fn` to private `fn`
- qualified the doc comment so “no spans are merged or dropped” is explicitly slice-level; downstream assembly is still the user-visible effect
- added a direct negative unit test, `reorder_same_line_runs_preserves_disjoint_x_rows`, with two rows at `Δy = 4.5` and `fs = 10`, so they pass the Y threshold but are rejected by the X-gap validator
- added a defensive positive unit test, `reorder_same_line_runs_orders_suffix_superscript_by_x`, to make sure the validation does not recreate the “superscript as rightmost anchor” problem
- kept the existing positive integration tests for `8th`, `H2O`, and the three-glyph distinct-band run unchanged

yfedoseev:
Thanks for the revised version, Roland. Re-ran the conversion harness against a 90-PDF stratified corpus (academic / mixed / forms / government / newspapers / technical / diverse / theses) on top of `release/v0.3.44`. The new x-gap validation does help on a subset of cases, and the math/superscript fix is unambiguously a win for academic content. However, the regression I flagged in February is **still firing** on government PDFs, and I think I can pinpoint why the x-gap guard misses it.

## Numbers

| Outcome | Count | Share |
|---|---|---|
| Unchanged | 51 | 57 % |
| Changed   | 39 | 43 % |

Of the 39 changed:

- **~27 academic / arxiv math papers**: improvement — superscripts and subscripts (`∆∗`, `10³`, footnote markers) now attach to their bases. This is the intended fix, working as advertised.
- **9 of 9 sampled government PDFs**: regression — the same stacked-footer-row gluing I reported in v1.
- **2 forms**: neutral (label rearrangement, no glue).
- **1 mixed-content paper**: mixed (chemistry-formula subscript wins + body-text rearrangement).

## Why the x-gap guard misses the government-PDF case

The current guard scans the candidate run X-sorted and flags the *gap* between consecutive spans. Concretely (from `run_has_large_x_gap`):

```rust
let gap = cur.0 - prev.1;       // next.left - prev.right
if gap <= 0.0 { continue; }     // overlap → "no gap", skip
```

The shape it misses is a **wide span horizontally engulfing a chain of narrow fragments at a slightly different baseline**. Concrete numbers from the synthetic fixture below (extracted via `doc.extract_spans(0)` so this is what `reorder_same_line_runs` sees):

```
fragments (top row, font 6.5pt, baseline Y=20)
   x= 25.00  w=26.72   "Marker-A"
   x= 62.00  w=23.84   "Build-77"
   x=100.00  w=24.93   "Run-019"
   x=135.00  w=25.65   "Tag-XX1"
   ... 5 more, all gap ≤ 7pt apart ...
   x=315.00  w=19.87   "Lot-AB"
wide span (bottom row, font 5pt, baseline Y=18)
   x= 22.00  w=95.59   "operator-N at node-X with profile-Y stage-Z"
```

After X-sorting:

```
left  right    span
22.00 117.59   wide span
25.00  51.72   "Marker-A"  ← left=25 < prev right=117.59, gap = -93  → skip
62.00  85.84   "Build-77"  ← left=62 < prev right=117.59, gap = -56  → skip
100.00 124.93  "Run-019"   ← left=100 < prev right=117.59, gap = -18 → skip
135.00 160.65  "Tag-XX1"   ← left=135 > prev right=124.93, gap = 10 → ok, < 3*6.5
... all subsequent gaps are small ...
```

So the loop sees only negative-or-small gaps and concludes "horizontally contiguous" — even though the wide span and the fragments are clearly two separate logical lines that just happen to overlap in X. The reorder fires, X-sorting puts the wide span first (leftmost X = 22), and the output glues them: `operator-N at node-X with profile-Y stage-Z<fragment 1><fragment 2>…`.

In words: **the guard distinguishes "rows separated by a horizontal gutter" from "tightly-set mixed-baseline runs," but it doesn't distinguish "rows that happen to overlap in X" from the latter**. Real-world running headers + page-margin metadata commonly fall into the overlapping case.

## Synthetic fixture (available on request)

I have a minimal **`fixture_a_overlap_x_range.pdf`** (~2.8 KB, all synthetic content — no real-document text), happy to share it separately. It's a single page with:

- Four short body lines at the top.
- A row of 9 tightly-spaced short fragments at Y = 772, font 6.5pt.
- A single wide span at Y = 774 (Δy = 2pt, below the 0.5 × max_fs = 3.25pt threshold), x = 22, w ≈ 117 — engulfing the first four fragments.

Comparison:

```
── BASELINE (release/v0.3.44 without #413) ──
Marker-A Build-77 Run-019 Tag-XX1 Set-001 Profile-AA Stream-7 ID-9999 Lot-AB
operator-N at node-X with profile-Y stage-Z

── CANDIDATE (#413) ──
operator-N at node-X with profile-Y stage-ZMarker-A Build-77 Run-019 Tag-XX1 Set-001 Profile-AA Stream-7 ID-9999 Lot-AB
```

Same shape as the v1 regression I described in my earlier review. The fixture is small enough to land alongside `reorder_same_line_runs_preserves_disjoint_x_rows` as a regression test once the algorithm is updated.

## Suggested fix directions

Either alone (or both) should resolve the overlap case while keeping the math fix:

1. **Replace gap-check with a containment check.** If any candidate span's X range is *contained* within (or substantially overlaps) another candidate span's X range, the candidate is "stacked rows that happen to overlap," not "mixed-baseline runs." Mixed-baseline glyphs (super/subscripts) sit *above or below* their base — they don't engulf neighbors horizontally. So a check like "for every adjacent pair in X-sorted order, the smaller of (left/right) overlap fractions is below some threshold" rejects engulfing while keeping superscript-of-token cases.

2. **Or add an absolute Y-delta floor in addition to the existing relative threshold.** Stacked footer rows in real documents have Δy ≥ ~1pt absolute even when the relative ratio is small (because their fonts are tiny). A condition like "Δy > 1.5pt absolute AND no candidate span's X bbox is contained in another's" would split this case from a true superscript (which has Δy proportional to its font size, but its X bbox sits inside its base's X bbox).

3. **Or require that the candidate run's X bboxes form a strictly increasing nested-or-disjoint sequence after sort.** Mixed-baseline runs satisfy this (every glyph's X bbox is to the right of the previous, with possibly small overlaps for kerning); engulfed-row runs do not.

Happy to re-run the 90-PDF harness against any further revision.

## Bottom line

The math-extraction fix is real and we want it. But on this corpus the change is still net-mixed (~27 wins on academic vs ~9 regressions on government), so I don't think it should land on `release/v0.3.44` as-is. Once the engulfing-overlap case is handled, I'd be glad to merge.

RolandWArnold:
Thanks for taking another look. Could you send `fixture_a_overlap_x_range.pdf` when you get a chance? My commit email on this branch works for direct send, or a download link / attachment in a follow-up comment is also fine, whichever is easiest.

Realistically I’ll only be able to dig into this over the weekend, but I’ll come back with a revised version once I’ve tested against the fixture.

RolandWArnold:
Hi @yfedoseev - could you send  the `fixture_a_overlap_x_range.pdf` fixture you mentioned when you have a moment? My email, an attachment here, or a download link would all be fine. Thanks!

yfedoseev:
@RolandWArnold I already merged your changes into v0.3.46 and it will be launched in several hours. I think it's all good. Thank you so much 🙏

RolandWArnold:
Thanks @yfedoseev, I’m glad it was helpful. I’m fairly tied up at the moment, so I don’t want to promise any quick follow-ups, but the layout and superscript handling is something I’d like to dig into more when I have a clear window.

yfedoseev:
Closing — changes from this branch were included in the v0.3.46 release bundle (PR #491, merged 2026-05-11). The fix is already in main.

PR #452, opened 2026-04-30
fix(cli): require -o for merge; route remaining default outputs beside input (#412)

### LINKED ISSUE #412: watermark: no -o flag silently writes to cwd instead of failing or writing alongside input
## Bug / UX Issue

When `pdf-oxide watermark` is called without `-o`, it silently creates a file in the **current working directory** named `<original-basename>_watermarked.pdf`, regardless of where the input PDF lives.

## Steps to reproduce

```bash
# From /tmp, watermark a PDF that lives elsewhere
cd /tmp
pdf-oxide watermark /some/other/dir/document.pdf "DRAFT"
# Output: Added watermark 'DRAFT' to N page(s) → document_watermarked.pdf
# File is created at /tmp/document_watermarked.pdf  ← surprising
```

## Expected behaviour

One of:
1. **Require `-o`** — error out with a clear message when no output path is given, since `watermark` produces a binary PDF (unlike `text`/`markdown`/`html` which sensibly default to stdout).
2. **Write next to the input file** — output `<input-dir>/document_watermarked.pdf` so the result lands beside the source.
3. **Document it explicitly** — at minimum, the `--help` text should say "defaults to `<basename>_watermarked.pdf` in the current directory" rather than the generic "defaults to stdout for text outputs" (which is misleading for binary commands).

## Actual behaviour

File is created in cwd silently. In a pipeline running from an unexpected cwd this either leaves stray files or fails with a cryptic `IO error: No such file or directory` if cwd is read-only.

## Context

Discovered during v0.3.39 regression testing. The `compress`, `merge`, `rotate`, `split`, and `flatten` commands may share the same pattern — worth auditing all binary-output commands for consistency.

## Environment

- `pdf-oxide 0.3.39`
- Linux

truffle-dev:
Closes #412.

Picks up where 9dd94c0 left off. The earlier commit landed `output_beside()` / `output_dir_beside()` and converted watermark, compress, flatten, rotate, and split. The remaining four binary-output commands kept their inline `Path::new(&format!(\"{stem}_X.pdf\"))` blocks, which resolve relative to cwd, so e.g. `pdf-oxide decrypt /docs/foo.pdf` from `/tmp` still wrote `/tmp/foo_decrypted.pdf` instead of beside the input.

### Pattern-A conversions
- `decrypt.rs` → `output_beside(file, \"_decrypted.pdf\")`
- `crop.rs` → `output_beside(file, \"_cropped.pdf\")`
- `delete.rs` → `output_beside(file, \"_trimmed.pdf\")`
- `reorder.rs` → `output_beside(file, \"_reordered.pdf\")`

No behavior change vs cwd when the input is a bare filename — `output_beside` already returns `./<stem>_X.pdf` for that case (covered by `output_beside_bare_filename`).

### Merge requires `-o`
Per the issue thread, `merge` has no single input anchor — the previous default `<first-arg-dir>/merged.pdf` could silently overwrite an unrelated file in the first input's directory. New behavior: error up front with a message that names the flag.

```
$ pdf-oxide merge a.pdf b.pdf
Error: Invalid operation: Merge requires -o/--output to specify the destination path (e.g. -o merged.pdf). There is no single input file to anchor a default output to.
```

The validation runs before opening any file so the error reports without partially constructing an editor.

### Split

Already correct after 9dd94c0 (defaults to `output_dir_beside(file)`); no change needed.

### Verification

End-to-end run from `/home/phantom` against an input in `/tmp`:

```
$ pdf-oxide decrypt /tmp/in.pdf --password ''
Decrypted /tmp/in.pdf -> /tmp/in_decrypted.pdf
$ pdf-oxide crop /tmp/in.pdf --margins '10,10,10,10'
Cropped 1 page(s) ... → /tmp/in_cropped.pdf
$ pdf-oxide delete /tmp/in.pdf --pages '1'
Deleted 1 page(s) → /tmp/in_trimmed.pdf
$ pdf-oxide reorder /tmp/in.pdf --order '2,1'
Reordered 2 pages → /tmp/in_reordered.pdf
$ ls /home/phantom/in_*.pdf
ls: cannot access '/home/phantom/in_*.pdf': No such file or directory
```

Stash-bisect on the merge regression test (`run_without_output_returns_invalid_operation`):

- pre-fix code → `expected InvalidOperation, got Io(Os { code: 2, kind: NotFound })` (the test reaches the `DocumentEditor::open(&files[0])` path)
- with fix → `1 passed`

`cargo test -p pdf_oxide_cli` 12/12 pass, `cargo fmt -p pdf_oxide_cli --check` clean, `cargo clippy -p pdf_oxide_cli --all-targets -- -D warnings` clean.

### DIFF
--- a/pdf_oxide_cli/src/cli/commands/crop.rs
+++ b/pdf_oxide_cli/src/cli/commands/crop.rs
@@ -46,13 +46,9 @@ pub fn run(
         editor.set_page_crop_box(idx, crop_box)?;
     }
 
-    let out_path = output.map(|p| p.to_path_buf()).unwrap_or_else(|| {
-        let stem = file
-            .file_stem()
-            .and_then(|s| s.to_str())
-            .unwrap_or("output");
-        Path::new(&format!("{stem}_cropped.pdf")).to_path_buf()
-    });
+    let out_path = output
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| super::output_beside(file, "_cropped.pdf"));
 
     editor.save_with_options(
         &out_path,
--- a/pdf_oxide_cli/src/cli/commands/decrypt.rs
+++ b/pdf_oxide_cli/src/cli/commands/decrypt.rs
@@ -4,13 +4,9 @@ use std::path::Path;
 pub fn run(file: &Path, _password: &str, output: Option<&Path>) -> pdf_oxide::Result<()> {
     let mut editor = DocumentEditor::open(file)?;
 
-    let out_path = output.map(|p| p.to_path_buf()).unwrap_or_else(|| {
-        let stem = file
-            .file_stem()
-            .and_then(|s| s.to_str())
-            .unwrap_or("output");
-        std::path::PathBuf::from(format!("{stem}_decrypted.pdf"))
-    });
+    let out_path = output
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| super::output_beside(file, "_decrypted.pdf"));
 
     editor.save_with_options(
         &out_path,
--- a/pdf_oxide_cli/src/cli/commands/delete.rs
+++ b/pdf_oxide_cli/src/cli/commands/delete.rs
@@ -35,13 +35,9 @@ pub fn run(
         editor.remove_page(idx)?;
     }
 
-    let out_path = output.map(|p| p.to_path_buf()).unwrap_or_else(|| {
-        let stem = file
-            .file_stem()
-            .and_then(|s| s.to_str())
-            .unwrap_or("output");
-        Path::new(&format!("{stem}_trimmed.pdf")).to_path_buf()
-    });
+    let out_path = output
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| super::output_beside(file, "_trimmed.pdf"));
 
     editor.save_with_options(
         &out_path,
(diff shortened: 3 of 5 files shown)


yfedoseev (approved):
Code reviewed — all five changes are correct. The four Pattern-A conversions (crop, decrypt, delete, reorder) are mechanical and uniform: the suffix strings preserve existing naming, `output_beside(file, suffix)` anchors correctly to the input's parent directory, and the helper's edge cases (bare filename, root-level path) are already covered by the existing tests in `mod.rs`. The merge change is the right call — early `ok_or_else()` validation fails before any file I/O, the error message is clear and names both `-o`/`--output`, and the new test correctly exercises the pre-I/O path. No issues found. Ready to merge.

yfedoseev (approved):
Code reviewed — all five changes are correct. The four Pattern-A conversions (crop, decrypt, delete, reorder) are mechanical and uniform: suffix strings preserve existing naming, `output_beside(file, suffix)` anchors correctly to the input's parent directory, and the helper's edge cases are already covered by existing tests in `mod.rs`. The merge change is the right call — early `ok_or_else()` validation fails before any file I/O, the error message names both `-o`/`--output`, and the new test correctly exercises the pre-I/O path. Ready to merge.

yfedoseev:
Changes were cherry-picked into the release/v0.3.41 branch and have been merged into main as part of the v0.3.41 release.

PR #587, opened 2026-05-25
feat(extractors): expose Optional Content Group (PDF layer) name on extracted paths

### LINKED ISSUE #586: [Feature]: expose optional content group (layer) name on extracted paths
## Summary

When `extract_paths()` returns a path, surface the **Optional Content Group** (OCG, a.k.a. "layer name") that the path belongs to. The PDF content stream wraps geometry in `BDC … EMC` markers that reference an OCG with a human-readable name (e.g. `A-GRID`, `S-COLS`, `A-WALL-DIM`); today that information is silently dropped during path extraction — `document.rs::extract_paths` does not match the marked-content operators and they fall through `_ => {}`.

Concretely, add a `layer: Option<String>` field to `PathContent` and expose it on the Python dict alongside `bbox`, `stroke_width`, `operations`, etc.

## Use case

We are building a construction document processing platform (architectural / structural PDF drawings exported from AutoCAD and Revit). Detecting the **gridline system** — P / N / O verticals; 1 / 2 / 3 horizontals, the reference axes engineers anchor every dimension to — is a recurring problem.

Without layer information we have to rely on geometric heuristics: "thin stroke + dashed pattern + axis-aligned + spans ≥ X% of the page". These fail on plans where the gridline strokes are stylistically similar to dimension lines or rebar callouts, and they fail entirely on flattened-then-re-exported PDFs where stroke widths converge.

With the OCG name we get a precise classifier on every PDF that preserves authoring metadata (which is the vast majority of CAD-exported PDFs from Revit, AutoCAD, ArchiCAD, Bentley). The same field unlocks correct downstream classification for walls, columns, MEP ducts, dimension lines, and text annotations — currently each one has its own geometric heuristic that breaks on edge cases.

This continues the same construction-document-processing use case behind PR #261 (path operations exposure) and the upcoming circular-vector-pick / column-detection workflows.

## Proposed shape

```python
paths = doc.extract_paths(0)
for p in paths:
    print(p["layer"], p["bbox"], p["stroke_width"])
    # → "A-GRID"   (0.0, 0.0, 100.0, 0.0)   0.13
    # → "A-GRID"   (200.0, 0.0, 200.0, 500.0)  0.13
    # → "S-COLS"   (...)   (...)   (...)
    # → None       (...)   (...)   (...)    # ← path outside any OCG
```

`Option<String>` so PDFs without `/OCProperties` (or with paths emitted outside any `BDC /OC … EMC` bracket) keep working — the new field is just `None`.

## Where the change lives

PDF spec reference: ISO 32000-1:2008 §8.11 (Optional Content) + §14.6 (Marked Content). The on-the-wire structure is:

```
/OC /Layer_5 BDC      ← begin marked content; "/Layer_5" is a name in /Resources /Properties
  100 100 m
  200 100 l
  S
EMC                    ← end marked content
```

`/Layer_5` indirects through the page's `/Resources /Properties` dict to an indirect reference to an `OptionalContentGroup` object whose `/Name` entry holds the human-readable layer name.

The plumbing is mostly already in place:

- `Operator::BeginMarkedContentDict { tag, properties }` and `Operator::EndMarkedContent` are already parsed in `src/content/operators.rs` (and tests confirm they survive content-stream round-tripping — `test_prescan_forward_ctm_preserves_marked_content`).
- OCG writer support exists in `src/writer/layers.rs` (`Layer`, `LayerBuilder`, `/OCProperties` emission), so the project already models the spec.

What is missing is hooking those operators into the path extractor:

1. `src/elements/path.rs` — add `pub layer: Option<String>` (with `#[serde(default)]` so existing serialized payloads stay valid).
2. `src/extractors/paths.rs` — maintain a `Vec<Option<String>>` stack on `PathExtractor`; when a path is finalized (`stroke`, `fill`, `close_fill_and_stroke`), copy the stack top into the resulting `PathContent`.
3. `src/document.rs::extract_paths` — handle `Operator::BeginMarkedContentDict { tag, properties }`: when `tag == "OC"`, resolve `properties` (inline dict → read `/Name`; resource-name reference → look up in `page_dict["/Resources"]["/Properties"]`, dereference, read `/Name`) and push onto the extractor's layer stack. Push `None` for non-`/OC` BDCs to keep stack depth balanced. On `EndMarkedContent`, pop.
4. `src/python.rs::path_to_py_dict` — three lines to expose the new field on the Python dict.

Estimated diff: ~80–120 LOC additive, plus a fixture-based test in `tests/test_path_extraction.rs` (small PDF with two OCG-wrapped paths and one path outside any OCG) and a Python-side test in `tests/test_python.py`.

## Non-goals (out of scope for this issue)

- Filtering paths by OCG visibility (`/D` default visibility settings, view/print/export intents). The caller can do that — we only need the **name**.
- Resolving nested OCG hierarchies (`/OCGs` group membership, `/Order` arrays). Just the immediate OCG name on each path.
- XObject-level OCG inheritance. If the spec turns out to require that for some real-world PDF, we can add it as a follow-up.

## Happy to send the PR

Same pattern as PR #261 (path operations exposure) — additive change, no breaking impact on existing consumers. I will open the PR shortly with the patch + tests.

willywg:
## Description

Adds a `layer: Option<String>` field to `PathContent` (and the dict returned by `extract_paths()` Python binding) that surfaces the **Optional Content Group** name — a.k.a. the PDF "layer name" — the path belongs to.

Today the marked-content operators `BDC` / `BMC` / `EMC` are silently dropped during path extraction (`document.rs::extract_paths` does not match them and they fall through `_ => {}`), so the layer membership written by the source CAD application is unrecoverable from the Python API even though the raw bytes are still in the content stream.

**Use case:** construction document processing — architectural and structural PDFs exported from AutoCAD / Revit preserve layer names like `A-GRID`, `S-COLS`, `A-WALL-DIM`, `M-DUCT`. Surfacing those names lets consumers classify gridlines / columns / dimensions / MEP geometry with the source-of-truth metadata instead of brittle stroke-width-and-dash heuristics. Same pattern as PR #261 (path operations exposure).

Closes #586.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [x] Tests
- [ ] CI/CD changes

## Related Issues

Closes #586 (request + design + non-goals).

Continues the same construction-document-processing motivation as PR #261 — that PR exposed path operations, this one exposes the layer the path lives on.

## Changes Made

8 files, 651 insertions:

1. **`src/elements/path.rs`** — `PathContent` gains `layer: Option<String>` with `#[serde(default)]` for backward compat. New `with_layer()` builder.
2. **`src/extractors/paths.rs`** — `PathExtractor` maintains a `Vec<Option<String>>` stack via `push_oc_layer()` / `pop_oc_layer()`. The stack stays depth-balanced for non-`/OC` `BDC` / `BMC` (Span, Article, …) by pushing `None` for those tags, so an outer `/OC` region continues to apply to nested non-OC content. Each entry is stored *pre-resolved* to the effective layer so `current_layer` is O(1). `oc_layer_depth()` + `truncate_oc_layers()` isolate a Form XObject's marked-content scope from its caller (defends against a malformed unbalanced `BDC` leaking out). `current_resources()` exposes the active resource dict so the resolver can read `/Properties` against the right scope.
3. **`src/document.rs`** — `extract_paths` and `process_form_xobject_paths` both handle `BeginMarkedContent` / `BeginMarkedContentDict` / `EndMarkedContent`. `/OC` property operands are resolved through `resolve_oc_layer_name(resources, properties)`, which:
   - inline-dict branch hands straight to `read_oc_name`;
   - name-reference branch (`/MC0` etc.) resolves through whatever resource scope the caller passes — page `/Resources` at page level, the XObject's own `/Resources` when extracting inside a Form XObject (§14.6.2, §8.10.1).
   `read_oc_name` reads either an OCG `/Name` *or* follows an OCMD's `/OCGs` (dict-or-array per Table 99) to the first member OCG, with a recursion depth guard (8). `/Name` strings go through the existing `Self::decode_pdf_text_string(&bytes)`, so PDFDocEncoding (Annex D), UTF-16BE and UTF-16LE (with BOM) all round-trip identically to the rest of the library. A small `deref_object` helper centralises single-level reference loading inside the OC name path.
4. **`src/python.rs::path_to_py_dict`** — exposes `layer: str | None` on every dict emitted by `extract_paths`, `extract_lines`, `extract_rects`.
5. **`src/writer/content_stream.rs` + `src/writer/document_builder.rs`** — mechanical `layer: None` fixes at the 4 PathContent literal sites required by the new struct field.

## Testing

- [x] All existing tests pass locally (`cargo test --lib` — 5257 passed, 2 ignored, 0 failed; `cargo test --test test_path_extraction` — 29 passed)
- [x] `cargo clippy --lib` passes (only the pre-existing `unknown lint: clippy::manual_checked_ops` from `clippy.toml`, unrelated to this PR)
- [x] `cargo fmt --check` passes

New test coverage:

**Rust integration** (`tests/test_path_extraction.rs`, 6 tests in `path_extractor_tests`): `test_layer_attached_when_oc_active`, `test_layer_none_outside_oc`, `test_layer_cleared_after_pop`, `test_outer_oc_survives_nested_non_oc_bmc`, `test_nested_oc_inner_wins`, `test_underflow_pop_is_safe`.

**Rust unit — extractor** (`src/extractors/paths.rs`, 2 new): `test_oc_layer_truncate_isolates_unbalanced_xobject` (Form XObject leak-isolation), `test_oc_layer_effective_is_o1_top_of_stack` (O(1) `current_layer` semantics).

**Rust unit — resolver** (`src/document.rs::tests`, 10 new):
- `test_oc_name_ocg_ascii`, `test_oc_name_ocg_utf16le_bom` (regression for the UTF-16LE branch the previous inline reader misread as latin-1), `test_oc_name_ocg_utf16be_bom`.
- OCMD: `test_oc_name_ocmd_single_ocg`, `test_oc_name_ocmd_ocgs_array_first_wins`, `test_oc_name_ocmd_depth_guard` (20-deep OCMD chain, depth guard at 8).
- Resolver entry: `test_resolve_oc_name_via_resources_properties` (proves Case 2 honours the *passed-in* resources — the crux of the XObject `/Resources/Properties` fix), `test_resolve_oc_name_inline_dict`, `test_resolve_oc_name_unresolvable_is_none`.
- End-to-end: `test_extract_paths_layer_none_for_plain_stroke` (full page pipeline through `from_bytes`).

**Python** (`tests/test_python.py`): `test_extract_paths_layer_field_shape` now builds a synthetic path-bearing PDF via `PdfDocument.from_bytes` so the per-path assertions actually iterate (the committed fixtures carry no vector geometry).

- [x] Python bindings have been updated
- [x] Python tests added

## Documentation

- [ ] README.md or documentation site has been updated
- [ ] Code examples have been updated
- [ ] CHANGELOG.md has been updated

> Note: I did not update CHANGELOG.md as I am not sure which version this would target. Happy to add an entry under the appropriate `## [Unreleased]` / next-version section if you point me to it.

## Checklist

- [x] My code follows the coding guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary (resolver helpers, OCMD branch, stack invariants, encoding handling, XObject scope isolation)
- [x] I have checked my code for spelling errors
- [x] PR title follows conventional commits format

## Additional Notes

- **Non-breaking, additive change** — `PathContent.layer` defaults to `None` for every PDF that does not declare `/OCProperties` and for every path emitted outside any `/OC` region. The new Python dict key is purely added.
- Name resolution falls back to `None` (not an error) for malformed PDFs, missing `/Resources /Properties` entries, OCG / OCMD objects without a resolvable name, or OCMD chains that exceed the recursion depth guard. Path extraction continues normally in those cases.
- **Out of scope** for this PR (issue #586 lists them under "Non-goals"): OCG visibility filtering (`/D` defaults, intents), nested OCG hierarchies (`/OCGs` groups, `/Order` arrays), XObject-level OCG inheritance (§8.11.3.3, the XObject *itself* carrying `/OC`). The PR delivers the `/Name` only — those higher-level concerns are easier to add incrementally on top.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -9854,12 +9854,108 @@ impl PdfDocument {
                     }
                 },
 
+                // Marked content operators — maintain the active Optional
+                // Content Group (PDF "layer") so each finalized path gets
+                // tagged with the OCG it was emitted under. Per ISO 32000-1
+                // §14.6, every `BDC`/`BMC` must be balanced by an `EMC`,
+                // so we always push (with `None` for non-`/OC` tags) and
+                // always pop — keeps the stack depth in sync with the
+                // marked-content nesting.
+                Operator::BeginMarkedContent { .. } => {
+                    extractor.push_oc_layer(None);
+                },
+                Operator::BeginMarkedContentDict { tag, properties } => {
+                    let layer = if tag == "OC" {
+                        self.resolve_oc_layer_name(page_dict, &properties)
+                    } else {
+                        None
+                    };
+                    extractor.push_oc_layer(layer);
+                },
+                Operator::EndMarkedContent => {
+                    extractor.pop_oc_layer();
+                },
+
                 // Skip other operators (text, images, etc.)
                 _ => {},
             }
         }
 
         Ok(extractor.finish())
+    }
+
+    /// Resolve a `BDC /OC <properties>` property operand to the human-readable
+    /// `/Name` of the Optional Content Group it refers to (PDF spec
+    /// ISO 32000-1:2008 §8.11, §14.6).
+    ///
+    /// `properties` is the operand parsed by `Operator::BeginMarkedContentDict`
+    /// — per spec it is either:
+    ///
+    /// 1. An inline dictionary: read its `/Name` entry directly.
+    /// 2. A name (e.g. `/MC0`) that references `page_dict /Resources
+    ///    /Properties <name>` → an indirect ref to an OCG dictionary → read
+    ///    its `/Name` entry.
+    ///
+    /// Returns `None` for malformed PDFs, missing `/Resources /Properties`
+    /// entries, or OCG objects without a `/Name`. Callers treat `None` as
+    /// "path belongs to no named layer" — extraction continues normally.
+    fn resolve_oc_layer_name(
+        &self,
+        page_dict: &std::collections::HashMap<String, crate::object::Object>,
+        properties: &crate::object::Object,
+    ) -> Option<String> {
+        use crate::object::Object;
+
+        // Helper to pull a UTF-8 `/Name` (decoded as a PDF text string) out
+        // of an OCG-like dictionary. OCG `/Name` is a string (§8.11.2.1),
+        // not a `/Name` PDF name.
+        fn read_name_field(dict: &std::collections::HashMap<String, Object>) -> Option<String> {
+            let name_obj = dict.get("Name")?;
+            match name_obj {
+                Object::String(bytes) => {
+                    // PDF spec: text strings may be PDFDocEncoding or UTF-16BE
+                    // with BOM. Detect BOM; otherwise treat as latin-1.
+                    if bytes.starts_with(&[0xFE, 0xFF]) {
+                        let u16s: Vec<u16> = bytes[2..]
+                            .chunks_exact(2)
+                            .map(|c| u16::from_be_bytes([c[0], c[1]]))
+                            .collect();
+                        String::from_utf16(&u16s).ok()
+                    } else {
+                        Some(bytes.iter().map(|&b| b as char).collect())
+                    }
+                },
+                Object::Name(s) => Some(s.clone()),
+                _ => None,
+            }
+        }
+
+        // Case 1: inline dictionary — read /Name directly.
+        if let Some(dict) = properties.as_dict() {
+            return read_name_field(dict);
+        }
+
+        // Case 2: name reference — look up page /Resources /Properties.
+        let prop_name = properties.as_name()?;
+        let resources = page_dict.get("Resources")?;
+        let resources_obj = if let Some(r) = resources.as_reference() {
+            self.load_object(r).ok()?
+        } else {
+            resources.clone()
+        };
+        let properties_dict = resources_obj.as_dict()?.get("Properties")?;
+        let properties_obj = if let Some(r) = properties_dict.as_reference() {
+            self.load_object(r).ok()?
+        } else {
+            properties_dict.clone()
+        };
+        let target = properties_obj.as_dict()?.get(prop_name)?;
+        let target_obj = if let Some(r) = target.as_reference() {
+            self.load_object(r).ok()?
+        } else {
+            target.clone()
+        };
+        read_name_field(target_obj.as_dict()?)
     }
 
     /// Extract rectangles from a page (v0.3.14).
--- a/src/document.rs
+++ b/src/document.rs
@@ -9866,7 +9866,7 @@ impl PdfDocument {
                 },
                 Operator::BeginMarkedContentDict { tag, properties } => {
                     let layer = if tag == "OC" {
-                        self.resolve_oc_layer_name(page_dict, &properties)
+                        self.resolve_oc_layer_name(extractor.current_resources(), &properties)
                     } else {
                         None
                     };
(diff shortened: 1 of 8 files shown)


willywg:
Heads-up on the red ❌ — the only failing check is `Composer security audit` and it's unrelated to this PR. Quick summary:

```
✓ No security vulnerability advisories found.
✗ Found 2 abandoned packages:
    sebastian/code-unit                 | none
    sebastian/code-unit-reverse-lookup  | none
Error: Process completed with exit code 2.
```

`composer audit` uses a bitmask exit code where `2` means *abandoned packages only — no real CVE/GHSA advisories* ([docs](https://getcomposer.org/doc/03-cli.md#audit)). The two packages are transitive PHPUnit dependencies (Sebastian Bergmann recently flagged them as abandoned on Packagist as part of the PHPUnit cleanup), and `composer audit` hits Packagist live so the same `composer.lock` that passed earlier started failing today.

Confirming this is environment-side, not from this diff:

- This PR only touches `src/document.rs`, `src/elements/path.rs`, `src/extractors/paths.rs`, `src/python.rs`, `src/writer/content_stream.rs`, `src/writer/document_builder.rs`, `tests/test_path_extraction.rs`, `tests/test_python.py` — zero PHP files / workflows / `composer.lock`.
- `PHP Bindings CI` on `upstream/main` was green at the time this PR opened.

Two ways to fix the CI when you have a moment, both single-line in `.github/workflows/php.yml:223`:

```yaml
# Option A — keep auditing, just don't fail on abandonment-only:
run: composer audit --locked --no-interaction --abandoned=report

# Option B — narrow the audit to security advisories:
run: composer audit --locked --no-interaction --abandoned=ignore
```

Happy to send a follow-up patch if you'd prefer, or you can roll it into a separate CI-hygiene commit — your call. Everything else on this PR (Rust lib + integration tests + lint + format + clippy) is green.

yfedoseev:
Thanks @willywg — really nice PR, well-scoped, clean stack design, the tests cover the tricky nesting cases, and the issue write-up made the spec context easy to verify. I cross-checked the implementation against `docs/spec/pdf.md` (ISO 32000-1:2008 §8.11 + §14.6) and have three things I'd like addressed before merge, plus a couple of minors.

### Please address

**1. OCMD payloads aren't resolved.** `resolve_oc_layer_name` in `src/document.rs` reads `/Name` directly from the dict, but §8.11.3.2 says a valid `/OC` operand is either an OCG *or* an Optional Content Membership Dictionary (Table 99: `/Type=OCMD`, `/OCGs`, `/P`, `/VE` — no `/Name`). Today any `/OC /MCx BDC` whose target is an OCMD returns `None` even though the spec treats it as valid optional content. Suggested: when `Type=OCMD`, follow `/OCGs` (deref + handle the `dictionary OR array` shape per Table 99), pick the first OCG, and recurse into `read_name_field`. Your "Non-goals" list calls out OCMD *visibility filtering*, which is a separate concern — name resolution is the same `/Name`-on-OCG problem you've already solved.

**2. Text-string decoding can reuse the existing helper.** `read_name_field` does `bytes.iter().map(|&b| b as char).collect()` for the non-BOM branch, which is Latin-1 — close but not the same as PDFDocEncoding. `src/document.rs` already has `Self::decode_pdf_text_string(&bytes)` (around line 7430) that uses `crate::fonts::font_dict::pdfdoc_encoding_lookup` for the full Annex D mapping AND handles UTF-16LE BOM (currently missed). For ASCII layer names like `A-GRID` it doesn't matter, but any glyph at `0x18-0x1F` or `0x7F-0x9F` decodes differently — and the DRY win is free.

**3. Form XObject `/Resources/Properties` not consulted.** The resolver only reads `page_dict["/Resources"]["/Properties"]`. §14.6.2 says properties come from the *current* resource dict, which inside a Form XObject content stream is the XObject's own `/Resources` (§8.10). `PathExtractor` already walks XObjects (`processed_xobjects`, `max_xobject_depth=100`), so `/OC /MC0 BDC` inside an XObject with locally-defined `/Properties/MC0` will fail to resolve. This is distinct from the "XObject-level OCG inheritance" non-goal you listed (that's §8.11.3.3, the XObject *itself* carrying `/OC`). Plumb the active resources through into the resolver. This one shows up in real CAD exports that use Form XObjects for repeated symbols (gridline labels, callouts).

If #1 or #3 aren't in scope for this PR, that's fine — just add them to the "Non-goals" list in the description so the gap is documented for whoever picks it up.

### Minors (optional)

- `tests/test_python.py::test_extract_paths_layer_field_shape` will silently pass on an empty `paths` list (the for-loop iterates zero times). Add `assert paths, "extract_paths returned no paths"` after the call.
- `current_layer()` does an O(stack-depth) scan per finalized path; tracking `active_layer: Option<String>` separately and updating on push/pop is O(1). Not worth bouncing the PR over.
- PR description undersells the diff — it's 8 files, not 6. The 4 `layer: None` callsites in `src/writer/{content_stream,document_builder}.rs` are necessary mechanical fixes for the new struct field, not bonus scope.

Tests, format, clippy all green. Once these are in, ready to land. Thanks again for the spec-careful approach and the construction-PDF context in #586.

willywg:
Thanks for the spec-careful review @yfedoseev — addressed all three asks plus the minors in b58fbab.

**1. OCMD payloads (§8.11.3.2, Table 99).** Added `read_oc_name` as the common entry. When `Type=OCMD`, it dereferences `/OCGs`, picks the first member that resolves to a dictionary (handles both the single-OCG-dict and array shapes; null array entries are skipped), and recurses to read the OCG's `/Name`. Recursion is depth-bounded (8) so a malformed OCMD chain pointing back to another OCMD can't loop forever — `test_oc_name_ocmd_depth_guard` covers a 20-deep chain.

**2. Text-string decoding.** `read_oc_name` now goes through `Self::decode_pdf_text_string(&bytes)` for the `Object::String` branch. PDFDocEncoding (Annex D), UTF-16BE (BOM), and UTF-16LE (BOM, previously misread as latin-1) all round-trip identically to the rest of the library — `test_oc_name_ocg_utf16le_bom` exercises an "ÁREA-Ø" name to prove the LE path. The `Object::Name` fallback stayed (non-conformant but seen in the wild).

**3. Form XObject /Resources/Properties (§14.6.2, §8.10.1).** `resolve_oc_layer_name` now takes `resources: Option<&Object>` and the page-level call passes `extractor.current_resources()` — the same scope `swap_resources` already maintains for nested `Do` resolution (page or current XObject). `process_form_xobject_paths` also gained the three marked-content arms (it previously dropped BDC/BMC/EMC), so `/OC` regions and local `/Properties` inside reused Form XObjects (gridline labels, callouts in CAD exports) now surface their layer. To match the existing save/restore discipline around XObjects, I snapshot `oc_layer_depth` on entry and `truncate_oc_layers(base)` on exit — that way a malformed, unbalanced `BDC` inside an XObject can't leak its layer onto the caller's subsequent paths. `test_resolve_oc_name_via_resources_properties` exercises the resources-parameter contract, and `test_oc_layer_truncate_isolates_unbalanced_xobject` covers the leak-isolation.

**Minors:**
- `current_layer` is O(1): `push_oc_layer` now stores the *effective* layer at each level (resolved OCG name when present, otherwise the previous top), so the stack top is always the active layer. Preserves the exact semantics of the previous `iter().rev().find_map` while being free per finalized path.
- `tests/test_python.py::test_extract_paths_layer_field_shape` builds a synthetic path-bearing PDF via `PdfDocument.from_bytes` so the per-path assertions actually run. `simple.pdf` carries no vector geometry (quick check confirmed `extract_paths` returns `[]` against it, so a bare `assert paths` would have failed), and `vector_graphics.pdf` is `#[ignore]`-gated as missing. A parallel Rust end-to-end test (`test_extract_paths_layer_none_for_plain_stroke`) covers the same path through `build_minimal_pdf` for in-tree validation.
- PR description: noted as 6 files but actually 8. I'll update the description shortly to reflect the cumulative scope (the `writer/` callsites + this round's resolver work).

12 new tests, full `cargo test --lib` 5257 passed / 2 ignored / 0 failed; `cargo clippy --lib` and `cargo fmt --check` clean.

Let me know if anything needs another pass.

yfedoseev (approved):
Approving — this is exactly the follow-through I was hoping for. Thank you @willywg. 🙏

All three asks landed cleanly and spec-faithfully:

1. **OCMD payloads** — `read_oc_name` follows `/OCGs` (dict-or-array, null-safe), depth-bounded against malformed chains. Matches §8.11.3.2 / Table 99.
2. **Text-string decoding** — now through `decode_pdf_text_string`, so PDFDocEncoding (Annex D) and UTF-16 BE/**LE** names round-trip with the rest of the library.
3. **Form XObject `/Resources/Properties`** — resolver consults the active (XObject-scoped) resources, the XObject loop gained the BDC/BMC/EMC arms, and the `oc_layer_depth` snapshot + `truncate_oc_layers` on exit cleanly isolates an unbalanced `BDC` from the caller. The save/restore ordering (truncate before restoring resources) is exactly right.

The O(1) `current_layer` (pre-resolving the effective layer on push) and the real path-bearing Python fixture are nice touches, and the 12 new tests cover the tricky cases (OCMD depth, UTF-16LE, resources contract, leak isolation). Defensive, backward-compatible (`#[serde(default)]`), and well-documented against the spec throughout.

**We're taking this into v0.3.57.** I'll integrate it onto the release branch (reconciling against the in-flight `document.rs` changes there), add it to the CHANGELOG, and credit you in the contributors. The only red check is the unrelated Ruby `Security audit (bundler-audit + OSV-Scanner)` — nothing to do with this PR.

Great work on the spec rigor — and thanks again for the construction-PDF context.

yfedoseev:
🎉 Shipped in **[v0.3.57](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.57)**! OCG (PDF layer) name on extracted paths — OCMD resolution, PDFDocEncoding/UTF-16 decoding, XObject-scoped `/Resources/Properties` with leak isolation — all in `main`. Really appreciate the spec-careful iteration, @willywg. Credited in the CHANGELOG + release notes. 🙏

PR #588, opened 2026-05-26
feat: page_image_handles() - two-phase image API with Form XObject support

### LINKED ISSUE #532: [Feature]: implement count-limited image extraction to avoid decompressing surplus streams
### Problem Statement

`  extract_images(page_idx) `decompresses every image stream on the page before returning, even when the caller only needs the first N images. For PDFs with large numbers of embedded images (e.g., InDesign/Acrobat exports with 50–200 XObjects per page), this causes multi-second hangs proportional to total image count regardless of the requested cap.

  The internal `ImageExtractFilter` struct already applies pre-decompression guards for minimum dimensions and maximum pixel count (`document.rs`), but has no `max_count` field. The `for op in operators` loop in `extract_images_filtered()` has no early exit once N images are collected.

### Proposed Solution

  implement a public count-limited variant:
```rust
  /// Extract at most `max_count` images from `page_index`.
  ///
  /// Decompression stops once `max_count` images have been collected,
  /// skipping remaining `Do` operators without loading their streams.
  ///
  /// Image order follows content-stream Do-operator order (paint order),
  /// which matches the rendering sequence on the page.
  pub fn extract_images_limited(
      &self,
      page_index: usize,
      max_count: usize,
  ) -> Result<Vec<PdfImage>>;
```
  Internally, add `max_count: Option<usize>` to `ImageExtractFilter` and break out of the operator loop in `extract_images_filtered()` when `images.len() >= max_count`.

### Alternatives Considered

  - **Expose a lazy iterator (`image_iter(page_idx) -> impl Iterator<Item = Result<PdfImage>>`)**: more flexible but a larger API surface change; the loop-with-early-exit approach is simpler and sufficient for the count-limit use case.
  - **Caller-side XObject walk**: possible but duplicates internal logic, misses inline images, and diverges from paint order. Not a viable long-term approach.

### Feature Area

Performance

### Priority

Important

### Use Case

  Downstream libraries that cap images per page (e.g., `max_images_per_page = 5`) currently  have no way to avoid the decompression cost of images beyond the cap. This forces them to implement their own XObject dictionary walk as a workaround, which: (a) duplicates internal pdf_oxide logic, (b) misses inline images entirely, and (c) diverges from content-stream paint order. A first-class `extract_images_limited()` eliminates all three problems.

### Example Usage (Optional)

```rust
// before: decompresses all images on the page, then clips the returned Vec
  let doc = PdfDocument::from_bytes(bytes)?;
  let images = doc.extract_images(0)?;
  let capped: Vec<_> = images.into_iter().take(5).collect();

  // after: decompresses exactly 5 images, skips decompressing the rest
  let images = doc.extract_images_limited(0, 5)?;
```

### Additional Context

  The internal change is small; `ImageExtractFilter` gains one field and the operator loop in `extract_images_filtered()` gains one guard. The existing `extract_images()` path is unaffected. The `extract_images_filtered()` private method already propagates the filter through Form XObject recursion, so count-limiting would work correctly for nested Form XObjects as well (each recursion level checks the same shared count).

### Checklist

kh3rld:
implements a two-phase API for working with images on PDF pages: enumerate metadata cheaply without decompressing any streams, then selectively materialise only the images you need.

## API

```rust
// Phase 1 - cheap metadata only
let handles = doc.page_image_handles(0)?;

for h in &handles {
    println!("{}×{}  filters={:?}", h.width, h.height, h.filter_chain);
}

// Phase 2 - materialise on demand
let img = handles[0].clone().decode()?;           // image::DynamicImage
let raw = handles[0].raw_compressed_bytes()?;     // original compressed bytes (e.g. JPEG)
```

## `PdfImageHandle`

All fields are public and the struct is `#[non_exhaustive]`.

| Field                  | Type              | Description |
|------------------------|-------------------|-------------|
| `width`, `height`      | `u32`             | Pixel dimensions from the XObject |
| `color_space`          | `ColorSpace`      | DeviceRGB, ICCBased, Indexed, etc. |
| `bits_per_component`   | `u8`              | Typically 8 |
| `byte_size_compressed` | `u64`             | Length of the compressed stream |
| `filter_chain`         | `Vec<PdfFilter>`  | Filters in stream order (e.g. `[DCTDecode]`) |
| `is_inline`            | `bool`            | `true` for inline images (`BI`/`ID`/`EI`) |
| `paint_order`          | `usize`           | Zero-based order of appearance in the content stream |
| `bbox`                 | `Rect`            | Final axis-aligned bounding box in page space |
| `rotation_degrees`     | `f32`             | Rotation derived from the CTM (0/90/180/270) |

`PdfFilter` and `PdfImageHandle` are re-exported from the crate root. `PdfFilter` also implements `serde::Deserialize`.

## Form XObjects

Images inside Form XObjects (including nested and multiply-referenced Forms) are fully supported. Recursion uses the same cycle detection and stream caching as the rest of the library. Returned handles carry the correct final `bbox` and `rotation_degrees` after all ancestor transforms.

## Tests

13 integration tests in `tests/test_page_image_handles.rs` covering basic usage, decode/round-trip, raw bytes, size filtering, paint order, JPEG marker validation, out-of-bounds pages, empty pages, and Form XObject recursion (including self-referential cycle safety).

fixed #532


yfedoseev:
Hi @kh3rld — thanks for taking #532 and running with it. The shape lines up cleanly with the spec we sketched there, and the extras (`bbox`, `rotation_degrees`, the 4-corner CTM transform, the hand-built Form-XObject negative test) are real improvements over what I asked for. CTM tracking, filter chain parsing, and the inline-image handling all look right.

One blocker before this is ready: **Form XObject recursion.** Real-world PDFs (Adobe, Office, page-reuse pipelines) commonly reference shared images via Form XObjects, and the existing `extract_images_filtered` recurses into them via `extract_images_from_form_xobject` (`src/document.rs:12906`). The new handle walker skips them and documents it as follow-on, but that ships a real capability gap vs. `extract_images`: a caller asking "what images are on this page?" gets fewer images than the spec model says exist. Easier to land it now while the walker is fresh than to bolt it on later.

Smaller items, none blocking:

- **Inline image stores bytes twice** — `image_handle_from_inline` builds `Object::Stream` with `Bytes::copy_from_slice(&data)` *and* keeps `compressed_data: data`. For a 10 MB inline JPEG that's 20 MB resident. `Bytes` is refcounted — share one and clone the handle into both fields.
- **`#[non_exhaustive]` consistency** — `PdfImageHandle`'s `Debug` impl uses `.finish_non_exhaustive()` but the struct itself has all-`pub` fields with no `#[non_exhaustive]` attribute, so callers can match exhaustively anyway. Either add the attr to the struct or drop it from `Debug`.
- **Follow-up issue (not this PR):** unifying the new walker with `extract_images_filtered` so they share one CTM-tracking visitor, with `extract_images()` becoming a thin wrapper as we discussed in #532. Happy to file that as a separate issue tagged `good first followup` after this lands — no need to take it on in this PR.

Mark this ready-for-review once Form XObject recursion is in and the two small items are addressed. Thanks again — exactly the contribution shape I was hoping for.

kh3rld:
> Hi @kh3rld — thanks for taking #532 and running with it. The shape lines up cleanly with the spec we sketched there, and the extras (`bbox`, `rotation_degrees`, the 4-corner CTM transform, the hand-built Form-XObject negative test) are real improvements over what I asked for. CTM tracking, filter chain parsing, and the inline-image handling all look right....

nice!

yfedoseev:
@kh3rld — CI is mostly green (159 pass / 30 skip). One real failure to fix on your side, plus two unrelated infra reds that aren't your problem (tracked separately in #599).

### Please fix in your next push

`cargo doc -D warnings` fails on all 4 Test jobs (ubuntu stable + beta, macos, windows) with the same 3 broken intra-doc links in the new docstrings:

```
error: unresolved link to `PdfImageHandle::decode`
  --> src/document.rs:12630

error: unresolved link to `PdfImageHandle::raw_compressed_bytes`
  --> src/document.rs:12631

error: unresolved link to `PdfDocument::page_image_handles`
  --> src/extractors/images.rs:2248
```

The bracket-linked types aren't in scope at those doc sites. Two ways to fix:

- **Fully-qualify the links** — e.g. ``[`crate::PdfImageHandle::decode`]`` and ``[`crate::PdfDocument::page_image_handles`]``.
- **Or add `use` imports** at the top of the affected modules so the short names resolve.

Either works; fully-qualified is usually the lower-friction choice for rustdoc on public-API types.

For context: the actual `cargo test` runs are all green (5,245 unit tests pass on every platform). It's purely a docstring polish issue. Please bundle this with the items from my earlier review on this PR (Form XObject recursion is already in `f66e325f` — thanks for that; remaining items are the `Bytes` sharing comment, `#[non_exhaustive]`, doc tightening — refer to the prior review for the full list).

### Not on your plate

`Composer security audit` and `Dependency Check (cargo-deny)` are red on every PR right now (PHPUnit abandoned packages + yanked `aes 0.9.0`). Filed as #599 to fix on `main` independently — please ignore those in your re-push.

Thanks again!

kh3rld:
thanks for the updates,... I think this is ready!

yfedoseev (approved):
@kh3rld — verified everything against the two prior reviews:

- **Form XObject recursion** (the blocker) is in via `collect_handles_from_do` + `collect_image_handles_from_form_xobject` with proper cycle detection (ObjectRef stack + depth 100), Form `/Resources` fallback, `/Matrix`, `xobject_stream_cache` reuse, and `parent_ctm × form_matrix` composition at Form entry. The positive JPEG-inside-Form test plus the self-reference cycle safety test pin the behaviour nicely. ✅
- **Inline-image `Bytes` double-store** is fixed — the synthetic `Object::Stream::data` and `PdfImageSource::Inline::compressed_bytes` share one refcounted `bytes::Bytes` clone (`src/extractors/images.rs:2552-2566`), with an explicit comment about the 10 MB-becomes-20 MB scenario. ✅
- **`#[non_exhaustive]`** added to `PdfImageHandle` (`:2271`) so the `Debug::finish_non_exhaustive()` is no longer cosmetic. ✅
- **`cargo doc -D warnings`** — all three links rewritten as `` [`crate::PdfImageHandle::…`] `` / `` [`crate::PdfDocument::page_image_handles`] ``, the fully-qualified path I suggested. ✅

One observation, not blocking: `gh pr checks` and the GitHub check-runs API both show zero checks on `f64b248a` — looks like the workflows haven't re-fired on the last two commits. Probably fork-Actions gating; could you push an empty commit (`git commit --allow-empty -m "trigger ci"`) or I can re-run them from this side. Once the same green status returns (159 pass / 30 skip from your previous run), this is ready to merge.

Filing the "walker unification" follow-up issue we discussed (collapse `extract_images_filtered` + `page_image_handles` into one CTM-tracking visitor) as `good first followup` after this lands — tracked but not on your plate.

Thanks for the careful work on this — the comprehensive Form XObject implementation went well beyond the minimum, the `Bytes` refcount comment is exactly the kind of in-code rationale that helps future readers, and the foundation tests (positive Form recursion + self-reference cycle) are the right shape. Really enjoyable contribution to review.

yfedoseev:
🎉 Thank you, @kh3rld! `page_image_handles()` (two-phase image API with Form XObject recursion) is **picked for v0.3.57** — merged into the `release/v0.3.57` branch and shipping with the next release.

Clean merge, and thanks for proactively fully-qualifying the intra-doc links. All 13 `test_page_image_handles` cases are green. Credited in the CHANGELOG. 🙏

yfedoseev:
🎉 Shipped in **[v0.3.57](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.57)**! `page_image_handles()` (two-phase enumerate-then-materialize, with Form XObject recursion) is in `main`. Thanks @kh3rld — credited in the CHANGELOG + release notes. 🙏

PR #589, opened 2026-05-26
Add regression coverage for the '★' UTF-8 char-boundary panic

vincenzopalazzo:
## Summary
- **Bug:** v0.3.15 panicked during text extraction on documents containing runs of '★' (U+2605, 3-byte UTF-8) in a font-change region:
  ```
  thread 'tokio-rt-worker' panicked at src/extractors/text.rs:2963:31:
  end byte index 10 is not a char boundary; it is inside '★' (bytes 9..12 of string)
  ```
  The panic crashed the whole extraction task, forcing the caller to fall back to vision-based extraction on every affected document.
- **Status in upstream:** Already fixed — `crate::utils::safe_prefix` / `safe_suffix` now wrap both byte-slice sites. The old font-change-merge log was refactored away, and the [`FLUSH_TJ_SPAN_BUFFER` trace](src/extractors/text.rs:6735) now goes through `safe_prefix(&span.text, 20)`. **No source change needed on top of `main`.**
- **What this PR does:** Adds the regression scaffolding that was missing when the bug shipped, so this can't silently regress.

## Added
- [tests/fixtures/kangrejos_2026.pdf](tests/fixtures/kangrejos_2026.pdf) — the real document from the bug report (4.8 MB, 50 pages). Exercises the panic-path against the actual content that triggered it in production.
- [tests/test_kangrejos_2026_utf8_panic_regression.rs](tests/test_kangrejos_2026_utf8_panic_regression.rs):
  - `kangrejos_2026_extract_all_pages_no_panic` — extracts every page with `log::set_max_level(Trace)` so the previously-offending log format args are actually evaluated; asserts non-empty text emerges.
  - `char_boundary_slicing_handles_multibyte_at_cut_point` — synthetic prefix + suffix UTF-8 case that survives even if the fixture is ever pruned, demonstrating that `&s[..N]` / `&s[len-N..]` panics on a multi-byte codepoint at the cut point while the char-boundary-rounded shape does not.

## Test plan
- [x] Verified by temporarily reverting the upstream `safe_prefix` wrapping locally: the integration test reproduces the original panic message **verbatim** (`end byte index 10 is not a char boundary; it is inside '★' (bytes 9..12) of '★★★★★'`).
- [x] Restored upstream's wrapping: both tests pass.
- [x] `cargo build` ✓ · `cargo fmt --check` ✓ · `cargo clippy --all-targets -- -D warnings` ✓.
- [x] Full suite: `cargo test --test test_kangrejos_2026_utf8_panic_regression` 2/2 pass.

## Notes
- The 4.8 MB PDF fixture is larger than the existing 4 KB ones. It's the real reporter document. Happy to slim to just the offending pages on request — but the full doc has the advantage of exercising the same merge / TJ-flush paths the production crash hit.


yfedoseev:
Hi @vincenzopalazzo — thanks for taking the time to write this up. A few thoughts before we go further:

**1. The source fix is already in, and a regression test already exists.**

The panic was fixed in commit `cb71a70c` ("fix: prevent panic on multi-byte UTF-8 in log string slicing"), which first shipped in **v0.3.19** and has been in every release since (current: v0.3.55, with v0.3.56 in prep). That commit also added `tests/test_utf8_char_boundary_safety.rs` (141 lines), which already covers the same failure mode — including the exact `&s[..N]` panic shape on a multi-byte codepoint at the cut point. So neither the source path nor the regression scaffolding has an obvious gap to fill here.

A side note on the synthetic test in this PR: it re-implements `safe_prefix` / `safe_suffix` locally (because they're `pub(crate)`) and asserts against those local copies. That means it doesn't actually exercise the crate's helpers — if `safe_prefix` ever regressed upstream, this test would still pass. The existing test in `tests/test_utf8_char_boundary_safety.rs` is structured differently and doesn't have that issue.

**2. The PDF fixture would need some clarification before we can merge it.**

"Kangrejos" is a real Rust-for-Linux unconference, so `kangrejos_2026.pdf` is most likely third-party content — proceedings, agenda, slides, or registration material. Merging a 4.8 MB binary into `tests/fixtures/` would make this repo the redistributor of that content under its license, so it would help to know:

- Who owns the copyright and whether redistribution is permitted
- Whether the document contains attendee PII (names, emails, signatures — common in conference materials)
- The provenance of the file (yours, your employer's, public-domain, etc.)

This isn't specific to your submission — it's the policy we'd want to apply to any binary fixture going into the repo.

**Possible paths forward, if you'd like to keep going:**

- A synthesized minimal PDF (≤ 10 KB) that reproduces the panic path with no third-party content — a hand-written PDF with `★` glyphs in a font-change region should be tractable, and would be the cleanest outcome.
- Or, a licensing/provenance statement on the existing fixture if it's fully yours to relicense.
- Or, if `tests/test_utf8_char_boundary_safety.rs` already catches what you were concerned about, that's also a fine place to land.

Leaving this open — happy to revisit whenever you've had a chance to think about it. Thanks again for the careful writeup.

vincenzopalazzo:
Fair points across the board — closing this.

ACK on (1): I missed `cb71a70c` / `test_utf8_char_boundary_safety.rs` when triaging the panic against v0.3.15. Coverage there is already a strict superset of what this PR adds, including the synthetic-test flaw you flagged (the local re-implementation of `safe_prefix`/`safe_suffix` doesn't actually exercise the crate helpers).

ACK on (2): I don't have provenance to share for that PDF — it was a host document that surfaced the panic in the wild, not something I own or have a redistribution license for. A synthesized minimal PDF is the right shape if anyone ever wants a real end-to-end fixture, but with the string-level coverage already in place it's not earning its keep here.

Thanks for the thorough review.

vincenzopalazzo:
Closing per review discussion — upstream already has equivalent coverage.

PR #596, opened 2026-05-27
fix: cache poisoning when fonts collide on basename with different To…

### LINKED ISSUE #595: [Bug]: reliance on BaseFont name instead of the ToUnicode stream causes corrupted processing
### Bug Description

```rust
if let Some(to_unicode) = d.get("ToUnicode") {
      4u8.hash(&mut hasher);
      if let Some(r) = to_unicode.as_reference() {
          r.id.hash(&mut hasher);    // PDF object ID
          r.gen.hash(&mut hasher);   // PDF generation number
      }
  }
```

The ToUnicode stream is hashed by its PDF object reference ID — not its content. Two fonts in different PDFs with the same BaseFont name (e.g., AAAAAA+ArialUnicodeMS), same Subtype, same Encoding, but different ToUnicode streams will hash to the same key if their ToUnicode object IDs happen to match (or both are inline). The second document gets the first document’s cached FontInfo with the wrong ToUnicode mapping.

Both test PDFs use ArialUnicodeMS as a subset font. reportlab generates it as AAAAAA+ArialUnicodeMS. The CJK PDF’s font subset includes CJK characters in its ToUnicode CMap. The Unicode Extended PDF’s font subset includes Greek/Cyrillic/etc. But because the font names and object IDs can collide, the global cache returns the CJK font’s FontInfo for the Unicode Extended PDF, so ° resolves to 日 (the CJK mapping for that CID).

### Steps to Reproduce

1. process 02_cjk_mixed.pdf 
2. 03_unicode_extended.pdf 

the degree sign becomes 日

### Expected Behavior

pdf_oxide correctly processes the unicode using without font corruption caused by incorrect cache

### Actual Behavior

the degree sign becomes 日

_No response_

### Environment

- **OS**: macOS 26.5
- **Rust version**: rustc 1.93.0 (254b59607 2026-01-19)

[02_cjk_mixed.pdf](https://github.com/user-attachments/files/28302949/02_cjk_mixed.pdf)
[03_unicode_extended.pdf](https://github.com/user-attachments/files/28302948/03_unicode_extended.pdf)

- **pdf_oxide version**: 0.3.55
- **Python version** (if using Python bindings): 3.13

### Code Sample

```python
"""Reproducer: global font cache returns wrong ToUnicode for subset fonts.

Processing a CJK PDF first poisons the cache so that a subsequent PDF
with the same subset-prefixed font name gets the wrong CMap. The degree
sign ° (U+00B0) becomes 日 (U+65E5).
"""

import pdf_oxide
from pathlib import Path

CJK_PDF = "02_cjk_mixed.pdf"
UNICODE_PDF =  "03_unicode_extended.pdf"

# Process CJK PDF first — poisons the global font cache
doc1 = pdf_oxide.PdfDocument.from_bytes(CJK_PDF.read_bytes())
_ = doc1.extract_chars(0)

# Now process the Unicode PDF — gets wrong CMap from cache
doc2 = pdf_oxide.PdfDocument.from_bytes(UNICODE_PDF.read_bytes())
chars = doc2.extract_chars(0)
text = "".join(c.char for c in chars)

degree_chars = [(i, c) for i, c in enumerate(text) if c in ("°", "日")]
for i, c in degree_chars:
    ctx = text[max(0, i-5):i+5]
    print(f"  pos {i}: U+{ord(c):04X} ({c!r}) in {ctx!r}")

has_degree = "°" in text
has_corruption = "日" in text and "日" not in UNICODE_PDF.stem

if has_corruption:
    print(f"\nBUG: degree sign ° was replaced by 日 (CJK sun/day)")
    print(f"     Global font cache returned CJK ToUnicode for subset font")
    exit(1)
elif has_degree:
    print(f"\nOK: degree sign ° preserved correctly")
    exit(0)
else:
    print(f"\nUNEXPECTED: neither ° nor 日 found")
    exit(2)

```

### Error Messages / Logs

```shell

```

### Additional Context

_No response_

### Checklist

RayVR:
## Description

resolves issue #595 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Fixes #595

## Changes Made

1. font_identity_hash_cheap — Return type changed from u64 to (u64, bool). The bool indicates whether the font has a subset prefix (AAAAAA+). Subset fonts have document-specific ToUnicode mappings that are unsafe to share across documents.

2. load_fonts — The global cross-document font cache (Layer 6) is now skipped for subset fonts, both for lookup and insertion. The per-document cache (Layer 5) still works for subset fonts.

3. Tests — Updated 4 existing tests to destructure the new return type. Added 1 new test test_font_identity_hash_detects_subset_prefix that verifies AAAAAA+ArialUnicodeMS is detected as subset and ArialUnicodeMS is not.

Root cause: Two different PDFs with the same subset-prefixed font name (e.g., AAAAAA+ArialUnicodeMS) but different glyph subsets would share a cached FontInfo across documents. The first document's ToUnicode CMap would be used for the second document, mapping characters to wrong codepoints (e.g., ° → 日).

## Testing

- [x] I have run `cargo fmt`

- [ ] Python tests pass

## Documentation

## Checklist

## Follow-ups (not in this PR) 

Issues #597 and #598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to PDFOxide are documented here.
 
+## [Unreleased]
+
+### Fixed
+- **#595** — Cross-document font cache poisoning when subset fonts
+  (e.g. `AAAAAA+Arial`) from different PDFs collide on BaseFont name.
+  The global cache now skips subset fonts entirely, and fonts without
+  a `BaseFont` entry default to `is_subset = true` as a fail-safe.
+
 ## [0.3.55] - 2026-05-25
 
 > Ruby + PHP language bindings + multi-line heading reading-order fix
--- a/src/document.rs
+++ b/src/document.rs
@@ -10689,15 +10689,30 @@ impl PdfDocument {
     /// uniquely identify fonts within a document. For reference-only fields (ToUnicode,
     /// FontDescriptor, DescendantFonts), hashes their presence to avoid false positives
     /// between fonts with vs without these features.
-    fn font_identity_hash_cheap(font_obj: &Object) -> u64 {
+    /// Returns `(hash, is_subset)`.
+    ///
+    /// Subset fonts (e.g. `AAAAAA+ArialUnicodeMS`) have document-specific
+    /// glyph subsets and ToUnicode mappings that are NOT safe to share across
+    /// documents even when the BaseFont name matches. The caller must skip
+    /// the global cross-document cache for subset fonts.
+    fn font_identity_hash_cheap(font_obj: &Object) -> (u64, bool) {
         use std::hash::{Hash, Hasher};
         let mut hasher = std::collections::hash_map::DefaultHasher::new();
+        let mut is_subset = false;
 
         if let Some(d) = font_obj.as_dict() {
             // BaseFont: primary identity — unique per font within a document
             if let Some(Object::Name(n)) = d.get("BaseFont") {
                 1u8.hash(&mut hasher);
                 n.hash(&mut hasher);
+                // Detect subset prefix: 6 uppercase ASCII letters followed by '+'
+                let name_bytes = n.as_bytes();
+                if name_bytes.len() > 7
+                    && name_bytes[6] == b'+'
+                    && name_bytes[..6].iter().all(|b| b.is_ascii_uppercase())
+                {
+                    is_subset = true;
+                }
             }
             // Subtype: Type1, TrueType, Type0, CIDFontType0, CIDFontType2
             if let Some(Object::Name(n)) = d.get("Subtype") {
@@ -10709,12 +10724,16 @@ impl PdfDocument {
                 3u8.hash(&mut hasher);
                 match enc {
                     Object::Name(n) => n.hash(&mut hasher),
-                    Object::Reference(_) => b"enc_ref".hash(&mut hasher),
+                    Object::Reference(r) => {
+                        b"enc_ref".hash(&mut hasher);
+                        r.id.hash(&mut hasher);
+                        r.gen.hash(&mut hasher);
+                    },
                     Object::Dictionary(_) => b"enc_dict".hash(&mut hasher),
                     _ => {},
                 }
             }
-            // ToUnicode: hash content via reference or inline presence
+            // ToUnicode: hash reference ID (unique within a document)
             if let Some(to_unicode) = d.get("ToUnicode") {
                 4u8.hash(&mut hasher);
                 if let Some(r) = to_unicode.as_reference() {
@@ -10737,7 +10756,7 @@ impl PdfDocument {
                 }
             }
         }
-        hasher.finish()
+        (hasher.finish(), is_subset)
     }
 
     /// Load fonts from a Resources dictionary into the extractor.
@@ -10913,10 +10932,11 @@ impl PdfDocument {
                         let font = self.load_object(font_ref)?;
 
                         // Compute identity hash (cheap: 3-6 dict lookups, ~200ns)
-                        let id_hash = Self::font_identity_hash_cheap(&font);
+                        let (id_hash, is_subset) = Self::font_identity_hash_cheap(&font);
 
                         // Layer 5: Per-font identity cache — skip from_dict when a
-                        // structurally identical font was already parsed elsewhere.
+                        // structurally identical font was already parsed elsewhere
+                        // in the SAME document.
                         let cached_identity_opt = self
                             .font_identity_cache
                             .lock_or_recover()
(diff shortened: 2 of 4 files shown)


yfedoseev:
Hi @RayVR — thanks for this, really clean fix. The root-cause analysis in #595 was spot-on, the subset-prefix detector matches PDF 32000-1 §9.6.4 exactly (`docs/spec/pdf.md:17842` — "exactly six uppercase letters" + `+`), and the synthesized regression test in `tests/test_subset_font_cache.rs` is exactly the right shape — much better than shipping the issue's binary PDFs. The Encoding-ref tightening (hashing `id`/`gen` instead of the constant `b"enc_ref"`) is a nice quiet improvement on top.

A few items before merge:

### Please address

**1. Legacy tests at `src/document.rs:15839`, `:15853`, `:15866` weren't updated for the new tuple return.** They compile and pass — `(u64, bool)` is `PartialEq` so `assert_eq!(hash1, hash2)` still works — but they read as "forgot to update." Mechanical destructure to `let (hash1, _) = ...` would match the style of your new tests at `:17131+`.

**2. Silent semantic change at `src/document.rs:10894` and `:11054`.** `Self::font_identity_hash_cheap(&font).hash(&mut h)` now folds the `is_subset` bool into the combined font-set cache key (because `(u64, bool)` implements `Hash`). Behavior-wise it's fine — the flag is a deterministic property of the font — but a future reader will trip over it. Either destructure and hash only `id_hash` (preserves prior semantics), or leave as-is with a one-line comment ("subset flag intentionally included; deterministic per font").

**3. `is_subset` is `false` when `BaseFont` is absent.** Today the detector only sets the flag inside the `Some(Object::Name(n)) = d.get("BaseFont")` arm. If a font dict reaches `load_fonts` without a `BaseFont` (malformed/inline), `is_subset` stays `false` and the font goes into the global cache. Worth either:
   - asserting `BaseFont` presence is checked upstream in `FontInfo::from_dict`, so this branch is unreachable; or
   - defaulting `is_subset = true` (fail-safe — exclude from global cache) when BaseFont is missing, since we can't prove safety without it.

The cost of option (b) is at most one re-parse per such font.

### Minors

- **Doc comment on `font_identity_hash_cheap`** still says "For reference-only fields (ToUnicode, FontDescriptor, DescendantFonts), hashes their presence to avoid false positives." That "presence only" phrasing is now misleading since the ToUnicode arm hashes `id`/`gen`, and Encoding-ref does too. Tighten to reflect that indirect refs participate in the hash by ID.
- **Test file `clear_global_font_cache()` is process-global.** Integration tests run in their own binary so cargo's per-binary isolation protects us today, but if anyone later adds a second `#[test]` here that relies on warm cache state, this will surprise them. A one-line comment at the top of the test ("relies on having the whole binary to itself") is enough.
- **CHANGELOG entry missing** — please add a `Fixed` entry under the next-release section noting the cross-document cache poisoning for subset fonts (cite #595).

### Out of scope for this PR — intentionally deferred

Per spec §9.6.5 (Type 3, `docs/spec/pdf.md:17860`) and @haberman's follow-up comment on #595, two more "false-hit" classes exist that this PR intentionally does not cover. I've filed them as separate issues so #595 can close cleanly once this lands:

- **#597** — Type 3 fonts must not be cached in the cross-document global font cache (inherently document-local glyph procs; no subset prefix, so the new guard still admits them to Layer 6).
- **#598** — Cross-document font cache returns wrong `/Widths` when two PDFs share a `BaseFont` name with different metrics (affects standard-14 overrides and custom non-subset fonts).

Both are real and the same shape as your fix; please mention them in this PR's description under a "Follow-ups (not in this PR)" section so the scope boundary is explicit. After this lands I'd suggest #595 stays open until #597 and #598 are also resolved — or we close #595 and let #597/#598 carry the residual.

Thanks again — careful spec work and a real regression test, exactly the contribution shape I love to see.

yfedoseev:
Thanks @RayVR — this addresses the substantive items cleanly. Verified against `6c977cd3`:

- Both call sites (`src/document.rs:10891`, `:11050`) now destructure to `(id_hash, _)` and only the hash participates in the combined font-set fingerprint — prior semantics preserved exactly.
- The fail-safe is in nicely: `is_subset` defaults to `true` and the BaseFont arm assigns the predicate (`:10702`, `:10711`), so any path that misses the BaseFont arm (no key, BaseFont isn't a `Name`, non-dict object) leaves the font excluded from the global cache. `test_font_identity_hash_null_object` was upgraded to actually verify this — nice touch.
- Legacy tests destructured as suggested (`:15845`, `:15858`), CHANGELOG entry added under `[Unreleased] / Fixed` with the #595 cite.

Two tiny things, neither a blocker:

- The "process-global state" comment landed on the unit test in `src/fonts/global_cache.rs:361` rather than on the integration test in `tests/test_subset_font_cache.rs`. Was that intentional? If you meant it for the integration test that was the original target — the unit test is fine either way since each cargo binary runs isolated.
- The doc comment on `font_identity_hash_cheap` (`:10690-10693`) now says both "hashes their values" and "hashed by object ID, not just presence." Either phrasing alone is fine; together they read slightly contradictory. One more pass that distinguishes "direct dicts → contents hashed, indirect refs → (id, gen) hashed" would tighten it.

If you can add a short "Follow-ups (not in this PR)" section to the PR body pointing at #597 and #598, I'll merge this. Otherwise happy to merge as-is and capture that on the issue.

RayVR:
@yfedoseev comment added to PR

yfedoseev:
@RayVR your 12:48Z comment came through empty on my end — just "comment added to PR" with no body and no inline-thread reply. Could you re-post? Otherwise I think we're at LGTM on this one.

RayVR:
@yfedoseev sorry that's just confirmation I have updated the PR message to reference the unaddressed issues, 597 and 598

yfedoseev:
@RayVR — first, thank you. This was genuinely heroic work. The root-cause analysis in #595 was exactly right and it took real insight: object IDs are per-document, so folding a ToUnicode `7 0 R` into the cache key can *never* disambiguate two documents that both happen to place their ToUnicode at object 7 — a false cache hit that silently decodes the second document's text with the first's mapping. That's a subtle, nasty bug and you nailed both the diagnosis and the shape of the fix.

I want to be transparent about how this is landing, because it isn't a plain merge and you deserve the full picture.

**What happened to the base.** While this PR was open, `main` moved on to v0.3.58, which independently added a `font_is_document_local` gate (the Type 3 cross-document exclusion) and folded width metrics into the identity hash. That work changed `font_identity_hash_cheap`'s signature out from under your branch, so your `(u64, bool)` approach no longer applies cleanly — and, importantly, **the subset bug you found was still not fixed on `main`**, because that new gate only covered Type 3 fonts. Your PR is the piece that actually closes #595.

**What we did, rather than ask you to rebase.** You'd already done the hard part, so we carried your work forward ourselves onto the v0.3.59 branch and reconciled it with the new gate:

- Your subset-prefix detector (exactly six uppercase letters + `+`, per ISO 32000-1 §9.6.4) is preserved verbatim — it was exact.
- Instead of the parallel `(u64, bool)` hash signature, we folded the subset check **into** `font_is_document_local`, so Type 3 fonts and subset fonts now share a single cross-document cache-exclusion predicate. Same correctness outcome, no hash/call-site churn, one place to reason about.
- Your end-to-end regression test (`tests/test_subset_font_cache.rs`) came across as-is. We confirmed it genuinely fails without the gate (PDF 2 decodes `"A"` from PDF 1's cached font) and passes with it (`"Z"`), and added unit coverage for the subset-prefix edge cases.

Your authorship is preserved via a `Co-authored-by` trailer on the integration commit, and you're credited by name in the v0.3.59 CHANGELOG.

It will ship in **v0.3.59**. I'll close this PR once that release merges, since the content is already integrated there — but the credit is, and stays, yours. The follow-ups you flagged (#597/#598) are already on `main` as of v0.3.58. Thank you again for the careful, high-quality work — it's a pleasure reviewing your contributions.

yfedoseev (approved):
Approving to formally bless this work. The root-cause analysis and the fix are both correct, and the regression test is a faithful, fails-without-the-gate reproduction.

For the record on how it lands: because `main` advanced to v0.3.58 while this was open (which added the `font_is_document_local` gate and changed the identity-hash signature), the change has been carried onto the v0.3.59 branch and reconciled there — the subset-prefix detector and your regression test are preserved, folded into the single document-local predicate. So the substance is approved and integrated; the commits reach `main` via the v0.3.59 release. Credit is preserved via a co-author trailer and a named CHANGELOG entry. Thank you again, @RayVR.

yfedoseev:

PR #600, opened 2026-05-28
feat: ocg ink filtering 

RayVR:
## Description

Add OCG (Optional Content Group) layer filtering and Separation/DeviceN ink filtering to text extraction. Allows callers to exclude text rendered in specific layers or spot-color inks — the primary use case is stripping prepress overlays (dielines, varnish marks, approval stamps) from production PDFs.

## Type of Change

- [x] Tests

### New public API
- `get_layers()` — list OCG layer names from the document catalog
- `get_page_inks(page)` — list Separation/DeviceN ink names on a page
- `extract_text_filtered(page, excluded_layers, excluded_inks)` — text extraction with layer/ink suppression, using the same full assembly pipeline as `extract_text` (structure-tree ordering, table detection, column detection)
- `extract_chars_filtered(page, excluded_layers, excluded_inks)` — character extraction with layer/ink suppression
- Python: `extract_text()` and `extract_chars()` gain `exclude_layers` and `exclude_inks` keyword arguments; region filtering composes correctly with both

### Layer filtering (`src/extractors/text.rs`)
- Tracks OCG scope via BDC/EMC marked content stack — content within an excluded layer's scope is suppressed at all span and character emission points
- Resolves both direct OCG dictionaries and OCMD (Optional Content Membership Dictionaries) per ISO 32000-1 §8.11.2

### Ink filtering (`src/extractors/text.rs`)
- Tracks fill color space changes (`cs`, `rg`, `g`, `k` operators) and checks Separation/DeviceN definitions against the exclusion set
- Falls back to full content stream parser when ink exclusions are active, since the text-only fast-path parser skips color operators
- Correctly restores ink exclusion state across graphics state save/restore (`q`/`Q`) and Form XObject boundaries
- Bypasses XObject span cache when any filter is active to prevent cross-call interference

### Refactoring
- Deduplicated extraction boilerplate into shared `extract_spans_impl` / `extract_chars_impl` helpers
- Extracted `postprocess_spans` and `assemble_text_from_spans` so filtered and unfiltered paths share the same pipeline

## Testing

- [x] I have run `cargo fmt`

11 integration tests in `tests/test_ocg_ink_filtering.rs`, all using synthetic PDF construction:
- Inline OCG layer filtering (BDC/EMC)
- OCMD membership dictionary resolution
- Separation ink filtering via color space tracking
- DeviceN all-or-nothing suppression (documents intentional semantics)
- Form XObject: span cache isolation under filters
- Form XObject: ink state restoration after implicit Q
- Parser fallback: ink filtering with full content stream parser
- Region + filter composition
- Empty filters produce identical output to unfiltered
- Non-matching layer exclusion produces identical output (pipeline parity)
- `get_layers()` / `get_page_inks()` correctness

- [x] Python tests pass

New methods: `get_layers()`, `get_page_inks(page)`. Existing `extract_text()` and `extract_chars()` extended with `exclude_layers` and `exclude_inks` keyword arguments.

## Documentation

Doc comments on all new public and key internal methods. DeviceN semantics and pipeline parity documented inline.

## Checklist

## Additional Notes

**DeviceN is all-or-nothing:** For `[/DeviceN [/Cyan /SpotGold] ...]`, excluding "SpotGold" suppresses all text in that color space. Tint values are not evaluated — this is the correct default for prepress workflows where any content touching a spot ink should be fully removed.

**OCMD visibility policies** (`/P /AllOn`, `/AnyOn`, etc.) are not evaluated — any excluded OCG in the membership triggers suppression. Full policy evaluation would require tracking OCG on/off state from the document's default configuration, which is out of scope.

**Performance:** Zero impact on unfiltered extraction. Ink-filtered extraction uses the full content stream parser instead of the text-only fast path. Layer-only filtering has no parser overhead.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -8652,9 +8652,19 @@ impl PdfDocument {
 
     /// Extract text from a page, excluding content from specified layers and inks.
     ///
-    /// Uses a simplified text assembly pipeline (no structure-tree ordering or
-    /// table detection). For most layer/ink filtering use cases this is
-    /// sufficient since the caller is performing targeted extraction.
+    /// # Text assembly differences
+    ///
+    /// This method uses a simplified text assembly pipeline compared to
+    /// [`extract_text`](Self::extract_text): spans are sorted by row-aware reading order and joined
+    /// with line breaks based on Y-gap heuristics. The full pipeline's
+    /// structure-tree ordering, table detection, and column detection are
+    /// **not** applied. This means:
+    /// - Multi-column layouts may interleave columns
+    /// - Tables will not be reconstructed
+    /// - Reading order depends on coordinate sorting, not document structure
+    ///
+    /// For precise spatial control, use [`extract_chars_filtered`](Self::extract_chars_filtered) and assemble
+    /// text yourself, or apply region filtering on top.
     ///
     /// **Ink filtering note:** For DeviceN color spaces, text is suppressed if
     /// ANY ink in the DeviceN array matches an excluded ink name. Tint values
--- a/src/document.rs
+++ b/src/document.rs
@@ -4694,9 +4694,19 @@ impl PdfDocument {
         page_index: usize,
         options: &crate::converters::ConversionOptions,
     ) -> Result<String> {
+        let base_spans = self.extract_spans(page_index)?;
+        self.assemble_text_from_spans(page_index, base_spans, options)
+    }
+
+    fn assemble_text_from_spans(
+        &self,
+        page_index: usize,
+        base_spans: Vec<crate::layout::TextSpan>,
+        options: &crate::converters::ConversionOptions,
+    ) -> Result<String> {
         self.require_authenticated()?;
 
-        let mut base_spans = self.extract_spans(page_index)?;
+        let mut base_spans = base_spans;
 
         // Drop spans that fall inside any caller-specified exclusion region.
         // This runs before the structure-tree and table pipelines so that
@@ -7989,7 +7999,26 @@ impl PdfDocument {
     /// # }
     /// ```
     pub fn extract_spans(&self, page_index: usize) -> Result<Vec<crate::layout::TextSpan>> {
-        let mut spans = self.extract_spans_raw(page_index)?;
+        let spans = self.extract_spans_raw(page_index)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn extract_spans_filtered(
+        &self,
+        page_index: usize,
+        excluded_layers: HashSet<String>,
+        excluded_inks: HashSet<String>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let spans = self.extract_spans_raw_filtered(page_index, excluded_layers, excluded_inks)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn postprocess_spans(
+        &self,
+        page_index: usize,
+        raw_spans: Vec<crate::layout::TextSpan>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let mut spans = raw_spans;
 
         // Drop spans whose bbox lies entirely outside the page's MediaBox.
         // PDFs that reuse one big Form XObject across pages (ExpertPdf and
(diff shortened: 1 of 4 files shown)


yfedoseev:
Read through this — thanks for the careful work, especially the bug-1 / bug-2 cache-correctness tests that are deliberately constructed against the pathological paths (`tests/test_ocg_ink_filtering.rs:1450-1531`). A few things I'd like to talk through before merging.

**One behaviour that surprised me**

`Python extract_text(page, region=..., exclude_layers=...)` at `src/python.rs:252-268` routes through `extract_chars_filtered`, applies `filter_by_rect`, and joins `.char` values into a string with no reading-order sort, no whitespace insertion, no line breaks. The docstring three lines above (`:234-238`) promises "the same full text assembly pipeline is used." On the region+filter combination it doesn't. Could we either route through `extract_text_filtered` then post-filter on spans, or trim the docstring to match the reality?

**Smaller things**

- `get_page_inks` (`src/document.rs:9043-9075`) walks page-level `/Resources /ColorSpace` only — XObject-local color spaces aren't enumerated. The test codifies this as expected, but a UI calling `get_page_inks` to populate a picker will silently miss spot inks defined inside Form XObjects. A doc-string warning would help; or descend into XObject resources.
- The happy path of `get_page_inks` (page own ColorSpace has `/Separation`/`/DeviceN`) isn't covered — only the negative-case (XObject-local) test exists. One more test row would close that gap.
- API shape: `extract_text_filtered(page, HashSet<String>, HashSet<String>)` is two positional `HashSet`s. The moment a third filter dimension shows up, this gets awkward. Worth folding into an `ExtractionFilters { excluded_layers, excluded_inks, … }` (or extending `ConversionOptions`) before this surface gets pinned in 1.0. Not blocking, but cheaper to do now.

**Things I noticed but consider follow-up, not this PR**

- Form XObject `/OC` entry on the dict (§8.11.4.4) isn't honoured — only BMC tags inside the stream are checked. A layered XObject referenced via `Do` won't be filtered.
- `/All` and `/None` Separation special names (§8.6.6.4) — rare in practice. File these as separate issues?

The opt-in design (empty filter set → short-circuits to the unfiltered path), the use of the existing marked-content stack for OCMD, and the absence of `unwrap`/`expect`/`panic!` in the new library code all read very nicely. Once the Python region+filter behaviour matches the docstring I'm comfortable merging this.

RayVR:
Thanks, pushed e1ef810 addressing the must-fix items:

  - Region + filters now uses the full pipeline. New extract_text_filtered_in_rect composes extract_spans_filtered with
  assemble_text_from_spans, and Python's extract_text routes through it. The new test test_extract_text_filtered_in_rect_uses_full_pipeline
  constructs a PDF with multiple text runs on two rows and asserts the output contains whitespace between same-row runs and a newline
  between rows — catches the original char-stream regression.
  - get_page_inks docstring now warns that XObject-local color spaces aren't enumerated.
  - Happy-path test added (test_get_page_inks_happy_path) covering page-level /Separation and /DeviceN declarations.

  On the "not blocking" items:
  - ExtractionFilters / ConversionOptions: agreed it's cleaner. Holding off here since extract_text_filtered and extract_chars_filtered are
  public API and folding into a struct touches both Rust and Python surfaces. Happy to do it as a follow-up once we agree on the shape.
  - Form XObject /OC dict (§8.11.4.4) and /All//None (§8.6.6.4): filing as separate issues.

yfedoseev (approved):
LGTM, addressed; merging when CI is green

yfedoseev:
🎉 Thank you, @RayVR! OCG ink filtering + the XObject cache-poisoning fix are **picked for v0.3.57** — merged into the `release/v0.3.57` branch and shipping with the next release.

Small integration fix-up: the `[`extract_text_filtered`]` / `[`extract_text_in_rect`]` doc links needed `Self::` qualification to pass `cargo doc -D warnings`. All 13 `test_ocg_ink_filtering` cases are green. Credited in the CHANGELOG. 🙏

yfedoseev:
🎉 Shipped in **[v0.3.57](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.57)**! OCG ink filtering for text extraction (`extract_text_filtered`) — plus the three latent OCG/cache bugs you caught — are in `main`. Thanks @RayVR, credited in the CHANGELOG + release notes. 🙏

PR #604, opened 2026-05-28
feat: render ocg filtering

RayVR:
## Description

Adds OCG (Optional Content Group) layer filtering to the PDF rendering pipeline. Callers can pass `excluded_layers` to render with specific layers suppressed — content within `/OC` marked-content scopes referencing those OCGs (and annotations with `/OC` entries) is omitted from the rendered output.

The renderer also respects the PDF's default OCG visibility configuration (`/OCProperties/D`) per ISO 32000-1 §8.11.4 — OCGs that are off in the default state are hidden during rendering even when the caller passes no explicit exclusions. This matches what a PDF viewer would display by default.

This is the rendering counterpart to the OCG layer filtering already supported in text extraction. The primary use case is prepress workflows where a single PDF carries multiple layers (artwork, dieline, varnish, watermark, "DRAFT" stamps) and downstream consumers need a specific subset rendered.

Also extracts a shared `optional_content` module so the text extractor and renderer use one canonical implementation of OCG/OCMD resolution, eliminating duplication and a non-ASCII-name decoding regression that existed in the renderer's earlier inline copy.

## Type of Change

- [x] Code refactoring
- [x] Tests

## Related Issues

Builds on `feature/ocg-ink-filtering-and-cache-fix` (OCG text extraction). No specific issue.

### New shared module: `src/optional_content.rs`

Both the text extractor and renderer previously had near-duplicate OCG/OCMD resolution code (the renderer's copy was missing UTF-16LE BOM and PDFDocEncoding fallback in its text-string decoder — a silent correctness regression). Consolidated into:

- `compute_default_off_ocgs(doc)` — reads `/OCProperties/D` (`BaseState`, `/ON`, `/OFF`) and returns OCG names off in the default state
- `decode_pdf_text_string(bytes: &[u8]) -> String` — UTF-16BE/LE BOM, UTF-8, PDFDocEncoding fallback
- `ocg_name_is_excluded(name_obj, excluded)` — handles `Name` and `String` OCG names
- `check_ocg_excluded(props_dict, doc, excluded)` — direct OCG and OCMD dispatch
- `resolve_bdc_properties(properties, resources, doc)` — handles inline dicts and `/Resources/Properties` name references, with indirect-reference resolution at every level
- `annotation_is_excluded(annot_dict, doc, excluded)` — for annotation `/OC` entries
- `OcmdPolicy` enum + `ocmd_is_hidden()` — full `/P` policy implementation (`AnyOn`, `AllOn`, `AnyOff`, `AllOff`)
- `evaluate_visibility_expression()` — full `/VE` boolean expression evaluator (`/And`, `/Or`, `/Not`, nested expressions, OCG references) with bounded recursion depth

### Renderer changes (`src/rendering/page_renderer.rs`)

- `RenderOptions.excluded_layers: HashSet<String>` field
- Per-render snapshot combines `compute_default_off_ocgs(doc)` with caller's `excluded_layers` — the effective "off" set respects both the PDF's default state and the caller's overrides
- BDC/BMC/EMC tracking via a `Vec<bool>` stack (correctly handles nested mixed-exclusion scopes — a flat depth counter would be wrong for `BDC excluded BDC not_excluded EMC EMC`)
- Suppression applied at all paint sites: paths (`S`, `s`, `f`, `F`, `f*`, `B`, `B*`, `b`, `b*`, `n`), text (`Tj`, `TJ`, `'`, `"`), XObjects (`Do`), shadings (`sh`), and clip operators (`W`, `W*`)
- Clip operators are no-ops inside excluded scopes (prevents clip leakage when `W n` appears without surrounding `q/Q`)
- Text advance is computed and applied even when suppressed, via new `TextRasterizer::measure_text` / `measure_tj_array` — visible text following a suppressed run paints at the correct X
- Annotation `/OC` consulted in `render_annotations` — watermark annotations tagged with their OCG are correctly filtered
- Form XObject recursion preserves the marked-content stack scope
- `Arc<HashSet<String>>` snapshot eliminates per-recursion clone of the exclusion set

### Python bindings (`src/python.rs`)

- `excluded_layers: Optional[List[str]]` keyword argument on `PdfDocument.render_page`, `PdfDocument.render_page_fit`, and `Page.render`

### FFI (`src/ffi.rs`)

- New `pdf_render_page_with_options_ex` entry point accepting UTF-8 layer name strings
- Original ABI preserved for backward compatibility

## Testing

- [x] I have run `cargo fmt`

22 tests in `tests/test_render_ocg_filtering.rs`, all asserting on actual pixel data from synthetic PDFs:

- Rectangle visibility with/without filter
- Unrelated layer exclusion produces identical output (pipeline parity)
- OCMD-referenced content
- Form XObject containing OCG-tagged content (scope crossing)
- Empty exclusion set short-circuits cleanly
- Nested mixed-exclusion BDC scopes (stack correctness)
- Clip path inside excluded scope doesn't leak
- Annotation with `/OC` is filtered
- Text positioning preserved when intermediate run is suppressed
- Non-ASCII OCG `/Name` stored as UTF-16LE BOM PDF string
- All four OCMD `/P` policies (`AnyOn`, `AllOn`, `AnyOff`, `AllOff`)
- All three `/VE` operators (`/Not`, `/And`, `/Or`) plus nested `[/And refA [/Not refB]]`
- Default-off OCG (`/D/OFF`) hidden with empty user exclusions

Full test suite: 5,281 lib tests + integration tests pass. Clippy clean.

- [ ] Python tests pass - No - pre-existing failures

Existing `extract_text` / `extract_chars` kwargs (`exclude_layers`) were already in place from the prior PR — this PR mirrors that surface on `render_page` and `render_page_fit`.

## Documentation

Doc comments on all new public APIs and the shared `optional_content` module. `OcmdPolicy` enum documents each variant's semantics with PDF spec §8.11.2.2 references. `evaluate_visibility_expression` references §8.11.2.4. `compute_default_off_ocgs` references §8.11.4.

## Checklist

01_baseline
02_no_background
03_no_annotations
04_production_view

Each PNG was rendered at 200 DPI from the same synthetic PDF.
The PDF has three OCG layers:
  Background   — light-blue rectangle covering most of the page
  Annotations  — red 'DRAFT - DO NOT USE' stamp
  Watermark    — large gray 'CONFIDENTIAL' text
Always-visible content: black title 'Sample Document' and body text.

Files:
  01_baseline.png  (205 KB)
    Baseline: all layers visible (Background, Annotations, Watermark)
  02_no_background.png  (190 KB)
    No Background: light-blue rect hidden; title, body, DRAFT stamp and CONFIDENTIAL still visible
  03_no_annotations.png  (173 KB)
    No Annotations: red DRAFT stamp hidden; everything else visible
  04_production_view.png  (147 KB)
    Production view: both DRAFT stamp and CONFIDENTIAL watermark hidden (real-world prepress use case)

## Additional Notes

**Default OCG state respected (spec-correct).** Per ISO 32000-1 §8.11.4, OCGs have a default visibility set by `/OCProperties/D/BaseState`, `/D/ON`, and `/D/OFF`. The renderer reads these on every render and treats OCGs that are off in the default config as hidden, even when the caller passes empty `excluded_layers`. Caller exclusions compose with this default state (additional offs). `/D/BaseState /Unchanged` is treated as `/ON` since libraries have no prior viewer state to preserve.

**Forced inclusion not supported.** The current API only supports excluding layers. A PDF with `/D/OFF [ref]` will render that OCG hidden, with no way to force it visible. If that's needed, an additional `included_layers` parameter can be added in a follow-up.

**Shared module benefit.** Extracting `optional_content` resolved a latent silent bug — the renderer's previous inline copy of `decode_pdf_text_string` was missing UTF-16LE BOM and PDFDocEncoding fallback, so OCG names stored as non-ASCII PDF strings would silently fail to match. The shared module fixes this for both text extraction and rendering simultaneously.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -8652,9 +8652,19 @@ impl PdfDocument {
 
     /// Extract text from a page, excluding content from specified layers and inks.
     ///
-    /// Uses a simplified text assembly pipeline (no structure-tree ordering or
-    /// table detection). For most layer/ink filtering use cases this is
-    /// sufficient since the caller is performing targeted extraction.
+    /// # Text assembly differences
+    ///
+    /// This method uses a simplified text assembly pipeline compared to
+    /// [`extract_text`](Self::extract_text): spans are sorted by row-aware reading order and joined
+    /// with line breaks based on Y-gap heuristics. The full pipeline's
+    /// structure-tree ordering, table detection, and column detection are
+    /// **not** applied. This means:
+    /// - Multi-column layouts may interleave columns
+    /// - Tables will not be reconstructed
+    /// - Reading order depends on coordinate sorting, not document structure
+    ///
+    /// For precise spatial control, use [`extract_chars_filtered`](Self::extract_chars_filtered) and assemble
+    /// text yourself, or apply region filtering on top.
     ///
     /// **Ink filtering note:** For DeviceN color spaces, text is suppressed if
     /// ANY ink in the DeviceN array matches an excluded ink name. Tint values
--- a/src/document.rs
+++ b/src/document.rs
@@ -4694,9 +4694,19 @@ impl PdfDocument {
         page_index: usize,
         options: &crate::converters::ConversionOptions,
     ) -> Result<String> {
+        let base_spans = self.extract_spans(page_index)?;
+        self.assemble_text_from_spans(page_index, base_spans, options)
+    }
+
+    fn assemble_text_from_spans(
+        &self,
+        page_index: usize,
+        base_spans: Vec<crate::layout::TextSpan>,
+        options: &crate::converters::ConversionOptions,
+    ) -> Result<String> {
         self.require_authenticated()?;
 
-        let mut base_spans = self.extract_spans(page_index)?;
+        let mut base_spans = base_spans;
 
         // Drop spans that fall inside any caller-specified exclusion region.
         // This runs before the structure-tree and table pipelines so that
@@ -7989,7 +7999,26 @@ impl PdfDocument {
     /// # }
     /// ```
     pub fn extract_spans(&self, page_index: usize) -> Result<Vec<crate::layout::TextSpan>> {
-        let mut spans = self.extract_spans_raw(page_index)?;
+        let spans = self.extract_spans_raw(page_index)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn extract_spans_filtered(
+        &self,
+        page_index: usize,
+        excluded_layers: HashSet<String>,
+        excluded_inks: HashSet<String>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let spans = self.extract_spans_raw_filtered(page_index, excluded_layers, excluded_inks)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn postprocess_spans(
+        &self,
+        page_index: usize,
+        raw_spans: Vec<crate::layout::TextSpan>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let mut spans = raw_spans;
 
         // Drop spans whose bbox lies entirely outside the page's MediaBox.
         // PDFs that reuse one big Form XObject across pages (ExpertPdf and
@@ -8652,19 +8681,9 @@ impl PdfDocument {
 
     /// Extract text from a page, excluding content from specified layers and inks.
     ///
-    /// # Text assembly differences
-    ///
-    /// This method uses a simplified text assembly pipeline compared to
-    /// [`extract_text`](Self::extract_text): spans are sorted by row-aware reading order and joined
-    /// with line breaks based on Y-gap heuristics. The full pipeline's
-    /// structure-tree ordering, table detection, and column detection are
-    /// **not** applied. This means:
-    /// - Multi-column layouts may interleave columns
-    /// - Tables will not be reconstructed
-    /// - Reading order depends on coordinate sorting, not document structure
-    ///
-    /// For precise spatial control, use [`extract_chars_filtered`](Self::extract_chars_filtered) and assemble
-    /// text yourself, or apply region filtering on top.
+    /// Uses the same full text assembly pipeline as [`extract_text`](Self::extract_text)
+    /// (structure-tree ordering, table detection, column detection), but with
+    /// layer/ink-excluded spans removed before assembly.
     ///
     /// **Ink filtering note:** For DeviceN color spaces, text is suppressed if
     /// ANY ink in the DeviceN array matches an excluded ink name. Tint values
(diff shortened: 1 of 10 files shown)


yfedoseev:
Heavy PR but a pleasant one to read — the new `src/optional_content.rs` as a shared OCG/OCMD resolver is exactly the right move, and the refactor in `src/extractors/text.rs:2547-2575` to delegate to it (rather than duplicate against #600) is the kind of cross-PR coordination I love to see. Pixel-sampling tests rather than snapshot PNGs is also smart — keeps the test suite robust to renderer drift.

**One bug I'd want fixed before this merges**

`PdfDocument::get_layers` at `src/document.rs:9050` independently decodes OCG `/Name` with `String::from_utf8(s.clone()).ok()` instead of calling `crate::optional_content::decode_pdf_text_string` — the shared decoder this PR adds. So a PDF with UTF-16BE/LE-encoded layer names (the exact case the shared decoder exists for) silently drops those layers from the list returned to Python callers, even though the renderer would correctly filter on the same name if the caller hard-codes it. One-line fix; happy to take a regression test with it.

**One naming consistency thing across #600 and #604**

#600 added Python kwarg `extract_text(exclude_layers=…)`. This PR adds `render_page(excluded_layers=…)`. They're inconsistent (verb vs past participle), so a user can't pass the same dict to both. Could we settle on one — I'm slightly biased toward `excluded_layers` since it reads as state — and alias the other for back-compat?

**Two smaller correctness things**

- OCMD `/VE` operand-validation in `src/optional_content.rs:343` region: `/Not` with no operand evaluates to `!true = false` (hidden) — fail-closed inside a module that otherwise fails-open. Either treat missing-operand `/Not` as `true` for consistency or document the asymmetry.
- `cm` issued inside a hidden `BDC` currently persists past `EMC`. In practice such transforms are usually wrapped in `q`/`Q`, but I'd like to either lock the current behaviour with a test or fix it (whichever you think is right).

**Follow-up, not this PR**

- `compute_default_off_ocgs` (`src/optional_content.rs:25`) walks the catalog per page; worth memoising on `PdfDocument` for multi-page renders.
- The `static OnceLock<HashSet<String>>` empty-set trick at `src/rendering/page_renderer.rs:438-440` is clever but reads heavy — an `Option`-threaded reference would be cleaner.
- PR body says "full `/OCProperties/D` chain"; the parts the renderer needs are there, but `/Configs`, `/Order`, `/RBGroups`, `/Locked`, `/AS` are out-of-scope. Worth a one-line clarification in the body.
- Add a test where `BMC` (no dict) is nested inside a hidden `BDC /OC` — current tests cover dict-BDC nesting only.

What's working well: 22 fixture tests + 13 unit tests, all four OCMD `/P` policies implemented and exercised, `/VE` with bounded depth (16), gating at operator-handler level (no per-glyph cost), `Arc<HashSet<String>>` snapshot avoids per-recursion clones, mixed-exclusion BDC nesting via `Vec<bool>` stack rather than a flat counter, `q`/`Q` correctly not gated. Once the `get_layers` decoder fix lands and the kwarg name is settled, I'm comfortable merging.

yfedoseev:
Reviewed and approved ✅ — this will be part of the **v0.3.59** release.

The visibility algorithm holds up against §8.11: OCMD `/P` policy + `/VE` expressions (§8.11.2.2), the hidden-content text-advance rule (§8.11.3), and BaseState/ON/OFF default config (§8.11.4) are all handled, the shared `optional_content.rs` resolver cleanly de-dupes the extract and render paths (nice coordination with #600), and the no-optional-content path is provably non-regressing. One documentation item I'll add during integration: `render_page` honors the PDF's default config (`/OCProperties/D`, including default-off OCGs) while `extract_text` only acts on caller-supplied `excluded_layers` — defensible per §8.11 NOTE 4, but worth stating explicitly so the two surfaces' behavior is clear. Thanks @RayVR!

yfedoseev:

PR #605, opened 2026-05-28
feat: separation plate rendering

RayVR:
NOTE: This builds on PR #600 

## Description

Adds a separation plate rendering API that produces one grayscale image per ink — the standard prepress output format used for color separations, plate-making, and ML pipelines that need per-ink masks. Each pixel value in a plate equals the tint coverage of that ink at that point (0 = no ink, 255 = full coverage), which is the right semantic for downstream consumers (RIPs, QC tools, segmentation models).

The primary use cases this targets:
- **Prepress QC** — per-plate ink coverage analysis, spot-color verification, dieline extraction
- **ML/AI training data** — free per-channel ground-truth masks from packaging/label PDFs (the dieline plate is a clean binary mask of all die-cut paths; varnish plate shows coating regions; each spot color plate isolates that ink)
- **Press workflows** — feeding individual plates to platemaking systems

## Type of Change

- [x] Code refactoring
- [x] Tests

## Related Issues

Builds on `feature/ocg-ink-filtering-and-cache-fix` (ink detection via `get_page_inks`). No specific issue.

### New module: `src/rendering/separation_renderer.rs`

- `SeparationPlate { ink_name, data, width, height }` — public struct representing a single grayscale ink plate
- `render_separations(doc, page, dpi) -> Vec<SeparationPlate>` — all plates for a page
- `render_separation(doc, page, ink_name, dpi) -> SeparationPlate` — single targeted plate

Implementation approach: walks the page operator stream once per requested ink, tracking the current fill/stroke color space alongside the graphics-state stack, and rasterizes paths/text into a single-channel grayscale buffer via tiny-skia. Form XObjects recurse with the caller's color state inherited per §8.10.1. Clipping (`W`, `W*`), graphics state save/restore (`q`, `Q`), and ExtGState alpha are applied.

### Color routing per PDF spec ISO 32000-1 §8.6

| Color space | Routing |
|---|---|
| DeviceCMYK (`k`, `K`) | C/M/Y/K components route to their respective plates |
| Separation `[/Separation /InkName /Alt /Tint]` + `scn` | Tint routes to the named ink's plate |
| Separation `/All` (§8.6.6.4) | Tint paints to every output plate (CMYK + named spots) — used for registration marks, crop marks |
| Separation `/None` (§8.6.6.4) | Paints nothing; never materializes a plate |
| DeviceN `[/DeviceN [/Ink1 ...] /Alt /Tint]` + `scn` | Per-component tints route to per-ink plates |
| DeviceRGB, DeviceGray | Skipped (these are display color spaces) |
| ICCBased | N=4 routes as CMYK (with `/N` stream-dict probe + heuristic fallback); other arities skipped |
| DefaultCMYK / DefaultRGB / DefaultGray (§8.6.5.6) | Lookup resolved; `DefaultCMYK` content routing partially implemented (see foundation tests) |

### Text rendering to plates

Text glyphs are routed through the existing `TextRasterizer` with a faked-grayscale `GraphicsState` (`fill_color_rgb = (tint, tint, tint)`) so the R-channel rasterization carries the tint at full anti-aliasing fidelity. BT/ET, Tf, Td/TD/Tm/T*, Tj/TJ/'/" all handled. Text rendering mode 3 (invisible) advances the text matrix without painting.

### Spec-correct initial state

- `cs`/`CS` resets fill/stroke components to the space's initial value per §8.6.4.2 (DeviceCMYK = [0,0,0,1], Separation = [1.0], DeviceN = [1.0;n], etc.)
- Form XObject recursion inherits the caller's complete color state per §8.10.1

### Performance

- CMYK plates always emitted; unused ones short-circuit via `collect_referenced_inks` (no operator walk for an empty plate)
- Per-plate dispatch avoids re-parsing the page stream for each ink

### Shared infrastructure: `src/rendering/ext_gstate.rs`

Extracted `ParsedExtGState` and its parser from `page_renderer.rs` so both renderers share one implementation (was previously `pub(crate)` reaching across modules).

### Python bindings (`src/python.rs`)

- `PdfDocument.render_separations(page, dpi=...)` returns `List[SeparationPlate]`
- `PdfDocument.render_separation(page, ink_name, dpi=...)` returns single plate
- `SeparationPlate` namedtuple in `python/pdf_oxide/__init__.py`

### `/All` and `/None` handling in `src/document.rs`

`get_page_inks` filters reserved colorant names so they never appear as ink-name strings to callers.

## Testing

- [x] I have run `cargo fmt`

20 passing tests in `tests/test_separation_rendering.rs` covering:
- Separation ink tint appears in correct plate (linear ramp at 0%, 50%, 100%)
- CMYK content routes to C/M/Y/K plates with correct intensities
- CMYK content nested inside Form XObjects detected
- DeviceN per-component routing
- DeviceRGB / DeviceGray correctly skipped (not converted)
- Anti-aliased edges produce intermediate plate values
- Overlapping rects (last-writer-wins per opaque painting)
- `q`/`Q` saves and restores color state
- Page rotation 90/180/270
- Spot-color text appears on its plate
- CMYK text routes to process plates
- `/All` paints to every plate (registration-mark scenario)
- `/None` paints nothing
- Plus 7 more covering edge cases

12 `#[ignore]`'d foundation tests documenting remaining spec gaps. Each names the ISO 32000-1 section and turns into a passing regression test once the underlying gap is closed:

| Test | Spec | Gap |
|---|---|---|
| `test_overprint_preserves_underlying_ink` | §11.7 | `OP`/`op`/`OPM` overprint mode |
| `test_render_mode_3_invisible_text_no_plate_pixels` | §9.3.6 | Tr 3 may leak |
| `test_stroke_mode_text_uses_stroke_color` | §9.3.6 | Tr 1/2/5/6 use fill color |
| `test_default_rgb_remap` | §8.6.5.6 | DefaultRGB resolution |
| `test_default_gray_remap` | §8.6.5.6 | DefaultGray resolution |
| `default_cmyk_remap_redirects_to_separation` | §8.6.5.6 | DefaultCMYK content routing through override |
| `test_form_xobject_inherits_line_width` | §8.10.1 | Full GS inheritance (line width, dash, font) |
| `test_nchannel_process_attributes` | §8.6.6.5 | NChannel `/Process` per-component blending |
| `test_out_of_range_tint_values_documented` | §8.6.6 | Silent tint clamping |
| `test_annotation_appearance_streams_contribute_to_plates` | §12.5.4 | Annotation `/AP` not rendered |
| `test_pattern_color_space` | §8.7 | Patterns and `sh` operator |
| `test_inline_image_devicen` | §8.9 | Inline image DeviceN samples |

Verified each ignored test fails when un-ignored (`cargo test -- --ignored` produces 12/12 failures with the expected messages).

Full lib suite: 5,269 lib tests pass. Clippy clean.

- [ ] Python tests pass

`render_separations` and `render_separation` exposed with the same DPI/page argument shape as the existing `render_page`. `SeparationPlate` namedtuple gives Python callers direct access to the grayscale bytes, width, height, and ink name.

## Documentation

Module-level rustdoc enumerates dropped features (patterns, shadings, inline images, annotations) so callers know exactly what to expect. Color routing logic documented per spec section.

## Checklist

## Additional Notes

**Display convention.** The `SeparationPlate.data` returned by the API is ML-friendly: pixel value = ink coverage (0 = no ink, 255 = full tint). For human display in prepress workflows (dark = ink on white paper), callers should invert: `display_pixel = 255 - data[i]`. 

**Overprint not simulated.** Per §11.7, real prepress uses overprint (`OP`/`op`/`OPM`) to control whether a fresh paint replaces or adds to underlying ink. The current renderer is always last-writer-wins. This matters most when CMYK(0,0,0,0) "paper-white" text is drawn over a spot-color bar — in real production the white text "knocks out" the bar, but our renderer keeps the bar's full coverage. Foundation test `test_overprint_preserves_underlying_ink` is in place; flagged as a follow-up because correct overprint requires cross-plate compositing logic that's out of scope here.

**`/All` semantics.** Content drawn with `/Separation /All` (the spec's reserved colorant for "paint to every output plate") routes to every CMYK and spot plate simultaneously. This is exactly the encoding real PDFs use for registration marks and crop marks — a single mark on the page appears identically on every plate so the press operator can align them. It does not duplicate the entire page content per plate; only the specific drawing operations issued while `/All` is the active color space.

**Foundation-test pattern.** The 12 `#[ignore]`'d tests serve as both documentation and regression-test scaffolding. Each fails when un-ignored (verifying the gap is real, not a no-op test) and names the ISO 32000-1 section so the next iteration knows exactly what spec text to implement against. Removing the `#[ignore]` line after fixing the gap converts each one into a passing regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to PDFOxide are documented here.
 
+## [Unreleased]
+
+### Added
+- **Separation plate rendering** — new `render_separations(page, dpi)` and
+  `render_separation(page, ink_name, dpi)` APIs (Rust + Python) that emit
+  one grayscale image per ink. Pixel value equals ink coverage (0 = no ink,
+  255 = full tint). Routes DeviceCMYK, Separation, and DeviceN content
+  per ISO 32000-1 §8.6; handles the reserved colorant names `/All` and
+  `/None` per §8.6.6.4 (registration / crop marks correctly appear on
+  every plate). New `SeparationPlate` namedtuple in Python.
+
 ## [0.3.55] - 2026-05-25
 
 > Ruby + PHP language bindings + multi-line heading reading-order fix
--- a/python/pdf_oxide/__init__.py
+++ b/python/pdf_oxide/__init__.py
@@ -110,6 +110,27 @@ class RenderedPixmap(NamedTuple):
     height: int
 
 
+class SeparationPlate(NamedTuple):
+    """A single ink separation plate rendered as grayscale.
+
+    Pixel intensity (0-255) represents the tint percentage of one ink at
+    each point. Used in prepress workflows, ink coverage analysis, and ML
+    pipelines that process packaging/label PDFs.
+
+    Attributes:
+        ink_name (str): Ink name (e.g., "Cyan", "PANTONE 185 C", "Dieline").
+        data (bytes): Grayscale pixels, row-major, top-left origin.
+            ``len(data) == width * height``. 0 = no ink, 255 = full tint.
+        width (int): Image width in pixels.
+        height (int): Image height in pixels.
+    """
+
+    ink_name: str
+    data: bytes
+    width: int
+    height: int
+
+
 from ._async import (  # noqa: E402
     AsyncOfficeConverter,
     AsyncPdf,
@@ -182,6 +203,7 @@ from .pdf_oxide import (  # noqa: E402
 
 __all__ = [
     "RenderedPixmap",
+    "SeparationPlate",
     "PdfDocument",
     "AsyncPdfDocument",
     "AsyncPdf",
--- a/python/pdf_oxide/__init__.py
+++ b/python/pdf_oxide/__init__.py
@@ -117,6 +117,11 @@ class SeparationPlate(NamedTuple):
     each point. Used in prepress workflows, ink coverage analysis, and ML
     pipelines that process packaging/label PDFs.
 
+    The pixel convention is ML/QC-friendly: ``value == ink coverage``.
+    0 means no ink, 255 means full tint coverage. To display the plate as
+    black ink on white paper (prepress viewer convention), invert before
+    showing: ``display = 255 - value``.
+
     Attributes:
         ink_name (str): Ink name (e.g., "Cyan", "PANTONE 185 C", "Dieline").
         data (bytes): Grayscale pixels, row-major, top-left origin.
--- a/src/document.rs
+++ b/src/document.rs
@@ -8652,9 +8652,19 @@ impl PdfDocument {
 
     /// Extract text from a page, excluding content from specified layers and inks.
     ///
-    /// Uses a simplified text assembly pipeline (no structure-tree ordering or
-    /// table detection). For most layer/ink filtering use cases this is
-    /// sufficient since the caller is performing targeted extraction.
+    /// # Text assembly differences
+    ///
+    /// This method uses a simplified text assembly pipeline compared to
+    /// [`extract_text`](Self::extract_text): spans are sorted by row-aware reading order and joined
+    /// with line breaks based on Y-gap heuristics. The full pipeline's
+    /// structure-tree ordering, table detection, and column detection are
+    /// **not** applied. This means:
+    /// - Multi-column layouts may interleave columns
+    /// - Tables will not be reconstructed
+    /// - Reading order depends on coordinate sorting, not document structure
+    ///
+    /// For precise spatial control, use [`extract_chars_filtered`](Self::extract_chars_filtered) and assemble
+    /// text yourself, or apply region filtering on top.
     ///
     /// **Ink filtering note:** For DeviceN color spaces, text is suppressed if
     /// ANY ink in the DeviceN array matches an excluded ink name. Tint values
--- a/src/document.rs
+++ b/src/document.rs
@@ -4694,9 +4694,19 @@ impl PdfDocument {
         page_index: usize,
         options: &crate::converters::ConversionOptions,
     ) -> Result<String> {
+        let base_spans = self.extract_spans(page_index)?;
+        self.assemble_text_from_spans(page_index, base_spans, options)
+    }
+
+    fn assemble_text_from_spans(
+        &self,
+        page_index: usize,
+        base_spans: Vec<crate::layout::TextSpan>,
+        options: &crate::converters::ConversionOptions,
+    ) -> Result<String> {
         self.require_authenticated()?;
 
-        let mut base_spans = self.extract_spans(page_index)?;
+        let mut base_spans = base_spans;
 
         // Drop spans that fall inside any caller-specified exclusion region.
         // This runs before the structure-tree and table pipelines so that
@@ -7989,7 +7999,26 @@ impl PdfDocument {
     /// # }
     /// ```
     pub fn extract_spans(&self, page_index: usize) -> Result<Vec<crate::layout::TextSpan>> {
-        let mut spans = self.extract_spans_raw(page_index)?;
+        let spans = self.extract_spans_raw(page_index)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn extract_spans_filtered(
+        &self,
+        page_index: usize,
+        excluded_layers: HashSet<String>,
+        excluded_inks: HashSet<String>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let spans = self.extract_spans_raw_filtered(page_index, excluded_layers, excluded_inks)?;
+        self.postprocess_spans(page_index, spans)
+    }
+
+    fn postprocess_spans(
+        &self,
+        page_index: usize,
+        raw_spans: Vec<crate::layout::TextSpan>,
+    ) -> Result<Vec<crate::layout::TextSpan>> {
+        let mut spans = raw_spans;
 
         // Drop spans whose bbox lies entirely outside the page's MediaBox.
         // PDFs that reuse one big Form XObject across pages (ExpertPdf and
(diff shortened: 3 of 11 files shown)


yfedoseev:
Read through this carefully — it's a genuinely strong first cut at a hard problem. The module-level rustdoc enumerating dropped content classes (`src/rendering/separation_renderer.rs:23-43`) is exemplary, the `/All` and `/None` handling per §8.6.6.4 is threaded through both `get_page_inks` (`src/document.rs:9102-9123`) and the runtime tint router (`separation_renderer.rs:641-661`), and the faked-grayscale `TextRasterizer` trick works without premultiplication or gamma weirdness because `create_fill_paint` (`src/rendering/mod.rs:47-62`) uses straight alpha. The 12 `#[ignore]`'d foundation tests citing ISO 32000-1 sections per gap is exactly the right pattern for honest scoping. Form XObject color-state inheritance per §8.10.1 lands correctly at `separation_renderer.rs:1414-1435`.

A few things to look at before merging — none of them blockers, all four could ship as follow-up issues if you prefer:

#### 1. The "parse once, dispatch to N plates" claim isn't what the code does

`render_separations` (`separation_renderer.rs:90-115`) loops per ink, and each iteration calls `render_single_separation`, which re-runs `parse_content_stream(&content_data)` at `:379` and re-parses every Form XObject inside `execute_separation_operators`. For a CMYK page with one spot ink that's 5 full parses + 5 full operator walks. The `collect_referenced_inks` short-circuit on `:103` only skips work for plates whose ink is *never* referenced; it doesn't deduplicate parsing for plates that *are* referenced. Either reword the PR description ("CMYK plates always emitted; unused plates short-circuit; referenced plates re-walk independently") or restructure `execute_separation_operators` to take `target_inks: &[&str]` and dispatch to N pixmaps in one walk. The latter is the real perf win and would make ExtGState/font caches per-page instead of per-(page × ink).

#### 2. ICCBased fallback is more eager than the docstring claims

`classify_resolved` (`separation_renderer.rs:519-539`) returns `IccCmyk` when the stream's `/N` cannot be read **or is anything other than 1/3/4**. So a 2-component or 5-component ICC profile (rare but legal — CIELAB, hexachrome) silently maps to CMYK and produces fake plate values. Make the unknown-N branch return `ResolvedSpace::Unknown`. The comment on `:537` says "caller decides via component vector length" but `tint_for_ink` doesn't actually inspect the vector when given `IccCmyk` — it just slices the first four components.

#### 3. Text fill_alpha / blend_mode silently clobbered

Lines `1494-1498` of `render_text_to_plate` set `faux.fill_alpha = 1.0` and `faux.blend_mode = "Normal"` on the cloned `GraphicsState`, intentionally ignoring any ExtGState `ca`/`BM` for text. Path fills do the same via `fill_separation`. This is the right semantics for "tint = ink coverage, ignore transparency" — but it isn't mentioned in the Limitations section. One bullet would prevent a caller who renders 50%-alpha spot text from being surprised when the plate shows 100% tint.

#### 4. Display-inversion convention only in the PR body

The `display_pixel = 255 - data[i]` note is in the PR description but not in `SeparationPlate` rustdoc (`separation_renderer.rs:67-78`) or the Python `SeparationPlate` docstring (`python/pdf_oxide/__init__.py:113-128`). Both currently say "255 = full tint", which is correct but ambiguous. A one-line "to display as black ink on white paper, render `255 - value`" prevents inverted-looking output downstream.

### Smaller / follow-up

- **No CHANGELOG entry.** Checkbox in the PR body is unchecked and the diff doesn't add one. This is user-visible (new Python class + two new methods); please add one.
- **Python DPI default**: `render_separations` / `render_separation` default to `dpi=150` in Python (`src/python.rs:638, 690`) while the Rust API has no default. Worth a docstring note.
- **`empty_plate_for`** (`separation_renderer.rs:160-174`) always allocates `width × height` bytes even for unreferenced plates. A3 at 300 DPI is ~17 MB of zeroed memory per unused plate, returned to Python as a `bytes` object. Either document or expose a `referenced: bool` flag.
- **`measure_text_advance`** (`:1559-1598`) doesn't honor `word_space`, `horizontal_scaling`, or `text_rise`. It runs in the render-mode-3 / "tint doesn't touch this plate" paths, so the worst case is per-plate text drift; flag in a comment that this is per-plate-approximate, not generally correct.
- **Three parallel stacks** for q/Q color state (`gs_stack` + `color_state_stack` + `clip_stack` at `:803-820`) are easy to desync for a future contributor. Consider folding `SeparationColorState` into `GraphicsState`. Today's `q_restore_preserves_color_state` test passes, so this is maintainability, not correctness.
- **Some ignored foundation tests are bare `unimplemented!()`** (e.g. `test_render_mode_3_invisible_text_no_plate_pixels` at `tests/test_separation_rendering.rs:997`). Converting these to real assertions before merge means `cargo test -- --ignored` produces value-mismatch failures (informative) instead of `unimplemented!()` panics. Others (`test_overprint_preserves_underlying_ink` at `:977`) already build real PDFs — those are the right shape.

### What's working well

- `/All` correctly routes **per-operation** (not per-page), and the test (`test_all_separation_paints_every_plate`) verifies a `/Separation /PANTONE` rect elsewhere on the same page does *not* appear on Cyan. That's the subtle case most implementations get wrong.
- 20 passing tests are real end-to-end PDFs with pixel sampling at math-derived coordinates and bounded value assertions — spot-checked `cmyk_content_appears_in_process_plates` (`:218`), `overlapping_same_ink_last_writer_wins` (`:464`), `form_xobject_inherits_caller_color_state` (`:616`), `devicergb_content_does_not_contribute_to_plates` (`:540`), all assert real values.
- `apply_separation_clip` mask compositing at `:1676-1683` correctly intersects nested clips via multiplicative alpha.
- The `ParsedExtGState` extraction into `src/rendering/ext_gstate.rs` is a clean refactor — the deletion at `page_renderer.rs:2312-2369` matches the new module verbatim, no behavioural change.
- `tint_for_ink` (`:587-679`) cleanly factors the per-space routing logic and is the right place to add overprint, NChannel-Process, and Pattern routing in follow-ups.
- I confirmed #605's `src/document.rs` `/All`/`/None` filter is genuinely new to this PR, not duplicated from #600. Merge order: #600 first, then #605. Cleanly composable, no overlap.

Happy to approve once #600 lands. Up to you whether the four "please address" items happen in-PR or as follow-ups.

yfedoseev (approved):
Verified the four "please address" items + smaller asks in `3a1a3d0` and `9b87a32`:

1. Parse-once dispatch — `execute_separation_operators(target_inks: &[&str], &mut [Pixmap])`, Form XObject recursion threads through; that's the real perf win, not a doc fix.
2. ICCBased `N != 1/3/4` → `Unknown` (`separation_renderer.rs:577-587`).
3. Transparency note in module rustdoc.
4. Display-inversion convention in `SeparationPlate` rustdoc + Python docstring.

CHANGELOG + Python DPI default + real foundation-test bodies (9 ignored, down from 12). 23/23 separation tests green. LGTM, merging after CI.

yfedoseev:
🎉 Thank you, @RayVR! Separation plate rendering is **picked for v0.3.57** — merged into the `release/v0.3.57` branch (on top of #600) and shipping with the next release.

Two tiny clippy fix-ups on integration in `test_separation_rendering.rs`: `manual_range_contains` → `(100..=140).contains(&center)` and `repeat().take()` → `std::iter::repeat_n()`. Separation tests are green (23 passed under the `rendering` feature). Credited in the CHANGELOG. 🙏

yfedoseev:
🎉 Shipped in **[v0.3.57](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.57)** — just released! Your separation-plate rendering (`render_separations` / `render_separation`, per-ink grayscale coverage, §8.6 with `/All`·`/None`) is in `main`. Thanks again @RayVR — credited in the CHANGELOG + release notes. 🙏

PR #606, opened 2026-05-29
fix: Fix Type3 font width scaling: apply FontMatrix[0] instead of hardcoded 1/1000

haberman:
## Description

Adds proper font width scaling for Type3 fonts.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Changes Made

- Replaces the hardcoded fs_factor = font_size / 1000.0 with fs_factor = font_size * font_matrix_a, where font_matrix_a is FontMatrix[0] parsed from the PDF font dictionary (defaults to 0.001 for standard fonts, matching the old behaviour).
- Adds FontInfo::font_matrix_a field and from_dict() parsing.
- Rescales the flags-based `default_width` fallback for Type3 fonts: the 500/550/600 heuristic is calibrated for `font_matrix_a = 0.001`, so when FontMatrix[0] differs (e.g. 1.0 for an identity matrix), the stored value is divided by `font_matrix_a / 0.001` so that callers multiplying by `font_matrix_a` still get the correct em-fraction.

## Testing

- [X] I have run `cargo fmt`

## Documentation

## Checklist


yfedoseev (approved):
**LGTM** — clean fix, spec-aligned, and matches how pdf.js and mupdf implement Type 3 advance. Cross-checked against ISO 32000-1 §9.6.5 + §9.2.4 and the `docs/spec/pdf.md:17905` table — your `FontMatrix[0]` substitution is exactly right.

Approving as-is. A few **nice-to-haves** (not blockers) if you fancy a polish pass, otherwise happy to handle as follow-ups:

**1. Comment slightly overstates the limitation** (`src/extractors/text.rs:6119` and three siblings)

> "Assumes FontMatrix[1] = 0 (no glyph-axis rotation)…"

Per §9.2.4 the spec says *"only the horizontal component of the transformed width shall be used"*, which equals `w × FontMatrix[0]` regardless of `FontMatrix[1]`. The math is correct even with rotation/shear — `b` only contributes to the discarded vertical projection. pdf.js's `evaluator.js` uses the same `fontMatrix[0] * fontSize` and explicitly handles rotated FontMatrices through it. So the assumption isn't really needed; consider rewording as a spec-compliance note rather than a limitation.

**2. Test silent-pass risk** (`tests/test_type3_fontmatrix_advance.rs:165` and `:202`)

```rust
if ch.is_none() {
    return;
}
```

If `extract_chars` ever stops surfacing the `d0` glyph (refactor, encoding bug, etc.), both tests turn into "pass by skipping" — they'd be green even with the fix reverted. Worth tightening to `.expect("Type3 'A' should extract")` so the failure mode is loud. If `d0` is genuinely too lossy, switching the CharProc to a `d1` with a 1×1 fill path would guarantee a renderable mark.

**3. Unrelated drive-by in `tests/test_multibyte_cid_support.rs:456-466`**

The bfrange overlap fix (`4..50` → `0x80..0xAE`) is a real and good fix, just unrelated to FontMatrix scaling. Either split into a separate commit/PR or call it out in the PR body — otherwise a future bisect on the CMap area will land here confused about why a Type3 PR touched it.

Thanks again — really solid root-cause work and the regression test for the `[0.2 …]` scaling is exactly the right shape.

yfedoseev:
🎉 Shipped in **[v0.3.57](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.57)**! Type 3 font width scaling via `FontMatrix[0]` is in `main`. (Follow-up from a review pass: we also added a guard so a degenerate `/FontMatrix[0]=0` falls back to 0.001 instead of inf/NaN, with a regression test.) Thanks @haberman — credited in the CHANGELOG + release notes. 🙏

haberman:
Ah you just beat me, but happy to let you make the patches. :)

PR #602, opened 2026-05-28
feat: add rendered_advance to TextChar

haberman:
## Description

Adds rendered_advance to TextChar: the distance the text position moves
after rendering this character, including glyph advance width, character
spacing (Tc) and word spacing (Tw for U+0020).

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Fixes: https://github.com/yfedoseev/pdf_oxide/issues/594

Fixes #

## Changes Made

- Adds rendered_advance to TextChar

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All new and existing tests pass locally
- [x] I have run `cargo test --all-features`
- [x] I have run `cargo clippy -- -D warnings`
- [x] I have run `cargo fmt`

## Documentation

- [x] I have updated the documentation (README, docs/, code comments)
- [ ] I have added/updated examples (if applicable)
- [x] I have updated CHANGELOG.md

## Checklist

- [x] My code follows the project's coding guidelines (see CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any misspellings
- [x] The PR title follows conventional commits format (e.g., `feat:`, `fix:`, `docs:`)

Issue #610, opened 2026-05-29
Type0 font with embedded UTF-8 CMap stream (/Encoding) extracts 0 chars — needs variable-width codespace segmentation (u16→u32)


yfedoseev:
## Summary

Type0 / CIDFontType2 fonts whose `/Encoding` is an **embedded CMap stream with a UTF-8 (variable-width) codespace** extract as **0 characters**. Found via the v0.3.57 cross-corpus sweep (pdf.js `issue18117.pdf`, `SourceHanSansTC-Regular`, Adobe-CNS1).

## Root cause

The embedded CMap declares a variable-width UTF-8 codespace:

```
begincodespacerange
<00>       <7F>          # 1 byte
<C080>     <DFBF>        # 2 bytes
<E08080>   <EFBFBF>      # 3 bytes
<F0808080> <F7BFBFBF>    # 4 bytes
endcodespacerange
```

`/CMapName` is `Uni-Utf8-H` (also the Adobe predefined family `UniGB-UTF8-H`, `UniCNS-UTF8-H`, `UniJIS-UTF8-H`, `UniKS-UTF8-H`). In `src/fonts/font_dict.rs` (~L1696) a non-`Adobe-`/non-`Identity` CMap-stream name falls through to `Encoding::Standard("Uni-Utf8-H")` and the embedded codespace is ignored, so `get_byte_mode` (`src/extractors/text.rs`) picks a fixed 1- or 2-byte width and mis-segments the content stream → no ToUnicode match → empty output.

The ToUnicode CMap **is** present and keyed by the same variable-width codes (`<20> <0020>`, `<e38081> <3001>`, …), so only correct **segmentation** is needed — once bytes are grouped per the UTF-8 codespace, the existing ToUnicode path resolves code→Unicode. No cidrange handling required for extraction.

## Why it isn't a small fix

`TextCharIter::Item = (u16, usize)` (`src/extractors/text.rs:2048`) — the char-code type is **u16**, which cannot hold 3- and 4-byte UTF-8 codes (e.g. `0xE38081`). Supporting a UTF-8 codespace therefore requires widening the char code `u16 → u32` across the decode hot path (`get_byte_mode`, `TextCharIter`, `char_to_unicode`, and downstream), which is used by **every** font — so it carries broad regression risk and must ship as a dedicated PR with a full corpus sweep, not a release-branch patch.

## Proposed implementation

[...]

PR #612, opened 2026-05-30
v0.3.57 — Community contributions + extraction-quality sweep

### LINKED ISSUE #553: text-extraction: page-number/header/footer chrome leaks mid-paragraph into body text
## Summary

Page-number chrome (running headers such as `Article number, page 1 of 10`, `A&A proofs: manuscript no. mainArxiv`, and bare page numbers like ` 1 `, ` 2 `) leaks into the body text in `extract_text` output, often **mid-paragraph** because the column-interleave bug (#549) puts the chrome in the middle of a wrapped line.

Additionally, the arXiv stamp `arXiv:2201.00021v3 [astro-ph.GA] 9 Apr 2022` from the LaTeX `\overlay` is emitted in the middle of the body text rather than at the top (where ground truth puts it).

## Reproduction

```python
import pdf_oxide  # 0.3.54
doc = pdf_oxide.PdfDocument("2201.00029.pdf")
print(doc.extract_text(0)[:200])
# '     1 \n\n\n \n\n\n \n\n\n \n\n\n \n\n\n \n\nExploring new techniques...'

doc2 = pdf_oxide.PdfDocument("2201.00021.pdf")
# Mid-paragraph chrome:
#   "...spots are observed: one is close to the head of the\n
#    Article number, page 1 of 10\n
#    A&A proofs: manuscript no. mainArxiv\n
#    cometary ultracompact H ii region C, ..."
```

PDFs: <https://arxiv.org/pdf/2201.00029.pdf> (13 stray standalone-number lines), <https://arxiv.org/pdf/2201.00021.pdf> (23 standalone numbers), <https://arxiv.org/pdf/2201.00200.pdf>.

## Expected

Header / footer / arXiv-stamp content removed from the body stream, or at minimum collated to the top/bottom of the page where the visual layout shows it.

## Actual

Standalone integers and `Article number, page N of M` strings appear on their own lines interspersed with body paragraphs.

## Impact

- `2201.00029`: 13 stray page-number lines in a 17 KB output (≈ 0.5 KB noise). Score 85.4% with this noise; with `to_plain_text_all` (which handles it differently) jumps to **96.1%** — i.e. **~10 pt loss** attributable to page-chrome on this single paper.
- Combined with the column-interleave bug (#549), chrome lines fragment paragraphs at unpredictable points.

## Hypothesis

There is no page-chrome detector in the `extract_text` path. The `to_plain_text_all` / `to_markdown_all` pipelines do some chrome stripping (see 2201.00029 score delta), but it's inconsistent — `to_plain_text_all` then *over*-strips and hurts other PDFs (2201.00069: 85% → 65%, 2201.00201: 85% → 60%).

## Suggested fix

Implement a simple recurring-line detector: any line that appears at the same Y-band in ≥ 30% of pages (configurable) and matches one of the header/footer signatures (`page N of M`, bare integer, journal masthead) should be tagged as chrome. Expose a flag `include_page_chrome=False` (default) on `extract_text` / `to_plain_text_all` to keep behaviour predictable across modes.

## Related

- py-pdf/benchmarks#19
- Interacts with #549 (column interleave)

### LINKED ISSUE #554: text-extraction: three plain-text APIs (extract_text / to_plain_text / markdown-strip) produce inconsistent output; best mode varies per PDF
## Summary

The three "plain text" public APIs on `PdfDocument` (`extract_text(page)`, `to_plain_text(page)` / `to_plain_text_all()`, and the markdown→strip path via `to_markdown_all()`) produce **substantially different output** on the same input, with no documentation of when to use which. None is uniformly best — best mode depends on the PDF.

## Measured behaviour (py-pdf/benchmarks corpus, 14 PDFs, Levenshtein vs ground truth)

| PDF | extract_text | to_plain_text_all | markdown-stripped | best mode |
|---|---:|---:|---:|---|
| 2201.00214 | **95.5** | 94.3 | 89.2 | extract_text |
| GeoTopo-book | **91.3** | 89.4 | 73.6 | extract_text |
| 2201.00151 | 52.8 | 62.7 | **62.9** | md/plain ≈ tie |
| 1707.09725 | 92.3 | **94.8** | 83.9 | to_plain_text_all |
| 2201.00021 | 66.8 | 63.1 | **90.0** | markdown |
| 2201.00037 | 92.6 | **93.6** | 86.8 | to_plain_text_all |
| 2201.00069 | 84.5 | 64.5 | **92.0** | markdown |
| 2201.00178 | 87.7 | **90.6** | 85.6 | to_plain_text_all |
| 2201.00201 | 85.4 | 59.5 | **95.9** | markdown |
| 1602.06541 | 76.7 | 59.3 | **92.4** | markdown |
| 2201.00200 | 75.3 | 69.5 | **94.8** | markdown |
| 2201.00022 | 63.1 | 58.6 | **73.1** | markdown |
| 2201.00029 | 85.4 | **96.1** | 89.8 | to_plain_text_all |
| 1601.03642 | 74.0 | 67.2 | **93.0** | markdown |
| **mean** | **80.2** | **75.9** | **85.9** | best-of-each: **89.7** |

## Why this matters

- A consumer who picks `extract_text` (the most discoverable name) loses ~10 pt of accuracy on average vs `to_markdown_all`-and-strip.
- A consumer who picks `to_plain_text_all` loses ~20 pt on column-heavy PDFs.
- A simple plain-text caller has no way to know which mode their PDF wants; the best-of-each strategy is **+9.5 pt** over `extract_text`.

## Hypothesis

Three different pipelines exist internally (legacy glyph-walk for `extract_text`, an intermediate path for `to_plain_text`, the XY-cut markdown converter for `to_markdown_all`) and the inconsistency reflects partial rollout of the #534 fix.

## Suggested fix

Either:
1. Unify all plain-text outputs onto the XY-cut markdown pipeline followed by markdown stripping, OR
2. Document explicitly which API is "high fidelity / structure-aware" and have `extract_text` delegate to it by default, with a `legacy=True` flag for backward compatibility.

Optional additional surface: expose a `to_plain_text_high_fidelity()` that does markdown→strip internally, so downstream code doesn't have to know about the markdown layer.

## Reproduction

```python
import pdf_oxide, re
from Levenshtein import ratio
doc = pdf_oxide.PdfDocument("2201.00201.pdf")
gt = open("2201.00201_gt.txt").read()
print("extract_text:", ratio(gt, '\n'.join(doc.extract_text(i) for i in range(doc.page_count()))))
print("to_plain_text_all:", ratio(gt, doc.to_plain_text_all()))
md = re.sub(r'[#*_`>]+', '', doc.to_markdown_all())
print("md-stripped:", ratio(gt, md))
```

## Related

- py-pdf/benchmarks#19
- #549 (column interleave in extract_text)
- #534, #543 (XY-cut)

### LINKED ISSUE #557: text-extraction: Arabic glyphs in mid-paragraph emitted in visual (reverse) order; BiDi shaping inconsistent within a single page
## Summary

After the RTL fix in #459 / #121 / #330 closed, the per-page extraction still emits **some** Arabic runs in visual-order (reversed-character) instead of logical order. The same page mixes correct and reversed runs, so it is not a pure no-RTL regression — it looks like the BiDi pass only fires for the first/last run on a line.

Affected fixture (single page, ~1.4 KB extracted): `right_to_left_01.pdf`.

Source: upstream Goldziher/kreuzberg corpus; mirrored locally at `tools/benchmark-harness/fixtures/kreuzberg/pdfs/`.

## Reproduction

```python
import pdf_oxide  # 0.3.54
doc = pdf_oxide.PdfDocument("tools/benchmark-harness/fixtures/kreuzberg/pdfs/right_to_left_01.pdf")
print(doc.extract_text(0))
```

## Observed

Most paragraphs render correctly (logical-order Arabic). One run on the page is **reversed glyph-by-glyph** — first letters of words appear last:

```
... عمليات تحليلية متقدمة، مثل النمذجة لتنفيذ
ةقد رثكأ تارارق ذاختا ىلإ اضً يأ يدؤي نأ نكمي لب ،تقولا رفوي طقف سيل اذه .ةريبكلا تانايبلا ليلحتو ةيئاصحإلاتانايبلا ىلع ةمئاق تاجاتنتسا ىلع ءانبً.
```

`ةقد رثكأ تارارق` is `قدرة أكثر قرارات` in reverse. Same paragraph then continues with correctly-ordered Arabic. `to_plain_text_all` and `to_markdown_all` show the same defect.

A secondary defect: at every Latin/Arabic boundary the markdown converter splits the inline Latin token onto its own block (`R`, `Python`, `pandas` each become standalone paragraphs / `#` headings), which destroys mixed-direction headings like `تحسين الإنتاجية وحل المشكلات من خلال البرمجة بلغة R و Python`.

## Hypothesis

The reversed run appears to be a CID/TrueType font where the show-text operator emits glyphs in the order the source operator already pre-shaped them (i.e. already-visual). When BiDi mirroring runs again on output, it reverses an already-visual run a second time. Triggered by font-resource switch mid-paragraph.

## Impact

`right_to_left_01.pdf` is the only RTL fixture in the kreuzberg PDF corpus, but it is the only canary for the §14.8.2.3.3 / #459 work. The defect would also bite any Hebrew/Persian/Urdu document where the source PDF mixes pre-shaped and unshaped runs.

## Related

- Reopens scope of #459 (RTL writing-system support), #330 (Arabic visual-order), #537 (Hebrew LTR/RTL)
- Cross-ref #549 for shared reading-order pipeline

### LINKED ISSUE #567: benchmark: olmOCR-bench (999 prose PDFs) as a CI regression corpus; baseline multi_column 0% / arxiv_math 0% / headers_footers 100%
## Summary

Adding `allenai/olmOCR-bench` (allenai/olmOCR-bench v1.0, ODC-BY) as a second public regression corpus for pdf_oxide. The bench ships 999 single-page PDFs across 3 prose-oriented subsets and uses checkable assertions (substring-order, substring-absent, latex-present) instead of full-text ground truth, which makes it cheap to run and diff in CI.

This issue tracks baseline numbers; subsequent issues will land per-failure root causes (one issue already filed for intra-word kerning splits; #549 comment cross-links the multi-column XY-cut evidence).

## Suites & 50-PDF sample baseline (pdf_oxide v0.3.54)

| Subset | n PDFs | n assertions | assertion pass | PDF all-pass | dominant failure |
|---|---:|---:|---:|---:|---|
| `multi_column` | 47 | 185 | 0 / 185 (0.0%) | 0 / 47 | #549 column interleave + new kerning bug |
| `arxiv_math` | 50 | 246 | 0 / 246 (0.0%) | 0 / 50 | math LaTeX form not preserved (expected, lossy plain-text extraction) |
| `headers_footers` | 50 | 147 | 147 / 147 (100%) | 50 / 50 | n/a — corpus PDFs are pre-stripped of chrome by authors, not a fair test of #553 |

100% on headers_footers is misleading positive; the corpus uses `_processed.pdf` filenames where the upstream had already removed page chrome before publishing the assertions. Useful baseline though — pdf_oxide does not RE-introduce chrome.

`multi_column` 0% is real. Worth a dedicated CI gate: even moving to 50% on this subset would represent dozens of customer-visible improvements.

## Where to fetch

```python
from huggingface_hub import snapshot_download
snapshot_download(
    repo_id="allenai/olmOCR-bench", repo_type="dataset",
    local_dir="./olmocr-bench",
    allow_patterns=[
        "bench_data/*.jsonl",
        "bench_data/pdfs/arxiv_math/*",
        "bench_data/pdfs/multi_column/*",
        "bench_data/pdfs/headers_footers/*",
    ],
)
```

Total download for those 3 subsets ≈ 212 MB (999 PDFs).

Other untested subsets (worth piloting once XY-cut fix lands):
- `long_tiny_text` (442 assertions) — small-type body text
- `old_scans` (526) and `old_scans_math` (458) — scanned, may stress OCR (not pdf_oxide's primary path)

## Suggested CI gate

Add a quick-bench job that:
1. Pulls only `multi_column.jsonl` + 211 `multi_column/*.pdf` (~57 MB).
2. Runs `to_markdown_all` (or fixed `extract_text`) and scores `order` assertions.
3. Fails the build if assertion-pass rate regresses by more than 5 pt vs main.

Estimated runtime: < 30 s on a CI runner once warmed.

## CJK/RTL validation note for v0.3.47 and v0.3.54

Inconclusive on olmOCR-bench — the corpus is overwhelmingly English. Only ~5 PDFs out of 600 sampled triggered Japanese / Chinese / Persian font warnings. Persian extraction is broken (see separate issue on NazaninNormal). For CJK validation, would need OmniDocBench (image-only, blocked) or a fresh CJK PDF corpus.

## Bench corpus

<https://huggingface.co/datasets/allenai/olmOCR-bench> (license: ODC-BY)
Paper: <https://arxiv.org/abs/2502.18443>

Related: #549, #553, #555, py-pdf/benchmarks#19.

### LINKED ISSUE #572: performance: PoC||GTFO polyglot PDFs (corrupt xref) take ~74s for first-page text on 33MB file; open and first extract both pay heavy cost
## Summary

Three PoC||GTFO polyglot zine PDFs (intentionally crafted with corrupted/non-standard cross-reference tables so the file is *also* a valid ZIP/HTML/etc.) take **10-50× longer** to open and extract from than normal PDFs of the same size, because `pdf_oxide` re-scans the entire file to rebuild xref after the initial parse fails, **and then re-does the heavy work again on the first page-level call** (open does not fully populate caches).

| File | Size | Pages | `PdfDocument(path)` | `extract_text(0)` |
|---|---|---|---|---|
| pocorgtfo01.pdf | 3.8 MB | 17 | **0.27s** | **3.52s** |
| pocorgtfo02.pdf | 14 MB | 32 | **4.08s** | **12.06s** |
| pocorgtfo07.pdf | 33 MB | 64 | **48.34s** | **25.44s** ⚠ |
| (baseline) issue3188.pdf | 8.1 MB | 1 | 2.38s | 0.81s |
| (baseline) 22060_A1_01_Plans.pdf | 6.3 MB | 1 | 0.70s | 0.81s |

pocorgtfo07 = **~74s total for first-page text** vs ~3s for a non-pathological 8 MB PDF.

Stderr during open: `Regular xref parsing failed: Invalid cross-reference table, attempting reconstruction` (from #558's category of warnings).

## Reproduction

```python
import pdf_oxide, time
# Public archive issues:
# https://www.alchemistowl.org/pocorgtfo/pocorgtfo07.pdf  (~33 MB)
t0 = time.time(); d = pdf_oxide.PdfDocument("pocorgtfo07.pdf")
print("open", time.time() - t0)            # ~48s on a fast workstation
t0 = time.time(); print(len(d.extract_text(0)), time.time() - t0)  # ~25s
t0 = time.time(); print(len(d.extract_text(1)), time.time() - t0)  # ~0.2s  (warm)
```

Pages 1+ extract in milliseconds — so the cost is concentrated in: (a) initial xref reconstruction during `PdfDocument()` and (b) some lazy re-parse triggered by the *first* `extract_text` call only.

## Expected

- Xref-reconstruction time should be proportional to file size, not to file size × constant factor (~1.5 s/MB observed for these inputs).
- `PdfDocument()` opening should fully populate whatever cache `extract_text(0)` triggers, so the perf cost lands once, not twice.
- Worst-case opening of a 33 MB pathological PDF should be in the order of seconds, not approaching a minute.

## Actual

See table. pocorgtfo07 spends ~70 s total before the caller sees any text from page 0.

## Hypothesis

1. `attempting reconstruction` does a full byte-level scan to rebuild xref. For a 33 MB file with many embedded objects this is O(n) but with a high constant — likely doing per-object regex / repeated parses.
2. The xref-reconstruction artifact is not fully cached: `extract_text(0)` re-walks something heavy. Subsequent pages don't pay the cost (0.05-0.22 s each in the same run), so a one-shot cache miss / lazy initialization is plausible.
3. Stderr spam during open (#558) adds I/O overhead — every recovered object emits a warning.

## Impact

- Any user pointed at one of these intentionally-malformed-but-renderable files (or any PDF with a corrupt-but-recoverable xref — there are many in the wild from buggy scanners and CMS systems) will experience apparent hangs of tens of seconds, with no progress signal.
- Related to #559 (content-stream cap) in that both are pathological-input perf issues; this one isn't capped — it just takes ages.

## Related

- #558 (stderr spam during recovery — same code path)
- #559 (content-stream cap on pathological PDFs)
- py-pdf/benchmarks#19

### LINKED ISSUE #575: perf: extract_text hangs (>30s/page) on archive.org / Google Books scanned PDFs with hOCR text layer
## Summary

`PdfDocument.extract_text(page_idx)` hangs (no progress, no return after >60s) on certain pages of large archive.org-style scanned PDFs that have an invisible OCR text layer in `v0.3.54`. The first page in particular takes 30s+ even on a "small" 24 MB book, and on larger 73 MB / 85 MB books no single page returns within 60s.

`pdftotext` (poppler) processes the same pages in <1 s.

## Repro

Reproduce with any of these archive.org public-domain scans:

| PDF | Size | Pages | Producer | extract_text(p0) | extract_text(p100) |
|---|---|---|---|---|---|
| alice_old.pdf (`alicesadventures00carrrich`) | 24 MB | 448 | LuraDocument PDF v2.68 | **32 s** | 0.2 s |
| iliad.pdf (`homerandiliadtr02homegoog`) | 9.6 MB | 453 | Google Books PDF Converter | **timeout >60s** | **timeout >60s** |
| lyell_geology.pdf (`principlesofgeol00lyel`) | 73 MB | 870 | LuraDocument PDF v2.68 | **timeout >60s** | n/a |
| war_peace.pdf (`warandpeace030164mbp`) | 85 MB | 725 | IA PDF 1.4.25 | **timeout >60s** | n/a |

```python
import time
import pdf_oxide

doc = pdf_oxide.PdfDocument("alice_old.pdf")
t = time.time(); txt = doc.extract_text(0)
print(time.time()-t, len(txt))   # 32 s, 0 chars on this page
t = time.time(); txt = doc.extract_text(100)
print(time.time()-t, len(txt))   # 0.2 s, 983 chars
```

## Symptoms

* `page_count()` returns instantly and correctly.
* `extract_text(small_idx)` on later pages of `alice_old.pdf` is fast.
* `extract_text(0)` blocks for ~30 s and returns empty string.
* On iliad/lyell/war_peace, every sampled page (0, 50, 100, 200) blocks past 60 s — effectively unusable.
* `classify_page` also blocks.
* `to_markdown_all()` triggers what looks like a per-page page-tree re-walk: 905 lines of `Malformed PDF: Pages node missing /Kids array` / `Page tree traversal failed ... trying fallback scan method` written to stderr per attempt (related to #558).

## Producers commonly affected

* "LuraDocument PDF v2.68" (Internet Archive recoded scans)
* "Google Books PDF Converter (rel 1 21/8/06)"
* "Internet Archive PDF 1.4.25; including mupdf and pymupdf/skimage"

These are the most common PDF producers across the millions of public-domain books on archive.org and Google Books. A real-world ingestion pipeline pointed at any IA / Google Books mirror will hang within the first few documents.

## Likely cause

These PDFs typically have an MRC structure with a deeply-nested or non-standard /Pages tree, and many have a `/Kids` array missing on intermediate nodes (hence the noisy fallback warning). The fallback "linear scan" path appears to be quadratic or worse and is invoked once per `extract_text` call, not cached.

## Suggested fixes

1. Cache the page-tree resolution after the first successful (or fallback) traversal so subsequent `extract_text` calls don't re-scan.
2. Add a parser timeout / op-cap (similar to #559) on the fallback scan so a single page can't pin a worker forever.
3. Profile the per-page extraction on these IA / Google Books PDFs; the order-of-magnitude gap vs poppler suggests an O(N) walk being done where a cache should suffice.

## Severity

Real-world blocker for archive.org / Google Books ingestion. Most large-corpus public-domain pipelines tap these as primary sources.

### LINKED ISSUE #597: Type 3 fonts must not be cached in the cross-document global font cache
### Bug Description

Follow-up to #595 / PR #596.

PR #596 fixes cross-document `global_font_cache` poisoning for **subset** fonts (BaseFont names beginning with a 6-uppercase-letter tag + `+`, per PDF 32000-1 §9.6.4). That covers the reported `° → 日` case, but it does **not** cover **Type 3 fonts**, which are inherently document-local yet carry no subset prefix and therefore still slip past the new `!is_subset` guard at `src/document.rs:10957` and `:10976`.

Per PDF 32000-1 §9.6.5 (`docs/spec/pdf.md:17860`):

> Type 3 fonts differ from the other fonts supported by PDF. A Type 3 font dictionary defines the font; font dictionaries for other fonts simply contain information *about* the font and refer to a separate font program for the actual glyph descriptions. In Type 3 fonts, glyphs shall be defined by streams of PDF graphics operators.

The glyph procs live in the font's `/CharProcs` dictionary and reference the **owning document's resources** (XObjects, ColorSpaces, ExtGState, etc.). Sharing a parsed `FontInfo` for a Type 3 font across `PdfDocument` instances will produce wrong glyph rendering / wrong metrics on the second document if the two PDFs happen to share `/Name` + `/Subtype /Type3` + `/Encoding` shape.

### Reproduction sketch

Same general shape as the #595 reproducer, but with two PDFs that:

1. Both define `Subtype /Type3` fonts with the same `/Name` and `/Encoding`.
2. Have different `/CharProcs` content streams.

After loading PDF A's Type 3 font, loading PDF B should return PDF A's cached `FontInfo` from `global_font_cache_get`, yielding incorrect glyphs.

### Proposed fix

Extend `font_identity_hash_cheap` (or wrap it) to also return `is_uncacheable_across_documents = true` when `Subtype == /Type3`, and gate Layer 6 on `!is_subset && !is_type3` (rename the predicate to something like `is_document_local`). The per-document Layer 4/5 caches remain safe and should still apply.

### References

- Spec: PDF 32000-1 §9.6.5 — `docs/spec/pdf.md:17860`
- Related: #595, PR #596
- Surfaced by: @haberman's comment on #595 (case #2: "Two Type 3 fonts from different documents are considered the same, even though Type 3 fonts cannot safely be cached across documents").

### Checklist

### LINKED ISSUE #598: Cross-document font cache returns wrong /Widths when two PDFs share a BaseFont name with different metrics
### Bug Description

Follow-up to #595 / PR #596.

PR #596 fixes cross-document `global_font_cache` poisoning for **subset** fonts. The same poisoning class still applies to **non-subset** fonts when two PDFs reference the same `BaseFont` name but ship different `/Widths` arrays.

`font_identity_hash_cheap` in `src/document.rs:10698` keys on `BaseFont` + `Subtype` + `Encoding` + presence of `ToUnicode`/`FontDescriptor`/`DescendantFonts`. It does **not** incorporate `/Widths`, `/FirstChar`, `/LastChar`, or the descendant CIDFont `/W` array. Two PDFs that both reference e.g. `Helvetica` (or a custom-named non-subset font) with materially different widths will collide on the global cache key after this PR.

### Why it matters

- Standard 14 fonts (Helvetica, Times-Roman, etc.) explicitly allow producers to ship custom `/Widths` that override the built-in metrics (PDF 32000-1 §9.6.2.2). Two PDFs that do this differently will get each other's metrics from the cache.
- Custom non-subset embedded fonts that share a `BaseFont` name across documents but were subset/optimized differently by different producers will similarly collide.

### Reproduction sketch

Same shape as the #595 reproducer:

1. PDF A: `/BaseFont /Helvetica` with `/Widths [600 600 600 ...]` (monospace override).
2. PDF B: `/BaseFont /Helvetica` with the real Helvetica width table.

After loading PDF A, loading PDF B should return PDF A's cached `FontInfo`, yielding wrong text-positioning / wrong glyph advances on PDF B.

### Proposed fix (two options)

**Option A — cheap key extension:** include a digest of `/Widths` + `/FirstChar` + `/LastChar` (and the descendant CIDFont `/W` array for Type 0) in `font_identity_hash_cheap`. Keeps Layer 6 useful for genuinely identical fonts.

**Option B — content-hash ToUnicode and widths together:** load and digest the actual byte content of ToUnicode + Widths arrays. More expensive on cache-miss but eliminates this class of false hit at the source. Could be done lazily — cache key stays cheap, validate on hit.

### References

- Spec: PDF 32000-1 §9.6.2.2 (standard font width overrides), §9.7 (CIDFont `/W` array)
- Related: #595, PR #596
- Surfaced by: @haberman's comment on #595 (case #1: "Two fonts with different /Widths map to the same cache key (leads to incorrect metrics)").

### Checklist

### LINKED ISSUE #607: Multi-column routing: gutter-corridor signal to read fragment-granular / short-line 2-column bibliographies column-by-column
## Summary

Track the concrete geometric fix for the multi-column interleave class (umbrella: #536, benchmark: #567). A full 2,124-PDF cross-corpus sweep (v0.3.57) showed ~64 olmOCR academic/reference pages clustering at **~0.50** similarity vs pdftotext + PyMuPDF (which agree ~1.00) because two-column pages are read **across-then-down** (columns interleaved row-by-row).

This is **pre-existing** (not a regression) and the prior multi-column fixes (#534 prose, #549 XY-cut bypass) are already shipped/closed — but the **fragment-granular, shared-baseline reference/bibliography class still fails**.

## Root cause (verified on geometry)

Blocked at the document-level routing gate `Document::is_multi_column_page` (`src/document.rs:9056`), which decides XY-cut vs the legacy row-aware `(y, x)` sort. On these pages both columns share Y baselines (e.g. 12_pg174_pg1.pdf: at `y=716` the left entry starts `x≈73`, the right `x≈318`, same rounded line), so the row-aware sort glues them.

Gate-by-gate (12_pg174, 286 spans):
- `len < 12` → pass; `has_bimodal_line_starts` → **fail** (every row physically begins at the left-column start `x≈73..82`, range < `MIN_GUTTER_PT=30`; the right column shares `round(y)` and never forms a second line-start peak);
- span-center histogram → 2 peaks pass (12_pg174/12_pg275); **`2503.09174` gets 0 peaks** (dense word-spans fill X continuously) — distinct sub-case;
- Y-overlap > 0.5 → pass; per-half count ≥ 15 → pass (142/144);
- **`dominant_cluster_fraction ≥ MIN_DOMINANT_FRACTION (0.5)` at `src/document.rs:9224` → FAIL — the blocker.** With `cluster_gap = stats.dominant_em.max(4.0)` (`:9193`, ≈7.5pt here) the largest contiguous left-edge cluster captures only **0.31 / 0.32** of each half (12_pg174) and **0.50 / 0.376** (12_pg275). Word-fragment x-starts scatter across dozens of positions; 7.5pt can't bridge inter-word spacing.

Secondary (even if the doc-gate passed): `detect_narrow_gutter_prose` (`src/pipeline/reading_order/xycut.rs:929`) concentration floor `best_size*10 < gap_positions.len()*7` (≥0.70, `:1029`) grazes at **0.65 / 0.69**. And for the reporter's short-verse Bible case, the prose classifier gate `mean_chars > 20` (`xycut.rs:707`, enforced `:887/:1049`) is the relevant blocker — short citation lines never score `Prose`.

## The right signal (how poppler / MuPDF do it)

Reference engines decide columns from the **vertical persistence of an empty x-corridor** (gutter whitespace projection over the page band) and the **count** of such corridors — **never from chars-per-line**. A 2-column bibliography = ONE deep, consistently-positioned empty gutter; a data table = MANY inter-cell corridors at varying x interrupted by rows. pdf_oxide already computes this corridor signal in `detect_narrow_gutter_prose` / `detect_two_column_prose` but discards it behind the `mean_chars > 20` prose gate.

## Plan (each step full-corpus gated)

1. **(low risk, preferred)** Add an *additive* within-line gutter-corridor accept path to `is_multi_column_page`: when the existing pre-checks pass but `dominant_cluster_fraction` fails, accept iff a large majority of body lines share **one** consistent vertical gutter corridor (low variance of gap-center x) **and** left-edges form exactly **two** clusters. Leaves `MIN_DOMINANT_FRACTION=0.5` untouched (no currently-routed page changes); XY-cut's own `classify_region_kind==Prose` gate re-rejects any table that slips through.
2. **(low–med)** Make the tight-gutter prose decision gate on *single-persistent-corridor + consistent-gutter-x* rather than chars-per-line; demote `mean_chars` to a tie-breaker. Optionally relax `detect_narrow_gutter_prose` floor 0.70 → ~0.62 so pages reaching XY-cut split.
3. Handle the arxiv `peaks = 0` sub-case (`2503.09174_pg2.pdf`) separately (lower priority).
4. **(avoid)** Widening `cluster_gap` at `:9193` — modifies an existing safety gate; risks bridging table cell-columns.

## Acceptance bar (hard external gate)

Full `run_corpus_bench.py` 2,124-PDF sweep: olmOCR `multi_column` uplift vs pdftotext + PyMuPDF; **`google_doc_document.pdf` table byte-identical** (the two prior reverts corrupted it — `"273.879.7501" → "1273.879.750"`); 0 panics / 0 timeouts; avg Lev not down; instrument & hand-audit every page whose `is_multi_column_page` verdict flips, reject if any real table flips to multi-column. Unit tests are explicitly insufficient (per the #534 post-mortem at `xycut.rs:73-101`).

Why not in v0.3.57: touches the central per-page reading-order routing gate; non-blocking (pre-existing); requires the full sweep above.

yfedoseev:
## v0.3.57 — Community contributions + extraction-quality sweep

Separation plates, OCG ink filtering, two-phase images, rendered-advance metrics, and OCG-layer-on-paths from four external contributors, **plus** a large text-extraction quality sweep: multi-column reading order, page rotation, CJK/UTF-8 CMap decoding, RTL logical order, indirect-ref page boxes, font-cache correctness, form/signature recovery, and chrome removal.

### ✨ Community contributions (integrated with authorship preserved)
- **#602** `TextChar::rendered_advance` — per-glyph cursor advance (Tc/Tw). Thanks @haberman
- **#605** Separation plate rendering — `render_separations()` per-ink grayscale (§8.6). Thanks @RayVR
- **#600** OCG ink filtering for text extraction — `extract_text_filtered()` + 3 OCG/cache bug fixes. Thanks @RayVR
- **#588** `page_image_handles()` two-phase image API (Form XObject recursion). Thanks @kh3rld
- **#606** Type 3 font width scaling via `FontMatrix[0]` (not hard-coded 1/1000). Thanks @haberman
- **#587** Optional Content Group (PDF "layer") name on extracted paths — OCMD resolution, PDFDocEncoding/UTF-16 decode, XObject-scoped `/Properties`. Thanks @willywg

### 🛠 Extraction fixes
- **RTL Hebrew/Arabic** now extracts in logical order across plain-text, Markdown/HTML, and tagged paths (#557)
- **Two-column references/bibliographies** read column-by-column instead of interleaved (#607, #549, #536)
- **CJK in UTF-8 CMap fonts** (`Uni-Utf8-H`, Adobe `Uni*-UTF8-H`) now decoded — previously 0 chars (#610)
- **Non-embedded Japanese (JIS) fonts** decode correctly instead of garbled Latin
- **Indirect-reference `/MediaBox`/`/CropBox` elements** resolved — page no longer collapses to empty (§7.3.10)
- **`/Rotate 180` pages** read in display order instead of fully reversed
- **Signature/`/AP` appearance-stream field text** recovered; **unchecked-checkbox `[ ]` noise** removed
- **Page-number / running-header chrome** no longer leaks into body text (#553)
- **Cross-document font-cache** glyph corruption fixed (Type 3 scoping + `/Widths` in cache key) (#597, #598)

### ⚡ Performance
- `extract_text` no longer hangs (O(n²)→windowed) on heavily OCR-layered scans (#575)
- Corrupt-xref PDFs no longer trigger a double full-file rescan (#572)
- Regression guards for word-spacing, char-clustering scaling, `to_html` tables, multi-column detection

### 📚 Other
- Documented the three plain-text APIs and when to use each (#554)
- Added the olmOCR-bench CI regression harness (#567); configurable non-text drop heuristics

### ✅ Validation
- Full Rust suite **7898 passed / 0 failed** (incl. all new RTL/CJK/Type3/OCG tests); clippy + fmt clean; doctests 135/0 (`--features python`)
- **Main-vs-HEAD regression sweep: 156 diverse PDFs × {text, markdown, HTML} × 2 versions → ZERO regressions.** Every diff reviewed by hand; all are improvements or neutral. The table-corruption guard (`google_doc_document.pdf`) holds — cell data byte-identical, header flag placement improved.
- Cross-corpus sweep (2,124 PDFs) vs pdftotext + PyMuPDF: net positive, no table/status regressions
- Four minor polish items captured for a future release (not regressions) in `pdf_oxide_fix_issues.md`

---

Closes #557
Closes #553
Closes #554
Closes #567
Closes #572
Closes #575
Closes #597
Closes #598
Closes #607
Closes #602
Closes #605
Closes #600
Closes #588
Closes #606
Closes #587

Remaining open for follow-up (not closed by this PR): #536 (structured-extraction API), #608 (Tagged-PDF structure-tree reading order).


yfedoseev:
Thanks @copilot — triaged all six. Two real fixes landed; the rest are new-API enhancements (no regression) tracked for a follow-up.

**Fixed (commit 17d3e3b4):**
- **`font_matrix_a` inf/NaN** — a degenerate `/FontMatrix[0]` (0/near-zero/non-finite) made `default_width * 0.001 / font_matrix_a` produce inf/NaN. Now rejected → falls back to the standard 0.001 (ISO 32000-1 §9.2.4/§9.6.5), with a new regression test `type3_degenerate_zero_fontmatrix_falls_back_safely`.
- **Type3 tests early-returning** — both now assert the glyph extracts (`.expect(...)`) instead of silently returning, so a width-scaling regression can't pass vacuously. Verified: the glyph does extract.
- **Indexed color-space comment** — corrected the misleading comment that claimed indirect-ref resolution the code doesn't perform.

**Tracked for follow-up (new `page_image_handles()` API; no regression to existing behavior — documented in `pdf_oxide_fix_issues.md` §8):**
- **`decode()` resource-name color spaces** (`color_space_map=None`) — real gap for `/CS0`-style spaces; the rendering path still handles them. (§8a)
- **Indexed handle reports base color space** — a semantics decision (report `Indexed` vs base + accessor); decode re-parses independently so pixels are unaffected. (§8b)
- **`decode()`/`raw_compressed_bytes()` consume `self`** — a consuming API is intentional; `&self`/`Clone` ergonomics is an optional improvement, not a bug. (§8c)

Issue #611, opened 2026-05-30
[Feature]: REST API Support and Official Docker Image


mitslabo:
### Problem Statement

Integrating this PDF editing library into workflow automation systems is currently difficult.

For example, when using automation platforms such as n8n, PDF processing typically requires custom scripts, language-specific wrappers, or tightly coupled runtime environments. This increases implementation complexity and reduces portability.

In many automation scenarios, PDF editing is only one step within a larger workflow. A service-oriented approach would make integration significantly easier and more maintainable.

### Proposed Solution

Provide an HTTP (REST) API interface for the core functionality of the library, together with an official Docker image exposing the API.

### REST API

A lightweight REST API could expose operations such as:

* Upload or load PDF files
* Fill template fields
* Modify PDF content
* Merge or split PDFs
* Return processed PDF files

This would allow external systems and workflow tools to invoke PDF processing without requiring direct language bindings.

### Official Docker Image

An official Docker image would simplify deployment and integration.

### Custom Wrapper Services

Users can currently build their own wrapper service around the library using frameworks such as FastAPI or Express.

However, this introduces additional implementation and maintenance overhead, and different users may implement inconsistent APIs.

### Direct Library Integration

Another alternative is integrating the library directly into application code.

While this works well for software development use cases, it is less suitable for workflow automation platforms, low-code/no-code tools, or distributed environments where HTTP-based integration is preferred.

An official REST API and Docker image would provide a more standardized and reusable integration approach.

### Feature Area

Other

### Priority

Would be helpful

### Use Case

* Integration with workflow automation tools such as n8n
* Running as a microservice in Docker Compose or Kubernetes
* Self-hosted document processing pipelines
* Internal automation systems

A simple `docker run` or Docker Compose deployment would substantially lower the adoption barrier.

### Example Usage (Optional)

```rust

```

### Additional Context

If this direction sounds useful, I would be happy to help with testing and feedback.

Thank you for considering this request.

### Checklist

- [x] I have searched existing issues to make sure this is not a duplicate
- [x] I have provided a clear description of the feature
- [ ] I am willing to contribute to the implementation (optional)

yfedoseev:
Hi @mitslabo, thanks a lot for the thoughtful write-up — it's clear you've hit a real friction point and have genuinely thought it through. I really appreciate that. 🙏

[...]

PR #613, opened 2026-05-30
Release v0.3.58 — structure-tree reading order, structured extraction, /Rotate 90/270, emoji spacing, image-handle refinement, deps

### LINKED ISSUE #536: [Bug]: Two-Column Reference PDF Extraction
### Bug Description

Possible Duplicate of #534, but adding my own findings and [an example pdf](https://pdfhost.io/v/kMkM6prLuD_194760d7-387c-45f6-bdfe-0d57ec0d403f-1).

# Bug Report / Feature Request: Two-Column Reference PDF Extraction

**Filed against:** pdf-oxide  
**Type:** Feature Request (structural extraction API)  
**Severity:** Major — makes correct extraction of two-column reference works unreliable without extensive custom post-processing

---

## Summary

Two-column reference PDFs (Bibles, dictionaries, encyclopedias, academic editions with marginal section numbers) present systematic extraction challenges that the current pdf-oxide API does not address. Extracting these documents requires reconstructing logical reading order, distinguishing structural markers from body text, and maintaining cross-page section continuity — all of which must currently be implemented by consumers with fragile heuristics.

This report is generalized from real-world extraction of multiple Bible editions but applies to **any two-column reference PDF with marginal section numbering**.

---

### 1. No Column-Aware Reading Order

**Current behavior:** Text extraction returns words/blocks in an undefined or horizontal order, interleaving left-column and right-column content.

**Impact:** "And it came to pass" (left column) merges with "the LORD said unto" (right column) producing garbled output.

**What consumers must do today:** Detect column boundary via largest X-gap heuristic, then partition every word/block into LEFT or RIGHT, then sort each column top-to-bottom independently, then interleave: LEFT column first, then RIGHT column, then next page LEFT, then next page RIGHT.

**Request:** Expose a column-aware extraction mode that returns content per-column in reading order (LEFT top-to-bottom, then RIGHT top-to-bottom, per page).

### 2. No Distinction Between Marginal Labels and Body Text

**Current behavior:** All text is treated identically — a "7" that is a verse/section number at the left margin is returned the same as a "7" that appears in "7 camels" within a paragraph.

**Impact:** Section numbers are indistinguishable from inline numerals, causing structural markers to be lost or misinterpreted as content.

**What consumers must do today:** Apply X-position heuristics (e.g., `x_center < 80` for left margin, `x_center >= 180` for right margin) to guess whether a number is a marginal section label. This is fragile — column widths vary across pages and editions.

**Request:** Expose a "marginal label" classification for words/blocks that are visually separated from the body text column. This could be based on:
- X-position relative to the column body (significant left-indent suggests a label)
- Font size or style differences (marginal numbers are often smaller or bold)
- Structural analysis (a standalone short numeral at a column edge is almost certainly a label)

### 3. Section Headers Embedded in Flow Cannot Be Isolated

**Current behavior:** Inline section headers like "GENÈSE 6" or "Chapter 3" appear within the column text flow and are returned as regular words.

**Impact:** The "6" or "3" in a section header is frequently misidentified as a marginal verse/section number, corrupting the section count.

**What consumers must do today:** Pattern-match book/chapter names followed by digits, then suppress the digit word and track its Y-position. Any other word at the same Y gets suppressed too, which can accidentally remove legitimate content on the same line.

**Request:** Provide a "structural heading" classification that identifies lines/segments functioning as section dividers rather than body text. These are visually distinct (often bold, larger, centered, or preceded by whitespace) and semantically separate from the column flow.

### 4. No Cross-Page Section Continuity

**Current behavior:** Each page is extracted independently. There is no mechanism to determine that "verse 15" on page N belongs to the same chapter as "verse 1" on page N-1.

**Impact:** Chapter/section assignment must be reconstructed from scratch on every page. When a column has no explicit section header (e.g., a chapter continues from the previous page), consumers must carry state manually.

**What consumers must do today:** Maintain a `current_chapter` state variable per column, propagated across pages. When no section header appears on a page, fall back to the last known chapter. This fails when the page header announces a different chapter than the one actually continuing.

**Request:** Provide an optional mode that tracks section/chapter continuity across pages. At minimum, expose the semantic relationship: "this page's left column continues section N from the previous page's right column."

### 5. Content Spillover Across Boundaries

**Current behavior:** Book/section boundaries are detected by page headers only. A page with header "EXODus 1" is assumed to contain only Exodus chapter 1 content.

**Impact:** In two-column layouts, the **left column** of a boundary page frequently contains the **last section of the previous book** while the right column begins the new book. Stopping extraction at the header boundary cuts off the previous book's final content.

**What consumers must do today:** Extend the page range by one page past the detected boundary, then filter out content belonging to the next book by checking for its inline section headers. This is error-prone and produces missing content if the overlap is more than one page.

**Request:** Do not treat page headers as strict content boundaries. Instead, indicate per-column which book/section the content belongs to, based on inline markers. Allow consumers to request "all content for section N" regardless of which pages it spans.

### 6. Header and Footer Contamination

**Current behavior:** Running headers ("GENÈSE 7") and page numbers are returned as regular text blocks.

**Impact:** Page numbers (e.g., "7" at bottom-center) and running headers (e.g., "PSALMS 23" at top) are interpreted as marginal section labels, producing phantom verses/sections.

**What consumers must do today:** Filter blocks by Y-position (e.g., `y1 < 48` for headers, `y0 > 505` for footers) and by content pattern (regex matching known book names). This is brittle — different editions use different Y layouts, and page numbers appear at varied positions.

**Request:** Classify and separate running headers, page numbers, and footnotes from body content. These are visually and semantically distinct regions that should not be mixed with column text.

---

### Option A: High-Level Structured Extraction

```rust
/// Extract a two-column reference document with structural awareness.
fn extract_structured(pdf: &Document, config: &ExtractionConfig) -> Vec<Section> {
    // Returns sections in reading order, each containing:
    // - section identifier (chapter number, etc.)
    // - ordered paragraphs/verses with marginal labels separated from body text
    // - no header/footer contamination
}

struct Section {
    book: String,        // e.g., "Genesis"
    chapter: u32,        // e.g., 6
    start_page: usize,   // page where section begins
    end_page: usize,      // page where section ends (may span pages)
    content: Vec<Paragraph>,
}

struct Paragraph {
    label: Option<Label>, // e.g., verse number "7"
    body: String,         // the actual text content
    is_marginal: bool,    // true if label was at margin, not inline
}

struct ExtractionConfig {
    column_mode: ColumnMode,     // TwoColumn, SingleColumn, AutoDetect
    label_detection: LabelMode,  // MarginPosition, FontAnalysis, Both
    strip_headers: bool,         // Remove running headers
    strip_footers: bool,         // Remove page numbers
    strip_footnotes: bool,       // Remove footnotes
    reading_order: ReadingOrder,  // LeftFirst, RightFirst, AutoDetect
}
```

### Option B: Low-Level Enrichment (Smaller API Surface)

Enhance the existing `get_text` output with metadata without changing the API shape:

```rust
/// Word-level metadata now includes structural classification.
struct RichWord {
    text: String,
    x0: f64, y0: f64, x1: f64, y1: f64,
    column: Column,           // Left, Right, FullWidth
    role: WordRole,           // Body, MarginalLabel, SectionHeader, PageNumber, RunningHeader, Footnote
    section_id: Option<u32>,  // Which section this word belongs to (cross-page)
}

enum WordRole {
    Body,            // Normal paragraph text
    MarginalLabel,    // Verse/section number at column edge
    SectionHeader,    // "GENÈSE 6", "Chapter 3", etc.
    PageNumber,       // Footer page number
    RunningHeader,    // Top-of-page book/chapter indicator
    Footnote,         // Footnote text or marker
}

enum Column {
    Left,
    Right,
    FullWidth,  // Single-column content (headers, footers)
}
```

### Option C: Post-Processing Pipeline (Most Incremental)

Add a **reflow** pass to the existing text extraction that reorders output into reading order, strips headers/footers, and classifies marginal labels:

```rust
/// Reflow extracted text for two-column reference documents.
fn reflow(text: &ExtractedText, rules: &ReflowRules) -> ReflowedText {
    // 1. Detect columns from block X positions
    // 2. Sort blocks into reading order (LEFT top→bottom, then RIGHT top→bottom)
    // 3. Classify marginal labels vs body text
    // 4. Strip running headers and page numbers
    // 5. Return structured output per column
}
```

---

## Test Cases

Any implementation should correctly handle these patterns found across multiple Bible editions and two-column reference works:

### TC-1: Cross-Column Chapter Continuation
A chapter starts in the right column of page N and continues in the left column of page N+1. The left column has no chapter marker. The extractor should assign both columns to the same chapter.

### TC-2: Marginal Label Adjacent to Inline Section Header
An inline header "GENÈSE 6" at Y=101 has its chapter number "6" at Y≈101, same Y position as legitimate verse "6" in the margin. The extractor should classify the "6" in "GENÈSE 6" as a section header component, not a marginal label.

### TC-3: Book Boundary Spillover
Page N has header "EXODUS 1" but its left column contains verses 11-26 of the previous book's final chapter. The extractor should not truncate the previous book's content at the page header boundary.

### TC-4: Asymmetric Column Layouts
One column has an inline chapter marker while the other does not. Each column's chapter state must be tracked independently.

### TC-5: Page Number Misidentification
A page number "7" at Y≈530, X≈center should never be classified as a marginal section label even though "7" is a valid verse/section number.

### TC-6: Nested Numeric Markers
Text like "2Sing 7a psalm" where "2" is a marginal section number but the text body also contains numbers. The extractor should distinguish marginal labels from body numbers by position and context.

---

## Why This Matters

Two-column reference works are one of the most common PDF layouts for religious texts, legal codices, dictionaries, and academic editions. The current extraction experience requires every consumer to reimplement the same fragile heuristics:

- Column detection via X-gap analysis
- Marginal label detection via X-position thresholds
- Header/footer stripping via Y-position thresholds
- Cross-page chapter tracking via manual state management
- Section header filtering via regex pattern matching

These heuristics break when:
- Column widths vary across pages
- Font sizes differ between editions
- Margins shift between printings
- Section headers use varying formats

A structural extraction API or enriched word classification would eliminate an entire class of bugs that every consumer currently reproduces independently.

### Steps to Reproduce

1. [Open PDF](https://pdfhost.io/v/kMkM6prLuD_194760d7-387c-45f6-bdfe-0d57ec0d403f-1)
2. Attempt extraction, multiple issues.

### Expected Behavior

Should have accurately extracted all text from document.

### Actual Behavior

Content was split across boundaries and had multiple extraction errors.

_No response_

### Environment

- **OS**: Windows 11 25H2
- **Rust version**:  N/A
- **pdf_oxide version**: [v0.3.49](https://github.com/yfedoseev/pdf_oxide/releases/tag/v0.3.49)
- **Python version** Python3

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

_No response_

### Checklist

- [ ] I am using the latest version of pdf_oxide

### LINKED ISSUE #608: Prefer Tagged-PDF structure-tree reading order over geometric XY-cut when StructTreeRoot is trustworthy
## Summary

Complementary to the geometric column-routing fix (#607): when a PDF carries a **trustworthy logical structure tree**, prefer it as the authoritative reading order instead of reconstructing geometrically. This is the most robust path for tagged multi-column documents and is what MuPDF / pdf.js fall back on.

## Spec basis (ISO 32000-1, from `docs/spec/pdf.md`)

- **§14.7.1** (`pdf.md:35624`): logical structure is stored separately from visible content, "allowing the ordering and nesting of logical elements to be **entirely independent of the order and location of graphics objects** on the pages." → the structure tree *is* the authoritative reading order.
- **§14.8.2.3.1** (`pdf.md:37216`): "page content order" (content-stream sequence) and logical structure order "are logically distinct and may or may not coincide." The spec gives **no reading-order guarantee** for raw content order and **no geometric reconstruction algorithm** — so geometric XY-cut is necessarily a heuristic fallback.
- The spec explicitly admits declared order can be unreliable: a writer that can't determine correct order marks content `<< /TagSuspect /Ordering >> BDC … EMC` and sets `Suspects=true` in `MarkInfo` (`pdf.md:35640`, `:37252`).

## Proposed behavior

Reading-order precedence:
1. **Trustworthy structure tree** → depth-first traversal of the structure hierarchy, resolving each leaf MCID back to its marked-content glyph run (§14.7.4), gives column-by-column order for free.
2. Else → geometric XY-cut (#607).

"Trustworthy" predicate (all required before preferring the tree):
- catalog `/StructTreeRoot` present (`pdf.md:5597`),
- `/MarkInfo` `Marked == true` (default false, `pdf.md:35638`),
- `Suspects == false` / no `/TagSuspect /Ordering` marked content (`pdf.md:35640`, `:37252`).

Tagged 2-column bibliographies and academic papers then get correct reading order independent of the fragment-granular geometry that defeats the gutter heuristics in #607. Untagged / suspect-tagged PDFs (the majority) keep using XY-cut.

## Notes

- pdf_oxide already has a struct-tree extraction path (`extract_text_structure_order` / `_cached_with_spans`); this issue is about (a) gating on the trustworthy predicate and (b) making it the preferred order for the plain-text / markdown accessors when present, not just the tagged accessors.
- Acceptance: must not change output for untagged PDFs; for tagged multi-column fixtures, verify column-by-column order; full-corpus sweep with byte-identical table output (shared gate with #607).

### LINKED ISSUE #616: AcroForm fill: non-ASCII field values written as raw UTF-8 → mojibake (§7.9.2.2)
### Summary
When a non-ASCII value (Japanese/CJK, emoji, accented Latin) is written to an AcroForm field via `DocumentEditor::set_form_field_value`, the engine stores the **raw UTF-8 bytes** inside the PDF string object instead of a conformant PDF *text string*. A reader with no leading `U+FEFF` BOM interprets the bytes as **PDFDocEncoding** (single-byte), so `山田太郎` renders as `å±±ç"°å¤ªéƒ"` in Acrobat / Stirling-PDF / Preview / pdf.js.

### Spec basis (ISO 32000-1)
- **§7.9.2.2 "Text String Type"** — a text string is encoded as **PDFDocEncoding** *or* **UTF-16BE with a leading `U+FEFF` BOM** (`0xFE 0xFF`). Raw UTF-8 is neither → non-conformant.
- An AcroForm field value `/V` is *Variable Text* (§12.7.3.3) and is a text string, so it must follow §7.9.2.2.

### Reproduction (confirmed on `release/v0.3.58`, HEAD 16dc01db)
Build a 1-field AcroForm with the engine's own `PdfWriter`, fill `山田太郎`, save:
```
DocumentEditor::from_bytes(form)
  .set_form_field_value("full_name", FormFieldValue::Text("山田太郎"))
  .save_to_bytes()
```
Observed `/V` in the saved bytes (uncompressed):
```
/V <E5B1B1E794B0E5A4AAE9838E>      ← raw UTF-8 of 山田太郎, no BOM
```
Required:
```
/V <FEFF5C71753059 2A90CE>          ← UTF-16BE + BOM
```

### Root cause + fix location
`src/editor/form_fields.rs` — the `FormFieldValue → Object` conversion uses `Object::String(s.as_bytes().to_vec())` (raw UTF-8) for `Text` / `Choice` / `MultiChoice`. Field names (`/T`) and tooltips (`/TU`) have the same latent bug.

The correct, **already-tested** encoder exists in the crate: `Object::text_string()` / `encode_pdf_text_string()` (`src/object.rs`), which emits PDFDocEncoding-compatible bytes for ASCII/Latin-1 and `FEFF`-prefixed UTF-16BE for anything above U+00FF. The fix is to route the value/name/tooltip writes through it (ASCII stays a plain literal — no UTF-16 bloat).

### Acceptance
- Filling `山田太郎` → saved `/V` contains `FE FF 5C 71 75 30 59 2A 90 CE` and **not** the raw UTF-8 `E5 B1 B1 …`.
- Filling `"John"` stays a plain literal and round-trips.

### Context
Root cause of the mojibake reported by @mitslabo in #611 (REST/Docker API whose first operation is Japanese AcroForm fill). Blocks the `pdf_oxide_api` `japanese_form_fill_roundtrip_no_mojibake` round-trip. Affects the published crate → benefits every binding. Refs #611.

### LINKED ISSUE #617: AcroForm fill: terminal field loses /FT on save → extract_fields returns 0 (§12.7.4.1)
### Summary
After `DocumentEditor::set_form_field_value(...)` + `save_to_bytes()`, re-parsing the output and calling `FormExtractor::extract_fields` returns **0 fields** — even though the saved bytes still contain `/AcroForm` and `/V`. A *no-op* load→save (no `set_*`) preserves the field, so the regression is in how a **modified** field is rewritten, not in the reader.

### Root cause + fix location
`FormFieldWrapper::build_field_dict` (`src/editor/form_fields.rs`) rebuilds the field dictionary with `Type=/Annot`, `Subtype=/Widget`, `/P`, `/V`, and **either** a `/Parent` ref **or** a `/T` — but it never emits **`/FT`** (the field type, e.g. `/Tx`) on a *terminal* field that has no `/Parent`. `/FT` is currently only written by `build_parent_dict`.

Per **§12.7.4.1 "Field Dictionaries"**, `/FT` is an *inheritable, required* key: a terminal field takes its type from `/FT` on itself or an ancestor. A standalone single-widget field (what `PdfWriter` and most real forms produce) has no parent to inherit from, so dropping `/FT` turns it into an untyped widget annotation. `FormExtractor::extract_fields` then no longer recognises it → 0 results. (The source fixture had `/FT /Tx`; the rewrite discards it.)

### Fix
In `build_field_dict`, when the field has **no `/Parent`**, emit the field's type (and flags) on the field dict itself:
```rust
if self.parent_ref.is_none() {
    dict.insert("FT".into(), Object::Name(field_type_name(self.field_type))); // "Tx"/"Btn"/"Ch"
    if let Some(ff) = self.field_flags { dict.insert("Ff".into(), Object::Integer(ff as i64)); }
}
```
Reuse the `FormFieldType → /FT` mapping already in `build_parent_dict`, and preserve `/DA` / `/MaxLen` etc. so the rewritten field is a faithful superset of the original.

### Reproduction (confirmed on `release/v0.3.58`, HEAD 16dc01db)
```
fields in freshly-written 1-field form (before edit): 1
set_form_field_value("full_name", Text("山田太郎")) + save_to_bytes()
fields after fill+save (FormExtractor::extract_fields): 0   ← bug
```

### Acceptance
- 1-field AcroForm → fill → save → re-open → `extract_fields` returns exactly **1** field, with value `山田太郎` (combined with #616).
- No regression in existing form tests / corpus.

### Context
Blocks fill→re-read round-trips for @mitslabo's #611 (REST/Docker API, Japanese AcroForm fill first). Affects the published crate → benefits every binding. Refs #611.

yfedoseev:
## v0.3.58 (release date 2026-05-31)

Version bumped **0.3.57 → 0.3.58** across all crates, bindings, and lockfiles. All changes are **additive (no breaking API)** and covered by **TDD**, and each major item was **independently re-verified** by a separate review pass. Local gate is green: `cargo fmt --check`, `cargo clippy -p pdf_oxide --lib -- -D warnings`, `cargo doc -D warnings`, **5343 lib tests** + integration tests.

> **Release gate:** automated work stops here at the open green PR — please verify before merge/tag/publish.

### Reading order & structure
- **#608 trustworthy structure-tree reading order** — shared `struct_tree_trustworthy()` predicate + public `prefers_structure_reading_order()`; `extract_text`/`to_markdown` honour `/MarkInfo /Suspects` (suspect → geometric), trustworthy tagged (incl. PDF-1.4 catalog-`/StructTreeRoot`-only) → logical order. Non-suspect/untagged **byte-identical**. (F1–F5 fixtures.)
- **#536 two-column reference / short-verse routing** — short-line two-column bodies (bibliographies, Bible/lexicon verse) now read column-by-column via a length-independent corridor admission path + xycut `Prose` arm, gated by concentration, coverage, column char-balance, a grid-row signal, and left-edge clustering so **multi-cell tables stay off the column path**. Display-math rows excluded from gutter coverage. **Validated table-safe: `google_doc_document.pdf` markdown byte-identical** (before/after diff).

### Extraction fidelity
- **Math-font decimal/punctuation recovery (#536)** — `/Differences`-mapped `period`/`comma`/`hyphen`/`minus` in symbolic TeX fonts (CMMI etc.) recovered via AGL when the resolved Unicode is a non-sensible symbol; tightly gated, genuine `¬` and correctly-mapped fonts untouched.
- **Mixed Arabic/Latin/numeral bidi** — embedded LTR sub-runs (dates/numbers) get their UAX #9 §3.3.4 sublevel; gated to confidently-RTL lines, non-RTL output byte-identical.
- **Emoji→letter spacing** — `📄 README.md` no longer collapses to `📄README.md` (§9.10 reader latitude; arrows/math excluded).
- **`/Rotate 90/270`** — pages mapped into displayed orientation (W/H swap, §8.3.3); `rot==0` identity, `rot==180` numerically identical. *(Within-line char re-order for rotated multi-glyph spans — the issue14415 180° residual — is the one remaining tracked follow-up: a 180° page's glyphs already decode to correct Unicode, so reversing blind would corrupt text; needs the repro to land safely.)*

### New API (additive)
- **`extract_structured(page) -> StructuredPage`** (#536) — typed `StructuredRegion`/`RegionRole` + `column_index`; reuses per-span artifact/heading signals. Thanks @lggcs.
- **Image handles (#588)** — `decode()`/`raw_compressed_bytes()` borrow `&self` + `Clone` (two-phase inspect→raw→decode); `decode()` resolves resource-name `/ColorSpace` (`/CS0` via `/Resources/ColorSpace`, §8.6.6/§8.9.7) across page/Form/inline scopes; Indexed handles report `ColorSpace::Indexed` with the new `indexed_base()` accessor (§8.6.6.3). Decoded pixels unchanged.

### Dependencies (dependabot, replicated on the branch)
serde_json 1.0.150 (#578), log 0.4.30 (#592), quick-xml 0.40.1 (#579), cbc 0.2.1 (#577), purego 0.10.1 (#593), rubocop-rspec 3.9 (#581), setup-java v5.2.0 (#585), codeql 4.36.0 (#582), taiki-e install-action 2.79.12 (#580), cargo-deny 2.0.20 (#584), golangci-lint 9.2.1 (#583). The 11 dependabot PRs can be closed once this merges.

---

### Issue closure

Fully delivered in this release (TDD-covered, re-verified) — auto-close on merge:

Closes #608 — trustworthy structure-tree reading order over geometric XY-cut.
Closes #536 — two-column / short-verse routing + `extract_structured` API + math-font punctuation recovery.
Closes #616 — AcroForm fill: non-ASCII field values written as raw UTF-8 → mojibake (§7.9.2.2).
Closes #617 — AcroForm fill: terminal field loses `/FT` on save → `extract_fields` returns 0 (§12.7.4.1).

Referenced (not closed — partial or out of scope):

- Refs #611 — REST API + Docker feature is **not** in this release; the two AcroForm bugs found while scoping it (#616/#617) are fixed here, API work tracked separately for a later version.
- Refs #588 — image-handle two-phase API (merged separately); this PR refines `decode()`/`raw_compressed_bytes()` to `&self` + `Clone`.
- Refs #607 — multi-column gutter-corridor predecessor (already closed); #536's routing builds on it.

Partial / tracked follow-up (no auto-close): within-line char re-order for rotated multi-glyph spans — the issue14415 180° residual — remains open; `/Rotate 90/180/270` page mapping is delivered, the blind-reverse hazard for already-correct 180° Unicode is not, pending a safe repro.

The 11 dependabot PRs (#578, #592, #579, #577, #593, #581, #585, #582, #580, #584, #583) are replicated on this branch and can be closed once this merges (`Closes` does not auto-close PRs).


yfedoseev:
## Corpus-validated (156-PDF regression manifest, v0.3.57 → v0.3.58)

Rebuilt the wheel and ran the regression sweep (`out_head` v0.3.57 vs fresh v0.3.58) on the curated 156-PDF manifest:

- **`google_doc_document.pdf` markdown byte-identical (1242 == 1242)** — table safety holds end-to-end.
- **Only 19/156 PDFs changed — every one an improvement or neutral, zero regressions:**
  - **Tagged IRS / W-2 forms** (1065, 1120, 941, fw2, irs_f1099msc) recover correct reading order the prior version mis-ordered, e.g. `current quarter.quarter was less than` → `current quarter. If line 12 for the prior quarter was less than` — the **#608 intra-MCID span ordering** paying off broadly.
  - **`issue14415.pdf` (180°)**: `states that Company] authority that's This` → `This agreement hereby states that both [Client.Company] and [Sender.Company]` — the within-line residual is **fixed** (by MCID ordering, not unsafe char-reversal).
  - **`rotation.pdf` (90°)** reads upright; **`arxiv_2510.26287` emoji** `📄README` → `📄 README`; **PDFBOX bidi** + Bengali `IA_0-contant` settle better.
  - Two long books (`war_peace`, `lyell_geology`) that timed out in the baseline now fully extract.
  - The one dense page (Prague transit timetable) surfaces *more* real text (`Sokolovská`, `Dopravní podnik hl. m. Prahy`, `INTEGROVANÁ DOPRAVA`) vs the prior garbage number runs — also ≥ baseline.

### Two follow-up items closed by checking the actual repros
- **Emoji**: found the v0.3.58 rule was on the wrong function (glyph-merge); the `📄`→word boundary is a *span* boundary (gap == 0). Moved the rule to `document.rs::should_insert_space(prev, current)` with a `>= 0` gate. Fixed + corpus-confirmed.
- **Decimal `¬` (arxiv_2510.21165)**: diagnosed via debug logging as a genuine **font defect** — F3 is a CMSY symbol font whose `/Differences` truly encodes `58 /logicalnot` (¬) where the author draws a decimal. No `/period` name exists to recover; a "symbol-between-digits → period" heuristic would corrupt legitimate `5×3` / `2±1`, so it is correctly out of scope. CHANGELOG corrected to not overclaim.

---
**Also in v0.3.58 (post this comment):** while building the issue #611 Japanese form-fill round-trip, two engine bugs in the `DocumentEditor` fill→save path were found, reproduced, and fixed — **#616** (non-ASCII values written as raw UTF-8 → mojibake; now UTF-16BE+BOM per §7.9.2.2) and **#617** (terminal field loses `/FT` on save → `extract_fields` returns 0). Both are TDD-covered (`tests/test_form_fill_roundtrip.rs`) and ride this PR (commit bf79a8ba).

PR #614, opened 2026-05-30
feat: add ascent and descent to TextChar and FontInfo

haberman:
## Description

Adds `pub ascent: f32` and `pub descent: f32` to both TextChar and FontInfo.

`TextChar::ascent/descent` are in device space (the same units as advance_width, rendered_advance, and the origin coordinates), so callers can use them directly:

```
  top_of_glyph    = origin_y + ascent
  bottom_of_glyph = origin_y + descent
```

These fields exist because bbox is insufficient for vertical glyph extent: bbox.y is the baseline and bbox.height is the full em square, so it neither captures the descender region (below the baseline) nor the font's actual cap height (which is typically ~70-90% of the em). ascent and descent give the font-metric-specified extents above and below the baseline separately.

`FontInfo::ascent/descent` are stored as fractions of the em (the raw PDF `/Ascent` and `/Descent` values divided by 1000), since `FontInfo` is parsed once per font without access to a specific glyph's CTM. The device-space scaling is applied per-character at `TextChar` construction time.

Values are sourced from the PDF font descriptor (`/Ascent` and `/Descent` keys).  Descent values stored as positive magnitudes are normalised to negative.  For fonts without a descriptor, the 14 standard PDF fonts are looked up by name using Adobe AFM values; if that also fails, the fallback values 0.95 and -0.35 are used (matching Poppler's defaults).

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Changes Made

- Added `TextChar::ascent/descent`
- Added code to parse this info out of the font.
- Added code to populate the new struct members when yielding characters.

## Testing

- [X] I have run `cargo fmt`

## Documentation

## Checklist


yfedoseev (changes requested):
Thanks for this, @haberman — adding real `/Ascent` `/Descent` to `TextChar`/`FontInfo` is a genuinely useful addition, and the simple-font path is done right: the scaling math is sound (`em_fraction * font_size * sqrt(d² + b²)` keeps the metric in the same device-space units as the glyph bbox height), the standard-14 fallback table is plausible AFM-sourced, the docs are clear, and the 31-file blast radius is the unavoidable cost of adding non-defaulted fields to two widely-constructed public structs — not noise. The bulk of this is ready.

I'd like to request one change before it lands, plus a couple of smaller items:

**Blocker — Type0/CID fonts never load real ascent/descent.** A Type0 font dict has no top-level `/FontDescriptor`; the `/Ascent` and `/Descent` live on the **descendant CIDFont's** FontDescriptor (PDF 32000-1 §9.7.4 / Table 117, see `docs/spec/pdf.md:18599`). `parse_descendant_fonts` (`src/fonts/font_dict.rs:1010`) already drills into that descriptor for embedded font data, widths, and the cmap — but this PR doesn't extend it to read ascent/descent. As written, `raw_ascent`/`raw_descent` are always `None` for Type0, and the CID `base_font` won't match `standard_font_metrics()`, so **every CID/CJK glyph silently gets the 0.95 / -0.35 fallback** regardless of the real descriptor. For CJK documents that's the common case, not an edge case, and the new tests only cover simple Type1 fonts so it's both untested and unflagged. Could you thread descendant ascent/descent out of `parse_descendant_fonts` and prefer them when the top-level descriptor is absent, with a Type0 test?

**Minor — synthetic TJ-space char ignores font metrics.** The injected-space path for large negative TJ offsets (`src/extractors/text.rs:4769`) hardcodes `0.95`/`-0.35` even though `font` is already in scope (used for weight a few lines up). The real-glyph path correctly uses the font's metrics, so a consumer computing a line bbox will see a discontinuity at injected spaces. Trivial to align.

**Minor — CHANGELOG note for the API change.** The two public structs gain `pub` fields without `#[non_exhaustive]`, which is source-breaking for any downstream literal constructor. Worth a line in the CHANGELOG.

Once the Type0 path reads the descendant descriptor (or the limitation is documented loudly on the fields + CHANGELOG), I'm happy to take this into the v0.3.59 release. Appreciate the contribution.

haberman:
I made the requested changes, but it looks like there is a changelog conflict now.

Would you be able to resolve this conflict (I notice you usually rewrite the changelog entries a bit anyway)?

yfedoseev (approved):
All three review items resolved and verified.

- **Type0/CID blocker**: `d65aa6ed` threads `/Ascent` `/Descent` off the CIDFont descendant's FontDescriptor (§9.7.4) via `read_raw_ascent_descent_from_descriptor`, merged with the top-level descriptor via `.or()` so a (rare) top-level descriptor still wins. The descendant values flow through the same `/1000`-em normalization and positive→negative descent fix as the simple-font path. Covered by `test_type0_ascent_descent_from_descendant_descriptor`.
- **Synthetic TJ-space char**: now uses `font.ascent`/`font.descent` instead of hardcoded 0.95/-0.35, matching the real-glyph scaling convention.
- **CHANGELOG**: source-breaking `pub` field note added under a Changed section.

Rebased onto current main (CHANGELOG conflict resolved — ascent/descent entries sit under `next-version` above the released 0.3.59 section). Net diff vs main is exactly the ascent/descent feature. Compiles clean against main; `test_ascent_descent_font_metrics` (4/4) and the Type0 lib test pass.

Thanks @haberman — happy to take this into the next release.

PR #629, opened 2026-06-03
feat: cff encoding fix

RayVR:
## Description

Builds the CFF subset font's GID map from the PDF font dictionary's `/Encoding`
entry, then resolves glyph names via the CFF Charset. Per ISO 32000-1 §9.6.6,
`/Encoding` (typically `WinAnsiEncoding` plus optional `/Differences`) is the
authoritative byte → glyph-name source for simple fonts; the CFF Encoding table
inside a subset is often sparse and must not be treated as authoritative.

The prior code used the CFF Encoding table directly, which on subsets emitted
by aggressive subsetters maps only the handful of bytes whose name happens to
match a Standard Encoding entry. Every other byte in the content stream
resolved to GID 0 (`.notdef`) and was painted as a single fallback glyph.

## Type of Change

## Related Issues

Relates to text-extraction and rendering correctness on packaging-artwork PDFs
that ship subset CFF (Type 0 / CIDFontType0) fonts.

## Changes Made

- Add `parse_cff_gid_mapping_with_pdf_encoding` in `src/fonts/cff_encoding.rs`
  that walks the PDF `/Encoding` (BaseEncoding + `/Differences`) for each of
  the 256 single-byte codes, then resolves the resulting glyph name to a GID
  via the CFF Charset.
- Add pure helper `resolve_bytes_via_pdf_encoding` exercised directly by the
  unit tests so the encoding-resolution logic is testable without a full font
  fixture.
- Wire `FontInfo::from_dict` to the new entrypoint in `src/fonts/font_dict.rs`
  so both rendering paths (page renderer and separation renderer) pick up the
  fix transparently.
- Document the byte → GID resolution chain in `src/rendering/text_rasterizer.rs`
  so the next reader sees why CFF Encoding is not authoritative for subsets.

## Testing

- [x] I have run `cargo fmt`

Four new unit tests cover:

1. Every Charset glyph round-trips through `/Encoding` resolution.
2. `/Differences` overrides are honored ahead of `BaseEncoding`.
3. Identity / no-`/Encoding` font dictionaries short-circuit correctly.
4. Custom-string SIDs (non-Standard names) resolve via the CFF strings index.

5431 lib tests pass on the branch; 80 `cff_encoding` unit tests pass.

Black-plate coverage on a packaging-artwork fixture rose from 1.52% to 3.48%
at 150 DPI — every byte the prior code mapped to `.notdef` now paints its
intended glyph.

- [x] Python bindings updated (if needed) — no surface change; fix is internal
- [x] Python tests pass

`maturin develop --release --features python,ocr,barcodes` produces a working
wheel locally.

## Documentation

Inline rustdoc on `parse_cff_gid_mapping_with_pdf_encoding`,
`resolve_bytes_via_pdf_encoding`, and the dispatch comment in
`text_rasterizer.rs` explain the spec basis and why the CFF Encoding table
must not be consulted directly for subsets.

## Checklist

Before / after on the same packaging-artwork fixture: badge text like
"OVEN BAKED", "NO ADDED SUGAR", "HIGH-GRAIN", and the nutrition info pills
("10%", "3%", "6%", "4%") now render as the correct letterforms on every
separation plate, where previously they appeared as a single repeated fallback
glyph.

## Additional Notes

- Branches off `main`; no dependency on other in-flight branches.
- Incidentally fixes `office_oxide` font re-embedding, since
  `extract_embedded_fonts_with_unicode_maps_and_widths` reads the same
  `cff_gid_map` produced by this path.
- Scope intentionally narrow: the prior code path remains for fonts that
  legitimately have no PDF `/Encoding`; only the lookup order changed.


yfedoseev (commented):
Thank you for this, @RayVR — really nice contribution! 🙏 The diagnosis is spot-on and the fix is grounded exactly where it should be: per ISO 32000-1 §9.6.6 the PDF font dictionary's `/Encoding` is the authoritative byte→glyph-*name* source for simple fonts, and the CFF Charset resolves name→SID→GID. Routing through `/Encoding` + `/Differences` instead of the often-sparse internal CFF Encoding table is the correct model, and I really appreciate the care in the fallback chain (Identity / predefined-charset / empty-result → legacy path) so the change can never make a currently-working font worse. The four targeted tests and the black-plate coverage measurement are a great touch.

Two small spec-fidelity notes for a follow-up — neither is a blocker, since this is a strict improvement over the previous behaviour:

**1. Base-encoding selection (§9.6.6.1 / Annex D).** `resolve_bytes_via_pdf_encoding` resolves base bytes via `gid_to_standard_glyph_name` (WinAnsi names) for both `Encoding::Standard(name)` and `Encoding::Custom(_)`, regardless of the declared base. WinAnsi / MacRoman / StandardEncoding are identical across ASCII but diverge in 0x80–0xFF (Annex D), and §9.6.6.1 makes the base selectable (`/BaseEncoding`, or an implicit StandardEncoding for a non-embedded nonsymbolic font). So a font declaring e.g. `/BaseEncoding /MacRomanEncoding` could mis-resolve high bytes. The packaging-artwork fixture here is ASCII-only so it's unaffected — just flagging it so the high-byte case is on the radar (either select the name table per declared base, or a short scoping note).

**2. Charset parse capped at 256 entries.** If a subset enumerates more than 256 glyphs and a `/Differences`-referenced glyph happens to sit at GID > 256, it won't land in `name_to_gid` and would fall to `.notdef`. Simple fonts only address ≤256 codes, but their GIDs can exceed 256 — might be worth parsing the full `nGlyphs` charset rather than a fixed 256.

Thanks again — this is a clean, well-reasoned fix. 👍

yfedoseev (commented):
Re-reviewed the new commits — thank you, this fully addresses both points. 🙏

- **Per-base-encoding name table (§9.6.6.1 / Annex D):** `MacRomanEncoding → mac_roman_byte_to_name`, `StandardEncoding → standard_encoding_byte_to_name`, else WinAnsi is exactly right — the 0x80–0xFF divergence is now resolved per the declared base. The honest note that `Encoding::Custom` loses the named base and defaults its high bytes to WinAnsi (threading `/BaseEncoding` through as a follow-up) is a reasonable, well-documented boundary.
- **Full `nGlyphs` charset:** reading the count from the CharStrings INDEX header (CFF §9) and parsing the whole charset means a `/Differences` glyph at GID > 256 now resolves. 👍

Spec-aligned and well-tested. LGTM from my side.

PR #633, opened 2026-06-03
feat(paths): add PathContent::to_points for path flattening (#147)

### LINKED ISSUE #147: Path-to-points flattening for signal processing / medical PDF use cases
## Use case

A user is building an [ECG PDF-to-EDF converter](https://github.com/joelparkerhenderson/kardiamobile-1l-ecg-convert-pdf-to-edf-rust-crate) that extracts vector path graphics from KardiaMobile ECG PDFs and converts the waveform coordinates into voltage/time measurements.

Their current pipeline uses lopdf to manually parse content stream operators, track graphics state, and extract path coordinates — all of which `extract_paths()` already handles. But one thing that would make this significantly easier:

## Feature: `PathContent::to_points(tolerance)`

Flatten a path into a sequence of `(x, y)` points by:
- Keeping `MoveTo` and `LineTo` coordinates as-is
- Sampling `CurveTo` (cubic Bezier) into line segments using a configurable flatness tolerance
- Returning a `Vec<Vec<(f32, f32)>>` (one inner vec per subpath)

This is useful for any signal processing pipeline where you need discrete point samples from PDF vector graphics — ECG traces, chart digitization, CAD extraction, etc.

### Example API

```rust
let paths = doc.extract_paths(1)?;
for path in &paths {
    let subpaths = path.to_points(0.5); // tolerance in PDF points
    for points in &subpaths {
        for (x, y) in points {
            // process as signal data
        }
    }
}
```

## Workaround

Users can already iterate `path.operations` and handle `CurveTo` manually with De Casteljau subdivision, but a built-in method would make this much more accessible — especially for the growing number of "extract data from chart PDFs" use cases.

mbeschastn0v:
## What

Adds `PathContent::to_points(tolerance) -> Vec<Vec<(f32, f32)>>` — flattens an extracted vector path into polylines (one inner vec per subpath), so consumers get sampled `(x, y)` coordinates instead of raw drawing operators.

```rust
let paths = doc.extract_paths(page)?;
for path in &paths {
    for subpath in path.to_points(0.5) {   // 0.5 pt tolerance
        for (x, y) in subpath {
            // feed straight into signal processing
        }
    }
}
```

- `MoveTo`/`LineTo` pass through unchanged.
- Cubic Béziers (`CurveTo`) are **adaptively subdivided** (recursive de Casteljau, flatness measured as control-point distance to the chord) so the polyline stays within `tolerance` of the true curve.
- Non-positive / non-finite `tolerance` is floored, and recursion is depth-capped, so subdivision always terminates.

## Why

Closes #147. Vector graphics come back from `extract_paths` only as raw operators; there is no built-in way to turn `CurveTo` into points, forcing every "digitise a chart/ECG/CAD trace" user to re-implement Bézier subdivision. This adds the missing primitive.

## Spec compliance (ISO 32000-1:2008 §8.5.2, Table 59)

Subpath/operator semantics follow the spec text in `docs/spec/pdf.md`:

- `re` is a complete closed 5-point subpath, matching the spec's literal `m / l / l / l / h` equivalence; the current point afterwards is `(x, y)`.
- `h` closes **and terminates** the subpath — a following segment begins a new one.
- A lone or overridden `m` leaves "no vestige" (degenerate sub-2-point subpaths, which paint nothing, are dropped).
- Curve `P0` is the implicit current point.

One intentional deviation: §8.5.2.1 says an undefined current point should raise an error; this convenience method instead tolerates it (returns `Vec`, not `Result`), since real `extract_paths` output always establishes `P0` via `m`/`re` first.

## Testing

- **Unit tests** (`src/elements/path.rs`): straight-line passthrough, curve endpoint preservation, a rigorous *every-true-curve-point-within-tolerance-of-the-polyline* oracle, tolerance monotonicity, current-point continuity, rectangle/closepath structure, multi-subpath splitting, the three §8.5.2 operator-semantics rules above, empty path, non-positive/NaN tolerance termination, degenerate curve.
- **Doctest** on the example above.
- **Integration invariants** (`tests/test_path_to_points.rs`) against the committed curve-heavy fixtures (`1008.3918v2.pdf` ~281k curves, `1.pdf`): all points finite, within bbox, bounded count, and curves are actually densified.
- `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` all clean.

## Type of Change

- [x] New feature (purely additive — new method, no struct/API changes)

## Checklist

- [x] Tests pass
- [x] Code formatted (`cargo fmt`)
- [x] Clippy clean (`-D warnings`)
- [x] Docs updated (doc comment + CHANGELOG under `next-version`)
- [x] DCO sign-off


yfedoseev (approved):
Thank you for this — it's a genuinely high-quality contribution. 🙏 The de Casteljau flattening is correct (exact midpoint subdivision + a conservative control-point-to-chord flatness test, with the `p0 == p3` loop case handled and a depth cap for termination), the §8.5.2 modelling is faithful (`re` as a closed 5-point subpath, `h` close-and-terminate, the `m`-no-vestige drop of degenerate subpaths, implicit `P0`), and the test suite is exemplary — the *every-true-curve-point-within-tolerance* oracle plus the real 281k-curve fixture integration is exactly the right way to pin this. Purely additive, KISS, clippy/fmt/doc clean.

A few optional, non-blocking notes:

1. **Tolerance coordinate space.** `tolerance` is in the same space as `extract_paths` output (PDF user space, pre-CTM). Worth a line in the doc comment so a caller working with a scaled Form-XObject `/Matrix` knows to scale the tolerance accordingly.
2. **Open vs. closed subpath.** `Vec<Vec<(f32, f32)>>` can't distinguish a closed subpath (which ends with the start point repeated, via `h` / `re`) from an open polyline that happens to return to its start. That's fine for the digitise-a-trace use case in #147; if a fill/region consumer ever needs it, a parallel `is_closed` signal (or a small struct) would be a clean additive follow-up — not needed here.
3. **(FYI, not a request)** For the very densest curve sets, an analytic flattener (e.g. the Euler-spiral approach in `kurbo`) yields fewer points for the same tolerance — but it's far more code and a dependency; adaptive de Casteljau is the right call for a self-contained primitive.

LGTM — approving. Thanks again for the thoroughness; the spec citations and the oracle test made this easy to trust. 🎯

PR #634, opened 2026-06-04
feat: render smask

RayVR:
## Description

Closes the §11.6.5 (soft masks), §11.6.6.2 (knockout groups), and §11.3.5.3 (non-separable blend modes) gaps in the page renderer. Before this PR the renderer silently degraded all three: ExtGState `/SMask` was ignored, `/K knockout` groups rendered as non-knockout, and `BM=Hue|Saturation|Color|Luminosity` collapsed to `SourceOver`. Real-world artwork — drop shadows, vignettes, packaging overlays, translucent-overlap effects — was visibly wrong.

Composite-renderer scope only. Separation-renderer integration is deferred to a follow-up because the correct path is composite-then-separate (per industry convention for prepress-grade RIPs) and is a larger architectural piece than fits this PR.

## Type of Change

- [x] Performance improvement
- [x] Tests

## Related Issues

Relates to the rendering-fidelity gap surfaced in earlier prepress audit work.

## Changes Made

**ExtGState `/SMask` (§11.6.5.2 + §11.6.5.3) — page renderer**
- Parser captures `/SMask` as a `SoftMaskSpec` (`None` or `Dict`); apply-time is pure.
- Renderer maintains `soft_mask_stack` mirroring `clip_stack`; q/Q push and pop in lockstep. `/SMask /None` clears the stack slot.
- `materialise_soft_mask_alpha` renders the `/G` group into an offscreen pixmap at the install-time CTM, then extracts the mask buffer:
  - `/S /Alpha`: alpha channel.
  - `/S /Luminosity`: BT.601 luma of premultiplied RGB. `/BC` (backdrop colour) is parsed against the group `/CS` (DeviceGray / DeviceRGB / DeviceCMYK with naive uncalibrated conversion) and pre-filled into the offscreen pixmap so unpainted areas contribute the right luminance.
  - `/TR` (transfer function) applied pointwise after the subtype buffer: Type 2 exponential supported (`y = C0 + xᴺ · (C1 − C0)`, rejects `N ≤ 0` and multi-component `/C0` / `/C1`); `/Identity`, missing `/TR`, and other function types are no-ops.
- `effective_clip` helper returns the per-pixel min of the active clip mask and the active soft mask (§11.3.4 shape × opacity). Replaces the 10 paint-site clip lookups.
- Hard cap of 32 nested SMask materialisations against cyclic `/G` references.
- Cache for materialised alpha keyed by ExtGState dict_name with bitwise install-time CTM validity check. Scoped to one `execute_operators` invocation (no cross-page / cross-document leakage by construction).

**Knockout transparency groups `/K` (§11.6.6.2) — page renderer**
- `render_form_xobject` parses `/K` from the form's `/Group` dict (resolves indirect `/Group N 0 R` references; accepts boolean `true` and non-zero integer for both `/K` and `/I`). When true, snapshots the group's initial pixmap state and passes it as `knockout_backdrop` into the recursive `execute_operators`.
- New `knockout_aware_paint(pixmap, backdrop, alpha, fn)` wraps each paint primitive. When backdrop is `Some` AND `effective_alpha < 0.9999`, the paint targets a fresh clone of the backdrop and the result merges via `knockout_merge` (per-pixel: `temp[i] != backdrop[i]` → copy temp into dest).
- `knockout_paint_alpha(gs_alpha, blend_mode)` gates the alpha short-circuit on `BM=Normal`. Opaque paints with non-Normal blend modes (`Multiply`, `Hue`, `Saturation`, `Color`, `Luminosity`, etc.) still redirect to the backdrop because the blend formula reads the destination.
- All 10 paint sites covered: stroke, fill, fill+stroke (both passes), fill even-odd, four text variants (Tj/Quote/TJ/DoubleQuote), Do (image + form), shading.

**Non-separable blend modes (§11.3.5.3) — page renderer**
- `pdf_blend_mode_to_skia` previously fell through to `SourceOver` for `Hue` / `Saturation` / `Color` / `Luminosity`. The underlying `tiny_skia` already ships those as native `BlendMode` variants implementing the W3C / Acrobat formulas; the fix is four new match arms.

**Robustness, perf, and review-fix work bundled here**
- `/SMask` parsing handles indirect `/Resources` references, missing `/S` (warn-and-skip), unknown subtypes (warn-and-skip), `/Group /CS` as array form (e.g. `[/ICCBased <ref>]` resolves to base name).
- Skip the doubled per-SMask page-sized pixmap allocation (~33 MB saved per SMask at 300 DPI A4).
- `as_f64` helper dedupes the `Object::Real | Integer → f64` pattern across the new parsers.

## Testing

- [x] I have run `cargo test --features rendering`
- [x] I have run `cargo clippy --all-targets --features rendering -- -D warnings`
- [x] I have run `cargo fmt`

**25 new integration tests** across three files, designed to fail closed against the specific bug classes a buggy implementation could ship under:

`tests/test_smask_alpha.rs` (19):

- `ext_gstate_alpha_smask_blocks_paint_under_transparent_mask` — basic Alpha mask, top opaque / bottom transparent.
- `ext_gstate_smask_none_clears_active_soft_mask` — `/SMask /None` clears; first fill still landed.
- `ext_gstate_smask_survives_q_save_restore` — mask persists through `q`/`Q` push and pop.
- `ext_gstate_alpha_smask_honours_install_time_ctm` — `cm` before `gs` correctly scales the mask.
- `ext_gstate_smask_cyclic_g_does_not_stack_overflow` — adversarial self-referential `/G` is depth-capped.
- `ext_gstate_luminosity_smask_modulates_paint_by_group_luma` — mid-grey `/G` composes ~50% black-over-white.
- `ext_gstate_luminosity_smask_bc_white_backdrop_passes_paint` — `/BC [1 1 1]` backdrop unblocks paint where `/G` is unpainted.
- `ext_gstate_luminosity_smask_tr_type2_squared_attenuates_paint` — `/TR` `N=2` squares the mask (0.5 → 0.25).
- `ext_gstate_luminosity_smask_group_cs_devicegray_yields_50pct_paint` — `/CS /DeviceGray` valid case behaves like no-`/CS`.
- `ext_gstate_luminosity_smask_malformed_devicegray_with_rgb_paint_uses_bt601` — pins the BT.601-unconditional choice for malformed `/CS`.
- `smask_clips_image_paint` — image `Do` clipped by active SMask.
- `smask_clips_text_paint` — text glyphs in the masked-off region don't render.
- `bt601_luma_pure_red_yields_r_approx_179` — pins the red weight position in BT.601 (a swap to green-dominant gives R ≈ 105).
- `bt601_luma_pure_green_yields_r_approx_105` — pins the green weight position.
- `bt601_luma_pure_blue_yields_r_approx_226` — pins the blue weight position. Together these three uniquely identify the BT.601 coefficient triple; any pairwise swap fails ≥ 2 tests.
- `tr_type2_squared_pins_exponent_at_three_points` — multi-samples `/TR Type 2 N=2` at gray = 0.25, 0.5, 0.75. A single-point assertion would admit `x¹`, `x³`, `x⁴`; three points uniquely identify the exponent.
- `tr_type2_invalid_n_falls_through_to_identity` — actually exercises the `N ≤ 0` rejection path.
- `smask_cache_invalidates_when_ctm_changes` — same `/GS1` invoked at two different CTMs in one content stream. A cache that skipped the install-transform check would serve the stale identity-CTM mask on the second invocation.
- `smask_applies_to_paint_inside_nested_do` — page invokes a Form via `Do`; the Form's own content stream installs `/GS1` (SMask). Tests that nested-context SMask correctly clips the inner form's paints.

`tests/test_knockout_group.rs` (4):

- `knockout_group_blue_replaces_red` — two 50%-alpha overlapping rects with `/K true` produce blue-over-white; `/K false` produces red+blue blend. Asserts G and B channels each diverge by ≥ 30 levels.
- `knockout_group_opaque_non_normal_blend_redirects_to_backdrop` — opaque red + opaque blue with `/BM /Multiply` in a knockout group. Verifies the alpha short-circuit gates on `BM=Normal`.
- `knockout_group_pixel_exact_replacement` — same fixture rendered with the knocked-out red present vs absent; asserts byte-identical pages across the entire pixmap. Pins the spec-defining invariant of §11.6.6.2 — half-implemented knockouts that produced channel deltas under the +30 thresholds don't survive byte equality.
- `knockout_redirects_all_four_non_separable_blend_modes` — parameterised across `Hue`, `Saturation`, `Color`, `Luminosity`. A bug that only special-cased some mode names would slip through.

`tests/test_non_separable_blend_modes.rs` (1):

- `luminosity_blend_mode_does_not_overwrite_with_source` — red-then-blue layered paint with `BM=Luminosity` on the upper layer. Asserts B < 200 (a SourceOver fallback would give B = 255).

- [x] Python bindings updated (if needed) — no surface change; all behaviour is internal to the rendering pipeline.
- [x] Python tests pass

`maturin develop --release --features python,ocr,barcodes` produces a working wheel locally.

## Documentation

Inline rustdoc on `materialise_soft_mask_alpha`, `knockout_aware_paint`, `knockout_merge`, `knockout_paint_alpha`, `effective_clip`, `parse_soft_mask_transfer`, `parse_soft_mask_backdrop`, and `pdf_blend_mode_to_skia` documents the spec basis and intentional out-of-scope cases (CMYK blend-space approximation for `/Group /CS`, full ICC for `/BC`, byte-equality limitation of `knockout_merge`).

## Checklist

N/A — visible effects are spec-conforming compositing changes. Synthetic test fixtures pin pixel-level expectations.

## Additional Notes

**Out of scope for this PR (tracked for follow-up):**

- **Separation-renderer integration.** Per earlier research, the correct path is composite-then-separate — never per-plate alpha multiplier, which would put shadow tint on plates the colour never referenced. Architectural piece, not a paint-site change.
- **`/Group /CS`-aware luma dispatch.** The renderer rasterises everything to RGB; a spec-correct CMYK or wide-gamut blend space requires a non-RGB pixmap path. BT.601 on the rasterised RGB is the documented approximation.
- **Non-Normal blend modes inside knockout groups** beyond the redirect dance. Real-world output from authoring tools is ambiguous about backdrop-relative blending semantics; reference fixtures would help pin the right answer.
- **Nested non-isolated subgroup whose parent is knockout.** The recursive `execute_operators` passes `None` for nested forms, so the inner form renders normally. Correct for the common case.
- **`/TR` function types 0 (sampled), 3 (stitching), 4 (PostScript).** Treated as identity; Type 2 covers the overwhelming majority of real-world `/TR`.
- **`/BC` for non-Device colour spaces.** `Lab`, `ICCBased`, `CalRGB` etc. fall back to component-count inference. Spec-correct ICC profile evaluation is out of scope.
- **Page-level `/Group /K`.** The page's own group dict is read for `/S /Transparency` but `/K` on the page-level group isn't applied at the top-level render. Rare in practice; a future fix would propagate it through `render_page_with_options`.
- **Per-paint pixmap clone in `knockout_aware_paint` is O(W·H).** Plausible follow-up: hoist a single scratch pixmap across the knockout group and reuse it via `copy_from_slice`. For the dense-plot pathological case this is a meaningful speedup; for typical artwork it is not visible.
- **`knockout_merge` byte-equality limitation.** A paint that legitimately reproduces backdrop bytes (e.g. drawing white onto white-backdrop) is treated as untouched and the prior paint's value survives. Rare in real artwork; documented in the helper's docstring.

**TODO follow-up PR (refactor, no behaviour change):**

`src/rendering/page_renderer.rs` is now 3360+ lines, with `execute_operators` alone at ~1280 lines. A pure-code-movement pass — extracting the four `Set{Fill,Stroke}Color{,N}` match arms into named methods, plus the `SetExtGState` handler and the path/text paint dispatch — drops `execute_operators` by ~500 lines and the file by ~10% with zero behaviour change. Each extraction is independently reviewable. Will land as a separate PR off `main` after this one merges.


yfedoseev:
Thanks for this! The remaining CI failures (CI / **Clippy**, **Lint and Format Check**, **FIPS CryptoProvider**, **WASM Build**) all come from a single deterministic Clippy lint — `clippy::doc_lazy_continuation`, fatal here under `-D warnings`. A re-run won't help; it'll fail identically until the doc comment is adjusted.

Note the FIPS job's actual tests all pass (5563 passed / 0 failed) — only the clippy step trips, so the code itself is fine. It's purely a doc-formatting nit.

**Location:** `src/rendering/page_renderer.rs:2699–2701`

The nested `*` bullet list is followed by a continuation paragraph that's under-indented relative to the list item:

```rust
///       * Malformed groups (e.g. `/CS /DeviceGray` with RGB paint
///         operators) get BT.601 on the actual RGB; see
///         `tests/test_smask_alpha.rs::...`.
///     A proper `/CS` dispatch would need a non-RGB blend buffer   // <- 2699
///     (separate gray / CMYK pixmaps) which the renderer does not  // <- 2700
///     currently provide.                                          // <- 2701
```

**Fix:** align those 3 continuation lines under the bullet's text (indent to match the `Malformed groups` text), or drop them to a standalone top-level paragraph with a blank `///` line before them. Either clears all four jobs at once.

Quick local check before pushing: `cargo clippy --all-targets -- -D warnings`. Thanks again!

RayVR:
Fixed in 6d14b95. Broke the closing `/CS`-dispatch architectural note out
of the bullet list with a blank `///` separator — semantically right
(it applies to the whole `/Luminosity` block, not just the malformed-
groups sub-bullet that preceded it), and it satisfies
`clippy::doc_lazy_continuation` under `-D warnings`.

Verified locally:
- `cargo clippy --features rendering --all-targets --no-deps -- -D warnings`: clean
- `cargo clippy --target wasm32-unknown-unknown --no-default-features --features wasm,rendering,barcodes --lib --no-deps -- -D warnings`: clean
- `cargo check --lib --no-default-features`: clean
- `rustfmt --check src/rendering/page_renderer.rs`: clean

RayVR:
Closing this PR — the features it ships have been re-implemented on `feature/transparency-flattening` atop the post-refactor architecture (yfedoseev/pdf_oxide#649 is now merged). That branch is **in progress** (round 3 of a multi-round design/QA/fix loop is running as I post this); the upstream PR will follow once the work stabilizes.

SMask is part of Chapter 11 transparency (§11.6.5), as are knockout `/K` (§11.4.6.2) and non-separable blend modes (§11.3.5.3). It made sense to fold all of this PR's features into a single transparency implementation pass on top of the new pipeline rather than rebase and ship piecemeal.

## Why the re-implementation rather than a rebase

This PR was authored against the pre-refactor `page_renderer.rs` inline-arm pattern. The renderer-resolution-pipeline migration (#649) replaced that pattern with a `ResolutionPipeline` + `PaintBackend` shape. Conflict resolution here would have meant rewriting each implementation against the new shape — which is effectively what I did on the new branch, with the addition of a feature-coverage audit that surfaced gaps this PR didn't catch:

- `qa_round1_form_xobject_smask_alpha`: §11.4.7 SMask /S /Alpha on Form XObjects (this PR's `94912f3`) — re-implemented as part of round 2 gap 3
- `qa_round1_form_xobject_smask_luminosity`: §11.4.7 SMask /S /Luminosity (this PR's `c0ab679`) — re-implemented, with explicit BT.601 luma reference values
- `qa_round1_smask_bc_tr`: SMask /BC + /TR (this PR's `da0b0d7`) — re-implemented with Type-2 transfer functions evaluated per pixel
- `qa_round1_nonseparable_blend_modes`: §11.3.5.3 non-separable blend modes (this PR's `4849dec`) — round 1 audit found these were silently degrading to `SourceOver` in `tiny_skia`; round 2 implemented them out-of-band in HSL/HSY space per spec, not as tiny_skia BlendMode variants (which don't expose Hue/Saturation/Color/Luminosity natively)
- `qa_round1_knockout_group_K`: §11.4.6.2 `/K` knockout (this PR's `d069ae2`) — re-implemented via per-paint-operator backdrop-replay segmentation

## Tests worth preserving

The implementations are superseded but the test scenarios in this PR's later commits — particularly `1084cfe` (SMask under nested Do + CTM cache invalidation), `87457d4` (SMask clipping across image and text paints), `17cee28` (spec-compliance + malformed-input hardening), and `4d82947` (review-feedback test additions) — pin properties that aren't yet probed in the new audit suite. I've filed a follow-up to cherry-pick or rewrite them against the new architecture so the coverage doesn't regress.

Thanks for the review work on this — the spec-compliance + malformed-input hardening commit in particular shaped how I structured the audit-first approach on the new branch.

PR #645, opened 2026-06-04
feat: render wmode vertical

RayVR:
## Description

Closes the §9.4.4 / §9.7.4.3 / §9.7.5 gap in vertical writing mode handling. Before this PR, `Identity-V` and other `-V`-suffixed CMaps were recognised as encoding *names* and CID→Unicode mapping worked, but every advance was applied along the x-axis unconditionally — the extractor, the page renderer, the separation renderer, the text rasterizer, and the reading-order pipeline all treated vertical content as horizontal. The result: Japanese tategaki, Traditional Chinese vertical packaging, and similar real-world prepress artwork extracted characters correctly but positioned and ordered them as a jumbled horizontal stream. Glyphs rendered along x instead of y, multi-Tj/TJ blocks drifted across the page, and `to_markdown` / `to_html` destroyed column order.

Single-decision-point design: WMode is resolved once at font load, cached onto `GraphicsState.text_wmode: u8`, and consumed by exactly one axis-swap site (`GraphicsState::advance_text_matrix(displacement)`). Every advance path — extractor (`show_text`, `Tj`, `TJ`, `Td`/`TD` cm semantics), page renderer, separation renderer, text rasterizer — routes through that helper. Horizontal text pays one predicted-not-taken branch per advance; the smell pattern `Matrix::translation(advance, 0.0)` is grep-zero across the source tree.

## Type of Change

- [x] Performance improvement
- [x] Code refactoring
- [x] Tests

## Related Issues

Relates to the rendering-fidelity gap surfaced in earlier prepress audit work.

## Changes Made

**CMap `/WMode` directive parsing (§9.7.5.4)**
- `src/fonts/cmap.rs` — `parse_wmode_directive` reads `/WMode N def` from a CMap stream, default 0. PostScript `%`-to-EOL comments are stripped before the regex so a `/WMode 1 def` inside a commented-out block doesn't false-fire. Non-spec values (`/WMode 2`, etc.) log a `warn` and fall back to horizontal rather than silently dropping the directive.

**CIDFont `/W2` and `/DW2` parsing (§9.7.4.3)**
- `src/fonts/font_dict.rs` — `parse_w2_array` accepts both spec forms in one array and round-trips them through to a `HashMap<u16, VerticalMetrics>`:
  - Form A: `c [w1y v_x v_y w1y v_x v_y …]` — successive CIDs from `c`.
  - Form B: `c_first c_last w1y v_x v_y` — range of CIDs sharing one triple.
  - Mixed forms in a single `/W2` array supported.
  - Malformed inner-triple elements drop the *whole triple* rather than advance by one (preserves CID alignment across subsequent entries).
  - u16 overflow at the upper end of a range breaks the loop with a warn rather than collapsing all later entries onto u16::MAX via `saturating_add`.
- `parse_dw2` reads `[v_y w1y_default]` with spec defaults (`v_y = 880`, `w1y = -1000`) when absent or malformed.

**WMode resolution at font load with strict precedence**
- `src/fonts/font_dict.rs` — `FontInfo.wmode` resolved with this order:
  1. `/Encoding` CMap stream `/WMode 1 def` (active CMap)
  2. `/Encoding` CMap name `-V` suffix (Identity-V, UniJIS-UTF16-V, UniGB-UTF16-V, UniCNS-UTF16-V, UniKS-UTF16-V; bare legacy `V`)
- ToUnicode CMap `/WMode` is *not* consulted (§9.10.2: ToUnicode is for extraction-time character-to-Unicode mapping only). A disagreement between ToUnicode `/WMode` and `/Encoding` `/WMode` is logged at debug level. This closes a spec violation where a horizontal document with a stale `/WMode 1 def` in its ToUnicode (real-world tooling leftover) could silently flip to vertical.

**Single-decision-point axis swap**
- `src/content/graphics_state.rs` — `GraphicsState::advance_text_matrix(displacement)` owns the §9.4.4 axis-swap math. In horizontal mode it advances `(Tm.e, Tm.f) += displacement × (a, b)`; in vertical it advances `(Tm.e, Tm.f) += displacement × (c, d)`. Rotated text matrices work correctly because the helper multiplies the *active row* of `Tm`, not a hard-coded axis.
- `GraphicsState.text_wmode: u8` cached from the active font at `Tf`. q/Q save/restore preserves it; mid-BT `Tf` font switches between horizontal and vertical fonts are correctly tracked.

**Extractor — advance helpers + show_text + TJ-offset**
- `src/extractors/text.rs` — `advance_position_for_string`, `advance_position_for_offset`, `append_and_advance`, the Tj/TJ matrix-update path, and the `show_text` post-glyph advance all route through `gs.advance_text_matrix(...)`.
- Horizontal scaling `Tz` is dropped from vertical advance computation (§9.4.4: Tz is "horizontal direction of glyph stretching", not "writing-direction scale"). The vertical-mode formula is now `ty = (w1y − Tj/1000) × Tfs + Tc + Tw` — no Tz factor.
- Synthetic-space-from-large-negative-offset bbox geometry is wmode-aware (width in horizontal, height in vertical).

**Page renderer and separation renderer**
- `src/rendering/page_renderer.rs` — five inter-Tj advance sites (Tj, Quote, TJ, DoubleQuote × 2) substituted `Matrix::translation(advance, 0.0)` → `gs.advance_text_matrix(advance)`.
- `src/rendering/separation_renderer.rs` — six advance sites (path stroke/fill plus four text variants) same substitution.

**Text rasterizer**
- `src/rendering/text_rasterizer.rs` — per-glyph advance in `render_unicode_text` runs along y in vertical mode; per-glyph `(v_x, v_y)` from `cid_vertical_metrics` applies a `(-v_x/1000, -v_y/1000) × font_size` origin shift before painting. `render_tj_array` and `measure_tj_array` carry their own per-glyph cursors and route the aggregated advance through `gs.advance_text_matrix`.
- `vrt2` / `vert` GSUB glyph rotation is *not* applied — the rasterizer paints horizontal glyph shapes at vertical positions. This is correct per spec (PDF metrics are authoritative) but means CJK glyphs that have rotated alternates in the font (commas, brackets, dashes) won't display rotated. See Additional Notes.

**Pipeline — tategaki reading-order strategy**
- New `src/pipeline/reading_order/tategaki.rs` implements right-to-left columns, top-to-bottom within a column. Column grouping uses span x-position with monospace-derived tolerance (documented assumption).
- `src/pipeline/mod.rs` — `TextPipeline::process` intercepts at the top with a wmode-majority check (`vertical_count × 2 ≥ total → tategaki`). Pure-horizontal pages keep their configured strategy. Mixed-mode pages with horizontal majority also keep the configured strategy; per-span `wmode` is preserved on every span regardless of which sort ran.

**Per-span wmode tagging**
- `src/layout/text_block.rs` — `TextSpan.wmode: u8` records the active wmode at the paint site. Downstream consumers (export, search, layout-aware diff) can distinguish horizontal vs vertical content without re-walking the content stream.

**Font cache poisoning fix — descendant CIDFont metrics in identity hash (same class as `a327bcd`)**
- `src/document.rs` — `font_identity_hash_with_descendants` folds `/W`, `/W2`, `/DW`, `/DW2`, and `/CIDSystemInfo` from every resolved descendant CIDFont dict into the hash. Routed both call sites (`cached_font_identity_hash` and `load_fonts`) through the document-aware variant. Without this, two `PdfDocument`s with identical BaseFont/Subtype/Encoding/DW but different `/W2`/`/DW2` arrays silently shared parsed metrics — second document inherited the first's glyph displacements. Same poisoning vector as the recently-fixed ToUnicode-stream case.

**Cross-wmode span merge prevention**
- `src/extractors/text.rs` — `merge_adjacent_spans`'s `same_line` predicate gained a `current.wmode == span.wmode` clause. Previously a horizontal glyph and a vertical glyph whose baselines coincided (mid-BT Tf font switch) merged under the horizontal side's wmode, clobbering the vertical glyph's metadata. The same predicate is shared by all merge variants (cross-font glue, small-caps, decimal merge), so the gate covers them uniformly.

## Testing

- [x] All new and existing tests pass locally — **8109/8109**
- [x] I have run `cargo test --workspace --features rendering`
- [x] I have run `cargo clippy --workspace --all-targets -- -D warnings`
- [x] I have run `cargo fmt --all`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --lib` clean

**Test files added**

`tests/test_vertical_text_extraction.rs`
- `horizontal_identity_h_emits_x_advancing_spans` — control: horizontal text advances on x, per-span wmode=0.
- `vertical_identity_v_emits_y_advancing_spans` — Identity-V CIDFont with `/DW2`-default metrics. Per-glyph Y delta pinned to `font_size × |w1y|/1000` so a reduction to a relative-delta check that admitted accidental zero advance won't pass.

`tests/test_vertical_reading_order.rs`
- `vertical_spans_tag_wmode_1` — `TextSpan.wmode` reaches downstream consumers.
- `vertical_reading_order_is_right_to_left_top_to_bottom` — multi-column tategaki page sort order matches Japanese convention.

`tests/test_vertical_writing_mode_fixes.rs` (14 integration tests, the review-fix pass)
- `c1_page_renderer_multi_tj_vertical_paints_a_column` — page renderer advances along y between Tj operators in a single BT block.
- `c1_c2_extractor_multi_tj_vertical_advances_along_y` — extractor matches.
- `c2_c4_extractor_tj_array_vertical_advances_along_y` — TJ numeric offsets advance along y, not x.
- `c4_extract_chars_vertical_tj_offset_advances_along_y` — character-mode extraction path independent from span path.
- `c5_to_unicode_wmode_does_not_override_encoding_wmode` — ToUnicode `/WMode 1` does not flip a horizontal `/Identity-H` font.
- `c5_identity_v_with_silent_to_unicode_is_vertical` — counter: Identity-V with default-zero ToUnicode `/WMode` is vertical (encoding wins, ToUnicode silent).
- `i1_pipeline_routes_vertical_majority_through_tategaki_sort` — pipeline dispatch under wmode majority.
- `i1_to_markdown_and_to_html_route_vertical_through_tategaki` — both converter surfaces.
- `i2_tz_does_not_scale_vertical_advance` — `Tz 150` does not stretch vertical advances.
- `i3_w2_form_a_malformed_triple_does_not_desync` — malformed triple inside `/W2` Form A drops cleanly, subsequent CIDs stay aligned.
- `i4_w2_form_b_overflow_does_not_collapse` — u16 overflow at top of `/W2` Form B range breaks the loop rather than collapsing.
- `per_cid_w2_overrides_dw2_for_listed_cids` — end-to-end fixture with explicit `/W2 [1 [-500 250 600]]`; CID 1 uses per-CID metric, CID 2 falls back to `/DW2`.
- `mid_stream_tf_h_to_v_switches_span_wmode` — single BT/ET with H↔V `Tf` switches; both glyphs emit with the right per-span wmode (was previously a workaround using two BT blocks; workaround removed in this PR).
- `high_cid_w2_lookup_works` — `/W2` entry at CID 1280 (well past 256) flows through `extract_chars` correctly.

`tests/test_vertical_writing_mode_qa_probes.rs` (34 probes, the QA pass)
- 2 cache-poisoning regression tests (`probe_cache_poison_dw2_collision_across_documents`, `probe_cache_poison_w2_collision_across_documents`) that fail closed if `/W2`/`/DW2` are ever omitted from the font identity hash.
- 1 mid-BT Tf span-drop test (`probe36_mid_bt_tf_font_switch_drops_subsequent_spans`).
- 31 regression pins covering: Japanese packaging mixed page, Mongolian-shaped vertical (LTR), pure-horizontal regression, boundary-threshold dispatch (50/50, 1/3, 2/3), per-span wmode preservation through HTML, TJ large-negative-offset numerical safety, empty `<>Tj`, zero font size, multi-CID advances, `/W2` Form A+B interleave, missing horizontal metrics, u16::MAX, negative `v_x`/`v_y`, real-number `/W2`, indirect-reference `/W2`, multiple `/WMode` directives, `/WMode` inside a `begincidrange` block, `/WMode` after a `%` comment, Identity-V + absent `/W2` falls back to `/DW2`, missing `/CIDSystemInfo`, Mongolian-style LTR vertical (currently reverses — pinned gap), single column ordering, three-column RTL grouping, rotated-CTM interactions for both H and V text, `q/Q` preserves outer wmode, Form XObject containing vertical text, horizontal bulk perf bound, separation renderer vertical text smoke.

Plus unit tests in `src/fonts/cmap.rs`, `src/fonts/font_dict.rs`, `src/content/graphics_state.rs`, and `src/rendering/text_rasterizer.rs` pinning `/WMode` parser variants, `/W2` and `/DW2` array forms, axis-swap math under various Tm rotations, and `measure_text_bytes` vertical aggregation.

- [x] Python bindings updated (if needed) — no surface change; all behaviour is internal to the extractor / rendering pipeline.
- [x] Python tests pass

## Documentation

Inline rustdoc on `GraphicsState::advance_text_matrix`, `FontInfo.wmode`, `parse_w2_array`, `parse_dw2`, `parse_wmode_directive`, `TategakiStrategy`, and the new `cid_vertical_metrics` field documents the spec basis and the intentional design decisions (encoding wins over ToUnicode, Tz not applied to vertical, horizontal-glyph shapes painted at vertical positions per metric authority).

## Checklist

N/A — visible effects are spec-conforming text-position changes. Synthetic and QA-probe test fixtures pin per-glyph position expectations end-to-end.

## Additional Notes

**Targeted scope.** Vertical writing in modern prepress is concentrated in Japanese tategaki (sake, tea, soy sauce, miso, cosmetics, OTC pharma packaging — the majority of traditional-category Japanese consumer goods) and Traditional Chinese / premium Chinese categories (baijiu, premium tea, TCM, festival packaging). This PR brings extraction and rendering fidelity to those workflows. Mongolian and other vertical scripts are partially supported — see deferred items below.

**Honestly deferred (not regressions, real follow-up scope)**

1. **`vrt2` / `vert` GSUB shaping.** The rasterizer paints horizontal glyph shapes at vertical positions. CJK glyphs with rotated alternates in the font (full-width commas, brackets, dashes) won't display rotated. Correct per spec — PDF metrics are authoritative — but visually off for tategaki where Adobe / production RIPs apply the GSUB feature. Follow-up requires `rustybuzz` feature configuration for arbitrary subset fonts.

2. **Mongolian / Manchu (vertical LTR) reading order.** The tategaki strategy hard-codes right-to-left column iteration (CJK convention). Mongolian and Traditional Manchu read left-to-right vertically; their output currently reverses. Pinned explicitly by `probe29_mongolian_style_vertical_ltr_is_currently_misordered`. Addressable with a `Direction { RTL, LTR }` parameter on the strategy.

3. **HTML / Markdown converters do not emit `writing-mode` CSS.** Per-span `wmode` survives into the span model but converters render vertical-majority pages without a CSS hint. Downstream consumers must inspect `span.wmode` themselves. Pinned by `probe10_per_span_wmode_visible_via_extract_spans_but_not_in_html_output`.

4. **TJ negative-offset sign convention in vertical mode.** Currently a negative TJ numeric offset moves the cursor up in PDF user space (text-space +y). The §9.4.4 wording is ambiguous on direction-vs-magnitude for vertical; Adobe's behaviour may differ. Pinned by `probe11_large_negative_tj_offset_in_vertical_mode_no_overflow_or_nan` so the choice is visible to future calibration.

5. **`Ts` (text rise) axis semantics.** Routed perpendicular to the writing axis (along x in vertical) per a plausible reading of §9.3.5. The spec doesn't directly authorise this; if real fixtures show different expectations, refine.

6. **Real-world corpus.** Synthetic fixtures only — building a redistributable Japanese vertical-font sample isn't licensed and didn't fit the scope. The synthetic fixtures exercise the math and plumbing exhaustively; real-world fixture coverage is a follow-up when a permissively-licensed sample becomes available.

**Bug-class regression coverage.** This branch closes the same poisoning vector class as `a327bcd` (ToUnicode-stream font-cache collision): the `font_identity_hash_with_descendants` rewrite includes `/W`, `/W2`, `/DW`, `/DW2`, and `/CIDSystemInfo` so any future addition to the descendant CIDFont metric surface is naturally protected.


yfedoseev (commented):
Thanks for this, @RayVR — this is an exceptionally thorough piece of work. The single-decision-point design around `GraphicsState::advance_text_matrix` is exactly the right shape: one axis-swap site that multiplies the active Tm row, so CTM-rotated text stays correctly distinguished from font-level WMode 1. The `/W2`/`/DW2` parsing is spec-correct (DW2 `[v_y w1y]` order, sane defaults) and defensively written — malformed-triple drop preserves CID alignment, and overflow breaks instead of collapsing. Folding the descendant CIDFont metrics into the font identity hash to close the same poisoning class as `a327bcd` is a great catch. And the honesty in the deferred-items list (GSUB `vrt2`, Mongolian LTR, TJ sign convention, synthetic-only fixtures) is appreciated — all test-pinned.

A few things to confirm before merge, none blockers:

1. **Strategy override.** `is_vertical_majority` routes vertical-majority pages through `TategakiStrategy` even when a caller explicitly configured XYCut/StructureTree. I'm fine with this given it only triggers on real WMode-1 fonts, but I'd like the "tategaki always wins" behavior to be an explicit, documented contract — could you add a line to the public reading-order docs?
2. **Breaking field addition.** `FontInfo` (pub) gains three non-defaulted fields, which drives the test-file churn and would break external direct constructors. Worth a CHANGELOG note under breaking/API, and maybe `Default` to soften it.
3. **CHANGELOG.** The entry checkbox is unticked — this is a headline feature and should land with one.

**On CI:** the earlier Ruby 3.1 / windows-latest failure looked like a flaky native-ext build (an earlier run of this branch passed Ruby green, every other Ruby×OS leg passed, and this PR touches no binding code), so I've re-run it — it's currently still in progress. Everything else is green. **I'm leaning approve, but I'll hold the final sign-off until the Ruby re-run comes back green so we have a clean board.** Really nice work.

yfedoseev (approved):
Approving on the code. ✅ The single-decision-point axis-swap design, spec-correct `/W2`/`/DW2` parsing, and the font-identity-hash poisoning fix are all solid, and the test coverage is excellent.

Two follow-ups that are separate from the code approval (none block the review verdict):
- **CHANGELOG** entry for this headline feature (checkbox still unticked), plus a short breaking-API note for the new `FontInfo` fields.
- A one-line doc note that vertical-majority pages override the configured reading-order strategy.

On CI: the only red was the Ruby 3.1 / windows-latest flake, which I've re-run — if it (or any other CD job) needs attention we'll handle it separately. Really nice work, @RayVR.

yfedoseev (approved):
Re-approve. ✅ Thanks for the thorough follow-through, @RayVR.

Both follow-ups from the prior review are addressed:
- **CHANGELOG** — `## [Unreleased]` now documents the feature plus the breaking-API note for the new `FontInfo` fields (`wmode`, `cid_vertical_metrics`, `cid_default_vertical_metrics`), including the natural-construction defaults and the not-`#[non_exhaustive]` precedent.
- **Vertical-override doc** — `ReadingOrderConfig::strategy` now carries the vertical-majority override note, with the enum cross-reference.

The new commits since the last review hold up and are spec-correct:
- `/W2`–`/DW2` Form-A CID alignment fix (atomic triples + checked overflow break) matches ISO 32000-1:2008 §9.7.4.3; `/DW2` `[v_y w1y]` ordering + `[880 -1000]` default are right.
- Spans are no longer merged across writing-mode boundaries — gated at the single `same_line` chokepoint, so it only blocks genuine advance-axis mismatches.
- Descendant CIDFont `/W,/W2,/DW,/DW2` now fold into the font-identity hash, closing the Layer 5/6 cache cross-contamination.
- ToUnicode `/WMode` is correctly dropped (extraction-only, §9.10.2) with a logged disagreement rather than a silent override.

No production panics/unwraps in the new work. Two minor degenerate-input nits (non-transitive near-threshold column comparator; `%` comment-strip exactness in the CMap header parser) are safe to leave or fold into a follow-up — neither blocks. Nice work.

PR #646, opened 2026-06-04
feat: struct actualtext

RayVR:
## Description

Closes the §14.9.4 gap in struct-tree-scope `/ActualText` handling. Marked-content-scope `/ActualText` (inline `/Span << /ActualText (X) >> BDC ... EMC`) already worked. The other half — `/ActualText` on a `StructElem` dict in the structure tree, the form that authoring tools like InDesign emit for drop caps, ligature spans, and stylized text in tagged PDFs — was modeled (the `StructElem.actual_text` field was populated) but inconsistently applied: `extract_text` emitted the replacement **and** the raw descendant glyphs (duplication) and emitted at the wrong position; `to_markdown` / `to_html` / `to_plain_text` / `extract_structured` silently dropped it. In PDF/UA-1 / Section 508 / EU Web Accessibility Directive documents — gov.uk forms, IRS accessibility builds, anything authored through Acrobat / InDesign / NetCentric remediation tooling — this meant ligatures came out as PUA glyphs, drop caps as garbage, stylized small-caps runs as separate-glyph noise, and figures-rendered-as-text as nothing.

Hybrid eager index design. A single `ActualTextIndex` keyed by `(page, mcid)` is built once per document alongside the existing `structure_content_cache`. A per-page `actualtext_actions_for_page` helper computes a `HashMap<u32, ActualTextAction>` via consecutive-run dedup over structure-tree MCID order. All five extraction paths consume the same map; the content-stream layer (`extract_spans`) stays struct-tree-agnostic.

## Type of Change

- [x] Code refactoring
- [x] Tests

## Related Issues

Relates to the text-extraction-fidelity gap surfaced in earlier prepress / accessibility audit work.

## Changes Made

**Data model — spec-strict `(McidScope, mcid)` keyed index (§14.7.4.3 per-content-stream MCID scoping)**
- `src/structure/types.rs` — new `McidScope` enum (`Page(u32) | Form(ObjectRef) | Pattern(ObjectRef)`) discriminates the three MCID namespaces the spec defines: page content streams, Form XObject content streams, and Tiling Pattern content streams. Two Form XObjects on one page that each emit MCID 0 stay distinct because they push different `Form(form_ref)` scopes; same for Tiling Patterns.
- `src/structure/types.rs` — `ActualTextIndex` carries:
  - `mcid_to_actual_text: HashMap<(McidScope, u32), Arc<str>>` — innermost replacement per leaf MCID. Inner-wins precedence is automatic from the pre-order walker pushing/popping a current-scope stack.
  - `covered_mcids: HashSet<(McidScope, u32)>` — MCIDs whose raw glyph spans must be suppressed.
  - `suppress_only: HashSet<(McidScope, u32)>` — non-first-page entries of multi-page Page-scoped subtrees (structurally enforces emit-once-first-page). Form/Pattern scopes do not populate this set; each form/pattern is its own self-contained namespace.
- `StructChild::MarkedContentRef` gained `scope: McidScope`. The parser reads the MCR's `/Stm` entry and classifies it: `/Subtype /Form` → `Form(stm_ref)`; `/PatternType 1` → `Pattern(stm_ref)`; otherwise `Page(page)`. Bare-integer MCID children (no `/Stm` possible) are `Page(page)`. An unresolvable or unclassifiable `/Stm` logs and falls back to `Page(page)` — at worst the lookup misses and raw glyphs flow through unchanged, strictly safer than a collision.
- The earlier per-emission `ActualTextEmission` struct with `anchor_mcid` was removed entirely. Per-emission anchor matching was the root cause of three silent-content-loss bugs the first review found (nested-with-siblings drop, cross-page MCID collision, anchor-hidden-by-OCG drop) and the architectural fix collapses all three.

**Extractor — scope-tracking stack around Form XObject `Do`**
- `src/extractors/text.rs` — `TextExtractor` carries `mcid_scope_stack: Vec<McidScope>`. `set_page_index` (called by `extract_spans_impl`) initialises the bottom of the stack to `Page(page_index)`. `process_xobject` pushes `Form(xobject_ref)` before walking the form's content stream and pops it after (regardless of parse success). Every `TextSpan` created during the walk is stamped with `mcid_scope = current_mcid_scope()`, so spans extracted inside two distinct forms on one page carry distinct scope keys. Tiling Pattern content streams are not currently walked by the extractor (patterns rasterize independently); the spec-strict three-variant scope is still in the data model so future pattern-content walks can populate `Pattern(_)` without further refactoring.

**Public API surface bump — `TextSpan::mcid_scope`**
- `src/layout/text_block.rs` — `TextSpan` gained `pub mcid_scope: Option<McidScope>`. `Default` and the test/synthetic constructors initialise to `None`; real spans come pre-stamped from the extractor. `serde` skips the field when `None`, preserving existing JSON fixtures. The `OrderedTextSpan` wrapper transparently inherits the field via its embedded `span: TextSpan`.

**Builder — single pre-order walk**
- `src/structure/traversal.rs` — `build_actualtext_index(root, mcid_to_page)` walks the struct tree pre-order with a current-coverage stack. For each ActualText-bearing element it pre-scans the subtree for `first_page = min(Page-scoped descendant pages)` (Form/Pattern descendants do not participate in the cross-page emit-once rule — each form/pattern is a self-contained namespace), attributes each descendant MCR to either `mcid_to_actual_text` (Page-scoped on the bearing element's first page, or any Form/Pattern-scoped descendant) or `suppress_only` (Page-scoped descendants on non-first pages). Inner ActualText replaces outer for that descendant's `(scope, mcid)` key.

**Consumer — `(McidScope, mcid)`-driven emission via consecutive-run dedup**
- `src/document.rs` — `actualtext_actions_for_page` computes a per-page `HashMap<(McidScope, u32), ActualTextAction>` (EmitAndSuppress | Suppress). Action emission walks the structure-tree's per-page scoped-MCID order (`cached_mcid_order_for_page` returns `Vec<(McidScope, u32)>`) and emits a replacement string at most once per consecutive run that shares the same text. Inner-wins, cross-page-isolation, emit-once-first-page (for Page-scoped subtrees), and Form/Pattern namespace isolation all fall out of this construction. Span-based appliers (`apply_actualtext_to_spans`, `apply_actualtext_to_ordered_spans`) key their lookups by each span's `(mcid_scope, mcid)` pair.

**Decoupled Suspects gate (locked decision)**
- `src/document.rs` — new `struct_tree_marked()` returns the tree whenever it exists, independent of `/MarkInfo /Suspects`. Existing `struct_tree_trustworthy()` retains its Suspects gate for reading order. ActualText is *content replacement*, not reading-order ordering, so producer-supplied replacement text is honored even when Suspects=true.

**OCG-aware emission gate (locked decision)**
- The per-page action emission consults the OCG-visibility map alongside the covered-MCIDs set. An `ActualTextAction::Replace` fires when at least one of the action's covered MCIDs is visible under the active OCG state. If ALL covered MCIDs are filtered, the replacement is silently dropped (matches the locked "skip when all filtered" rule).

**MC-scope precedence (inner wins over enclosing struct ActualText)**
- `src/extractors/text.rs` — `MarkedContentContext` gained an `actualtext_emitted: bool` flag set on first show-text inside an MC-scope ActualText sequence. Subsequent show-text operators inside the same sequence skip re-emission. Fixes the pre-existing per-Tj duplication where `(X) Tj (Y) Tj` inside one ActualText scope produced `"mcmc"` instead of `"mc"`.
- A per-extraction `mc_actualtext_mcids` map records which MCIDs received MC-scope replacement; the assembler skips struct-scope replacement on those MCIDs (innermost wins, per §14.9.4).

**Stack-aware `current_mcid` for nested BDC**
- `src/extractors/text.rs` — `MarkedContentContext` gained `own_mcid: Option<u32>`; EMC pops and *restores* the outer scope's MCID rather than resetting to `None`. Fixes attribution of show-text operators that appear between an inner EMC and the next outer EMC.

**Converter integration — all five extraction surfaces**
- `extract_text`, `to_markdown`, `to_html`, `to_plain_text`, and `extract_structured` all route through the same applier. Cross-path consistency is asserted by a byte-equality test (`probe35_extract_structured_byte_equal_with_extract_text`).
- `OrderedTextSpan` gained `actualtext_replacement: Option<Arc<str>>` (used by the pipeline path); appliers also mutate `span.text` in place where the converter callers read it directly. Three previously-separate appliers (geometric, structure-order, pre-apply) collapsed into one centralized helper invoked from each path; the `assemble_text_from_spans` pre-applier was scoped down to the geometric/Suspects branch only, where it doesn't double-process.

**Per-page lifecycle for the MC-scope side channel**
- `src/document.rs` — `mc_actualtext_mcids` per-page entries are REPLACED on every extraction call (not appended monotonically as the pre-existing field accidentally was). Doc comment updated to reflect the contract.

**Collateral fixes uncovered by the work (also in this PR)**

- **BMC / BDC / EMC must flush the Tj span buffer.** `src/extractors/text.rs` — the buffered Tj path was accumulating across marked-content boundaries and capturing only the first Tj's MCID, silently gluing structurally-distinct elements together. Flushing at every BMC/BDC/EMC ensures each new Tj starts a buffer with the *current* MCID.
- **`merge_adjacent_spans` must consider MCID identity.** `src/extractors/text.rs` — adjacent spans with different MCIDs could fuse via cross-font-glue / small-caps / decimal merge, losing the boundary. Added `same_mcid` to the predicate shared by all merge variants.
- **`flush_tj_span_buffer` must propagate artifact_type.** `src/extractors/text.rs:7653` — was hardcoded `artifact_type: None`. Spans inside an `/Artifact` BDC bypassed the downstream filter, leaking pagination / header / footer content into body text. Pre-existing bug *amplified* by this branch because MC-scope `/ActualText` rides the same buffered path; a `/Pagination` artifact with `/ActualText "Page 1 of 10"` would inject the synthesized page number into body text. Fix mirrors the canonical pattern at `flush_tj_buffer` (line 6363).
- **`extract_text_structure_order_cached_with_spans` must filter `/Artifact` spans.** `src/document.rs:8929-8941` — the structure-order assembler had no artifact filter at all (the geometric branch dropped artifacts; the structure-order branch silently emitted them). Both reading-order paths must honor §14.8.2.2. Surfaced during the artifact_type fix verification — the source-side fix alone left `artifact_type=Some(Pagination(Other))` correctly flowing through but still landing in the output string of tagged-PDF extractions.

## Testing

- [x] All new and existing tests pass locally — **8088/8088**
- [x] I have run `cargo test --workspace --features rendering`
- [x] I have run `cargo clippy --workspace --all-targets -- -D warnings`
- [x] I have run `cargo fmt --all`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --lib` clean
- [x] `make c-header-check` clean (no FFI-reachable types changed)

**Test files added**

`tests/test_struct_actualtext.rs` — the initial 13 end-to-end fixtures (built with an embedded `PdfBuilder` DSL):
- `actualtext_simple_emits_replacement_no_raw`
- `actualtext_nested_inner_wins`
- `mc_scope_actualtext_wins_over_struct_scope`
- `actualtext_multi_mcid_subtree_emits_once`
- `actualtext_multi_page_emits_only_first_page`
- `actualtext_wins_over_alt`
- `alt_alone_does_not_replace_raw_glyph`
- `actualtext_on_figure_emits_replacement`
- `actualtext_unicode_round_trips` (CJK + Hebrew RTL + emoji)
- `actualtext_emits_when_suspects_true`
- `actualtext_ocg_excluded_all_suppresses_emission`
- `actualtext_ocg_visible_path_still_emits` (counter-test)
- `actualtext_cross_path_consistency` (extract_text / to_markdown / to_html / to_plain_text / extract_structured)

`tests/test_struct_actualtext_critical.rs` — the three review-CRITICAL bugs the architectural refactor closed:
- `critical1_nested_with_outer_sibling_emits_both` — outer Span ActualText "O" with inner Span ActualText "I" wrapping MCID 0 plus a sibling MCID 1: expects "I", "O" (was silent-dropping MCID 1).
- `critical2_cross_page_mcid_collision_does_not_swallow_unrelated_text` — page 0 H1 ActualText "Heading" over page-0 MCID 0; page 1 unrelated P over page-1 MCID 0 with raw "Body". The pre-architecture global-MCID-keyed model silently erased page-1's text.
- `critical3_partial_ocg_anchor_hidden_still_emits` — struct elem covers MCIDs 0 (anchor) and 1; MCID 0 is in excluded OCG; MCID 1 visible. The pre-architecture per-emission-anchor model dropped the emission; the mcid-driven model emits correctly.

`tests/test_struct_actualtext_majors.rs` — the three review-MAJOR bugs:
- `major1_nested_bdc_restores_outer_mcid` — `BDC[10] Tj BDC[11] Tj EMC Tj EMC`: the third Tj must attribute to MCID 10, not None.
- `major2_mc_scope_actualtext_emits_once_per_scope` — `/Span << /MCID n /ActualText (mc) >> BDC (X) Tj (Y) Tj EMC` emits "mc" once, not "mcmc".

`tests/test_struct_actualtext_gaps.rs` — non-Span/non-Figure StructElem types, long ActualText, malformed value forms, byte-equal cross-path consistency.

`tests/test_struct_actualtext_probes.rs` — 27 third-pass probes (QA layer):
- Architectural blast radius: consecutive-run dedup boundaries (3 consecutive same-replacement MCIDs collapse, mid-run inner override, non-consecutive runs don't merge), MCID iteration order is struct-tree pre-order.
- `suppress_only` semantics: first descendant on page 1 not page 0, multi-page skip middle page, document-wide ActualText.
- Stack-aware current_mcid: 3-level BDC nesting, BDC without `/MCID`, unmatched EMC doesn't panic.
- MC-scope once-per-scope: two Tj with Td in same scope, adjacent MC-scope sequences, nested MC-scope inner-wins, empty MC-scope ActualText pinned to suppress raw.
- MarkInfo/OCG: no MarkInfo + StructTreeRoot resolves; `/Marked false` + StructTreeRoot resolves; multi-layer OCG with all excluded drops; first-page-all-hidden pinned as current behavior.
- Value-format edge cases: null ActualText, indirect-reference string, line breaks verbatim.
- Cross-path consistency: extract_structured byte-equal with extract_text; clone-preserves-replacement; sequential extraction idempotent.
- Plus 2 reproducers (`probe34`, `probe34b`) for the artifact_type leak — un-ignored once the collateral fix landed.

Plus 9 unit tests in `structure::traversal::tests` pinning `build_actualtext_index` invariants (simple, nested inner-wins, nested outer shadowed, multi-page-first-page, multi-MCID, no-actualtext-empty, empty-actualtext-ignored, no-descendant-MCID-drops, figure).

- [x] Python bindings updated (if needed) — no surface change; behavior is internal to the extractor and converter pipelines.
- [x] Python tests pass

## Documentation

Inline rustdoc on `McidScope`, `ActualTextIndex`, `build_actualtext_index`, `actualtext_actions_for_page`, `struct_tree_marked`, `MarkedContentContext.own_mcid`, and the consecutive-run dedup helper documents the spec basis (§14.9.4 replacement semantics, §14.7.4.3 per-content-stream MCID scoping, §14.7.5.4.2 MCR `/Stm` resolution, §14.8.2.3.1 Suspects decoupling, §14.7.5 children models, §14.8.2.2 artifact filter) and intentional out-of-scope cases (`/ActualText` on OBJR children, Tiling Pattern content-stream walking — patterns rasterize independently in the extractor).

## Checklist

N/A — visible effects are spec-conforming text-content replacements. Synthetic test fixtures pin both the contract (cross-path byte equality) and the per-rule invariants end-to-end.

## Additional Notes

**Targeted impact.** Tagged PDFs are the primary beneficiaries: PDF/UA-1 / ISO 14289-1 documents, US Section 508 builds, EU Web Accessibility Directive output, and any artwork from authoring pipelines that emit struct-tree `/ActualText` for ligatures, drop caps, decorative figures, small-caps runs, or non-text glyphs that lack ToUnicode. Before this PR these extractions either dropped the producer-intended text or duplicated raw glyphs alongside it. Untagged PDFs are unaffected (the index is built lazily; `idx.is_empty() → None` short-circuits the hot path).

**Honestly deferred (not regressions, real follow-up scope)**

1. **`ExtractionContext` refactor for `mc_actualtext_mcids`.**  a per-extraction map living in a document-lifetime Mutex<HashMap> field. This PR fixed the *contract* (REPLACE on call, not append-monotonically) so behavior is correct, but the type signature still lies. A future refactor that introduces a per-call extraction context object would close the gap. Bounded scope; not done here to keep the diff focused.

2. **Cross-page fallback for all-first-page-hidden ActualText.** When a multi-page bearing element's first-page MCIDs are all OCG-filtered, the emission drops entirely (later pages are `suppress_only` and don't fall back). Matches the locked "skip when all filtered" rule literally but the cross-page case was not specified by that decision. Pinned by `probe19_first_page_hidden_via_ocg_drops_emission_entirely`. Worth a design discussion before changing.

3. **Untagged content at the tail of structure-order assembly.** Pre-existing behavior preserved: content emitted outside any struct element appears at the end of the structure-order stream. Real-world tagged PDFs tag all content, so this is moot in practice. Surfaced honestly; not introduced by this PR.

**Bug-class collateral coverage.** Two pre-existing bugs were uncovered and fixed alongside this work: artifact_type leak through the buffered Tj path (`text.rs:7653`) and missing artifact filter on the structure-order assembler (`document.rs:8929-8941`). Both are spec-conformance fixes (§14.8.2.2 artifact filtering, two reading-order paths must behave identically) that became important once MC-scope `/ActualText` rode the same content paths.


yfedoseev (commented):
Thanks for this — it's a genuinely impressive piece of work. The §14.9.4 struct-tree `/ActualText` gap is exactly the kind of accessibility-fidelity issue our tagged-PDF users hit, and the writeup (spec citations, locked-decision callouts, honest deferrals) made a +6.6k/41-file PR reviewable. The eager `(McidScope, mcid)` index with one builder and one applier across all five extraction surfaces is the right shape, and collapsing the earlier anchor-based content-loss bugs into the architecture is a clean win. Test coverage (143 fns, the critical/major reproducers, cross-path byte-equality) is excellent, and I re-ran the focused ActualText suites locally — all pass.

Leaving this as a comment rather than an approval because of the PR's blast radius; a few things I'd like to confirm before merge, none of them blockers:

1. **`mc_wins` keying.** The whole index goes to great lengths to key on `(McidScope, mcid)` to avoid cross-stream MCID collisions, but `mc_actualtext_mcids`/`mc_wins` is still bare `u32`-per-page (`HashMap<usize, HashSet<u32>>`). You flag this as deferred item #1, and I agree the real-world hit (MC-scope `/ActualText` inside a Form whose MCID numerically collides with a page MCID) is unlikely — just want to confirm we're consciously accepting it and tracking a follow-up rather than tightening the key now.
2. **`same_mcid` in the merge predicate.** This is the change with the widest reach beyond ActualText — it makes cross-font-glue / small-caps / decimal merges stricter for tagged content. `None == None` keeps untagged PDFs safe and the full suite is green, but could you confirm it's been run against a real tagged-PDF corpus? Adjacent same-word glyphs that legitimately carry different MCIDs are the case I'd want eyes on.
3. **New `/Artifact` filter on the structure-order assembler.** Correct per §14.8.2.2 and it aligns the two reading-order paths — just calling it out as a visible output change for tagged PDFs that over-tag artifacts.

Everything else (parser `/Stm` fallbacks, the Suspects decoupling, the `serde` skip on the new `TextSpan::mcid_scope`, no FFI surface change) looks solid.

**On CI:** the full suite is currently green on the latest commit, which is great — but I'm treating this as a comment, not a final approval, until points 1–2 are confirmed and CI stays green on any follow-up commits. Happy to give the final verdict then. Really nice work.

RayVR:
Addressed the three items.

**1. `mc_wins` keying — accepted as deferred.** Noted in the PR
description's deferred-items section and acknowledged in the conditional
approval. Real-world collision rate (a Form XObject MCID numerically
matching a page MCID) is low enough that the bare-`u32` keying is fine
short term; if a corpus surfaces a collision the fix is mechanical
(switch `mc_actualtext_mcids` / `mc_wins` from `HashMap<usize,
HashSet<u32>>` to `HashMap<usize, HashSet<(McidScope, u32)>>` exactly
the way the rest of the index already keys). Tracking as a follow-up;
no action on this PR.

**2. `same_mcid` merge predicate — pinned with two probes (d8d5c17).**

I went with a positive-probe approach rather than a corpus sweep
because (a) we don't have a large tagged-PDF corpus committed
(`tests/fixtures` has exactly one tagged PDF, `hello_structure.pdf`),
and (b) the underlying question is structural, not empirical.

**Spec defense.** ISO 32000-1:2008 §14.6 / §14.8 make the MCID the
structural unit. Two same-line fragments with different MCIDs belong
to different marked-content references and therefore either to
different structure elements or to different references to the same
element. The merge sites (same-font + tight gap, split-boundary,
cross-font glue, small-caps glue, decimal merge) all glue spans into
one — meaning one MCID survives and the other is silently dropped.
That is lossy by definition: structure-tree reading order, tree-scope
ActualText suppression, and table-cell membership all consume the
boundary the merge would erase. So the new conjunct enforces a
property that was already semantically required.

**Producer practice.** Real-world taggers (Acrobat, MS Office Save as
Tagged PDF, LaTeX's `tagpdf` package, Pandoc) emit MCIDs at logical-unit
boundaries — paragraph, sentence, list item, or larger. A single
visible word split across two MCIDs in their output means either (a)
producer error / over-granularity, in which case the structure is
already saying "these are distinct units" and we should respect it, or
(b) the producer wanted different per-half attribution (different
ActualText, different role) — in which case merging would *definitely*
be wrong.

**Empirical signal.** All 8088 lib tests pass locally, including the
full `tests/test_struct_actualtext*` suite (5 files, ~150 tests). No
regression in the tagged-PDF coverage that exists.

**New probes (d8d5c17).**
- `probe40_cross_mcid_fragments_do_not_merge` — synthesises a tagged
  PDF with "Hello" split as `(He) Tj` @ MCID 0 and `(llo) Tj` @ MCID 1
  in the same font, same baseline, zero gap. Asserts both fragments
  survive as separate spans with their MCIDs preserved, and the fused
  "Hello" does not appear.
- `probe41_same_mcid_fragments_merge_to_single_word` — regression
  sentry. Same fragments inside one `/Span /MCID 0 BDC … EMC`
  envelope must still merge to "Hello". Catches a future refactor
  that would tighten `same_mcid` into "always require different
  objects" and silently fragment every tagged word.

**Escape valve if regressions surface in the wild.** If a downstream
user reports word-fragmentation on a real tagged-PDF producer that
splits within words, the right fix is a structure-aware merge
predicate: merge if both MCIDs resolve to the same parent structure
element via the parent_tree. That's an O(1) lookup against the index
this PR already builds. I'd rather wait for a real reproducer than
write that carve-out speculatively. Happy to file a follow-up issue
if you want it tracked.

**3. `/Artifact` filter on structure-order assembler — acknowledged.**
Visible-output change for tagged PDFs that over-tag artifacts (the
`<Artifact>` ↔ regular `<P>` over-tagging pattern in some older
producers). The two reading-order paths (geometric and
structure-order) now agree per §14.8.2.2 — that was the whole point —
and the alignment is intentional. The PR description's "Collateral
fixes" section already names this; no further change needed.

Issue #650, opened 2026-06-06
docs: pdf.oxide.fyi Node.js "Quick Example" carries the same broken snippet as #648


yfedoseev:
Companion to #648 (fixed in the npm package for the next release).

**Problem**
The live docs site (pdf.oxide.fyi) "Quick Example" → **Node.js** tab shows the same broken quickstart the reporter hit in #648:
- `require("pdf-oxide")` — the package is ESM-only, so this throws `ReferenceError: require is not defined`.
- `new PdfDocument(path)` — the constructor takes an internal handle, not a path, so this throws `TypeError: invalid arguments`.

The npm `js/README.md` and a constructor guard were corrected in the release branch, but the **website snippet lives outside this repo** and is still broken for anyone copy-pasting from the docs site.

**Fix**
Update the docs-site Node.js snippet to the corrected form:
```js
import { PdfDocument } from "pdf-oxide";
const doc = PdfDocument.open("document.pdf");
console.log(doc.extractText(0));
```

**Action**
- [ ] Locate where the site sources its Node.js snippets and update them.
- [ ] Confirm the rendered Quick Example matches the fixed `js/README.md`.

Tracking only — no code change in this repo. Refs #648.

Issue #648, opened 2026-06-05
[Bug]:Could not run very primary sample on node.js


abeq:
### Bug Description

Hi,

First, exec ```npm install pdf-oxide```.

Then try below:

> https://pdf.oxide.fyi/docs/extraction/text##Quick Example
> Node.js tab

```
const { PdfDocument } = require("pdf-oxide");

const doc = new PdfDocument("report.pdf");
const text = doc.extractText(0);
console.log(text);
```

But it seems not run correctly
Please let me know how to fix or support for node.js, or not.

### Steps to Reproduce

1. Run on above but message that ```ReferenceError: require is not defined in ES module scope, you can use import instead``` has displayed.
  Modify that ```const { PdfDocument } = require("pdf-oxide");``` -> ```import { PdfDocument } from "pdf-oxide";```.

2. And rerun but message that ```TypeError: invalid arguments\n    at PdfDocumentImpl.extractText...``` has displayed.
  Modify page index as argument 0 to 1, but result is same.

### Expected Behavior

Display extracted text.

### Actual Behavior

Error has occurred:
```TypeError: invalid arguments
    at PdfDocumentImpl.extractText ...```

### PDF Information (if applicable)

- https://www.scs.stanford.edu/~zyedidia/docs/rust/rust_book.pdf

### Environment

- windows11 Professional with recent windows update
- node --version: v24.12.0
- **pdf_oxide version**: current, did ```npm install pdf-oxide``` at today.


### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

It very basically primary sample, but it does not run correctly.

### Checklist

- [x] I have searched existing issues to make sure this is not a duplicate
- [x] I have provided all required information above
- [x] I am using the latest version of pdf_oxide

PR #651, opened 2026-06-06
docs: fix Node.js PdfDocument open examples

lihouwenbin:
## Summary
- switch the Node.js README examples to ESM imports, matching the package type/module exports
- use PdfDocument.open(...) instead of constructing PdfDocument with a file path
- update the quick start, API tour, using, iteration, and async wrapper snippets consistently

Refs #648 and #650.

## Validation
- Checked js/README.md no longer contains require("pdf-oxide") or new PdfDocument(...)
- Checked examples against js/src/index.ts and js/index.d.ts, where PdfDocument.open(path) is the public file-open API
- git diff --check

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -16,9 +16,9 @@ npm install pdf-oxide
 ```
 
 ```javascript
-const { PdfDocument } = require("pdf-oxide");
+import { PdfDocument } from "pdf-oxide";
 
-const doc = new PdfDocument("paper.pdf");
+const doc = PdfDocument.open("paper.pdf");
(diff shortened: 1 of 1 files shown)


yfedoseev (approved):
Thank you, @lihouwenbin! 🙏 This is exactly the right fix — the package is ESM-only and `PdfDocument.open(path)` is the correct factory (the constructor takes an internal handle, not a path). Matches the corrected README usage and the docs-site follow-up (#650), and resolves the copy-paste-from-README failure reported in #648. Merging.

PR #652, opened 2026-06-06
feat(rendering): press-accurate CMYK→RGB via /OutputIntents ICC profile

RayVR:
## Description

Wires the document's `/OutputIntents` CMYK ICC profile through the composite render path so `/DeviceCMYK` paint, `/Separation` / `/DeviceN` colourants resolving to a `/DeviceCMYK` alternate, and `/ICCBased N=4` spaces lacking a usable embedded profile all convert to RGB through `qcms` against the press-target profile (ISO 32000-1:2008 §14.11.5 / §10) instead of the §10.3.5 additive-clamp fallback. The transform is built as `qcms::Transform::new_to(src = OutputIntent, dst = sRGB)`, so the OutputIntent profile is consulted via its AToB ("device-to-PCS") direction into the CIE PCS and the sRGB profile then takes PCS-to-device out — composite direction CMYK → CIE PCS → sRGB. Closes the press-vs-screen colour divergence on heavy-yellow / saturated-mid-tone branding artwork — a quarter-cyan paint that used to render through the `RGB = 1 - (C + K)` clamp as a near-screen-cyan (191, 255, 255) now renders against the OutputIntent CLUT as the muted press neutral the profile actually targets. When no `/OutputIntents` is declared the §10.3.5 path is preserved byte-for-byte.

The architecture threads `Option<&Arc<IccProfile>>`, `RenderingIntent`, and three `Option<&Object>` page-level default-colour-space overrides through the existing `ResolutionContext`. The `ColorResolver` consults them in spec-correct precedence at every CMYK → RGB site:

1. **Embedded `/ICCBased` profile on the colour space itself** wins when N=4 and the bytes compile through `qcms` (§8.6.5.5). This includes a precedence-inversion bug fix: pre-this-change, an embedded `/ICCBased N=4` with a parseable profile emitted `ResolvedColor::Cmyk` and was projected through the document `/OutputIntents` by the composite pipeline — inverting the spec. The four channels are now converted by the embedded profile directly.
2. **Page-level `/Default<Family>` overrides** win next when the active page or Form XObject's `/Resources /ColorSpace` declares them (§8.6.5.6). `/DefaultGray` / `/DefaultRGB` / `/DefaultCMYK` reinterpret the canonical `g`/`rg`/`k`/`K` operators as if they had named the override colour space. A malformed override (a string, integer, dictionary instead of a Name or Array) returns `None` so the resolver falls through to the next precedence layer instead of mis-rendering through `first_as_gray`.
3. **Document `/OutputIntents` CMYK profile** wins next for `/DeviceCMYK` paint and for `/Separation` / `/DeviceN` resolved alternates that land in `/DeviceCMYK`.
4. **§10.3.5 additive-clamp** is the final fallback.

The same precedence walk applies on the per-plate (separation) path: the new dual-payload `ResolvedColor::IccCmyk { rgba, cmyk }` variant carries both the pre-converted RGBA (for the composite backend) and the original CMYK quadruple (for the per-plate router), so embedded `/ICCBased N=4` paint preserves channel decomposition for the separations output without sacrificing the ICC-converted composite colour.

Rendering intent (`/RI`, §10.7.3) now actually dispatches into the colour stage. The operator was parsed but never consumed pre-this-change — every paint resolved under the initial `/RelativeColorimetric` regardless of any `/RI` operator in the content stream. The graphics-state intent now flows through `ResolutionContext::rendering_intent` into every `Transform::new_srgb_target` build. (See *Known limitations* below for the qcms downstream constraint.)

A per-page `IccTransformCache` lives on `PageRenderer` and amortises the 17⁴ CLUT precomputation `qcms::Transform::new_to` runs for CMYK input. The cache is keyed on `(profile.content_hash, intent)` and is n_components-agnostic at the storage level — it serves the N=1 Gray TRC arm, the N=3 RGB arm, and the N=4 CMYK arm identically. A page with 1 000 same-colour CMYK paints builds one transform, not one thousand; the test `output_intent_thousand_cmyk_paints_build_one_transform` (gated behind a new `test-support` feature) pins the hit rate.

`qcms` is Mozilla's pure-Rust colour-management module from Firefox. Already a dependency on the `icc` feature for image-pipeline ICC, this PR is the first consumer on the render path. ICC v2 and ICC v4 `DestOutputProfile` profiles are both supported through qcms 0.3.0's unconditional header-version check; the verification probe `qa_round3_iccbased_v4_profile_compiles_through_qcms_to_same_reference` builds the same LUT8 body under both versions and asserts byte-identical RGB.

## Type of Change

- [x] Bug fix (closes embedded `/ICCBased N=4` precedence inversion against document `/OutputIntents` per §8.6.5.5)
- [x] Bug fix (closes the `/RI` operator being parsed but never dispatched)
- [x] New feature (composite-path CMYK → RGB through `/OutputIntents` ICC, page-level `/Default<Family>` overrides, ICC v2/v4 support)
- [x] Performance improvement (per-page `IccTransformCache`)
- [x] Tests
- [x] Documentation (CHANGELOG, module docs, HONEST_GAP markers)

## Related Issues

- Closes the press-vs-screen colour divergence on `/DeviceCMYK` paint and `/Separation` / `/DeviceN` colourants resolving through `/DeviceCMYK` alternates against document `/OutputIntents`.
- Closes the §8.6.5.5 embedded-`/ICCBased N=4` precedence inversion (the embedded profile was being shadowed by the document `/OutputIntents` on the composite path).
- Closes the latent `/RI` dispatch gap (`Operator::SetRenderingIntent` emitted by the parser was being silently dropped by both renderers).
- Builds on top of the resolver-pipeline migration (`refactor/renderer-resolution-pipeline`), consuming the `ResolutionContext` borrows that migration parked behind tests for "add back when ready".
- Foundation for upcoming transparency-flattening and PDF/X-validation work.

### `src/rendering/resolution/context.rs` — three new context borrows + the cache type

- `output_intent_cmyk: Option<&'a Arc<IccProfile>>` (added back from the resolver-migration "park", with a real consumer this time).
- `rendering_intent: RenderingIntent`.
- `default_gray` / `default_rgb` / `default_cmyk: Option<&'a Object>` for the §8.6.5.6 overrides.
- `icc_transform_cache: Option<&'a IccTransformCache>` borrow.
- New `IccTransformCache` type: per-page, keyed on `(profile.content_hash, intent)`, n_components-agnostic, with `RefCell` interior mutability so callers continue to hold `&Context`.
- Builders `with_output_intent` / `with_rendering_intent` / `with_default_gray` / `with_default_rgb` / `with_default_cmyk` / `with_icc_transform_cache`.

### `src/rendering/resolution/color.rs` — the precedence rule lives here

- New `cmyk_to_rgb_via_intent(ctx, c, m, y, k)` `pub(crate)` helper. Routes through `ctx.output_intent_cmyk` + `ctx.rendering_intent` via the cache when present, falls back to §10.3.5 otherwise.
- `resolve_iccbased` now has three working ICC arms: N=4 (CMYK), N=3 (RGB), N=1 (Gray TRC). Each emits the correct variant (`IccCmyk` for N=4, `Rgba` for N=3 and N=1) and consults the per-page cache.
- New `resolve_device_default_override` fires at the top of the `LogicalColor::Device` arm. Routes bare `/DeviceGray` / `/DeviceRGB` / `/DeviceCMYK` paint through the override colour space via the same `resolve_spaced` machinery the explicit-`cs` paint operator uses. Malformed-shape overrides return `None` and fall through.
- The `ResolvedColor::Cmyk` arm of the composite projection (`run_pipeline_for_logical` near line 3209) calls `cmyk_to_rgb_via_intent` instead of bare `cmyk_to_rgb`. This is the §10.3.5-vs-OutputIntent fork point.
- `four_as_cmyk` (the per-tile composite helper in the same module, line ~264) takes the same OutputIntent-aware path.

### `src/rendering/page_renderer.rs` — wiring at the `ResolutionContext` construction site

- Calls `doc.output_intent_cmyk_profile()` once per page.
- Reads `/DefaultGray` / `/DefaultRGB` / `/DefaultCMYK` from the active resources (page-level + Form XObject scope).
- Threads `gs.rendering_intent` from the graphics-state stack.
- Hangs an `IccTransformCache` field on `PageRenderer`, cleared per page in `render_page_with_options`, and borrows it into every `ResolutionContext` via `with_icc_transform_cache`.
- New `/RI` operator arm dispatches into `gs.rendering_intent`.

### `src/rendering/separation_renderer.rs` — matching wiring at the per-plate call site

- Same `ResolutionContext` borrows so the two renderers see a single consistent colour policy for any given page.
- New `/RI` operator arm.
- The per-plate path does not host the cache (the renderer is a free function, not a stateful struct); the only Transform construction here is on `/ICCBased N=4` paint with a working embedded CMM, which is the design's cold path. Documented inline.

### `src/content/graphics_state.rs` — `rendering_intent` field

- New field defaulting to `/RelativeColorimetric` per §10.7.3.

### `src/rendering/resolution/resolved.rs` — `ResolvedColor::IccCmyk` variant

- Dual-payload variant carrying both pre-converted RGBA and the original CMYK quadruple. Source-breaking for downstream code that exhaustively matches on `ResolvedColor` (add the new arm to fix); the type is not `#[non_exhaustive]`.

### `Cargo.toml` — `test-support` feature

- New additive `test-support` feature exposes `PageRenderer::icc_transform_cache_build_count()`, backed by an instance-local `Cell<usize>` on the cache. Production builds compile the counter out entirely — zero overhead.

### `tests/test_render_output_intent.rs` — new integration suite

42 tests + 3 ignored. Synthetic minimal CMYK → Lab LUT8 ICC profile builders (constant-CLUT, parameterised by `target_l_byte`, byte-for-byte deterministic against qcms 0.3.0 references). PDF fixture builders for: bare `/DeviceCMYK` with and without `/OutputIntents`; `/Separation` / `/DeviceN` Type-4 alt-`/DeviceCMYK`; embedded `/ICCBased` N=1 / N=3 / N=4 with conflicting document `/OutputIntents`; page-level `/Default<Family>` overrides; v4 ICC profile headers; the green-mark branding regression sentry. Edge-case probes: `/N=3` OutputIntent filtered at the reader; missing `/DestOutputProfile`; mixed `[RGB, CMYK]` array picks CMYK; malformed garbage stream falls through.

### `tests/fixtures/icc/README.md` — fixture catalogue

- Documents the synthetic LUT8 builder, the v4 header switch, the constant-CLUT design choice (so qcms references are independent of intent and trivial to derive), and the HONEST_GAP rationale for not committing a real-corpus press fixture.

### `CHANGELOG.md` — `[Unreleased]` entry

- Press-accurate OutputIntent feature list, the source-breaking `IccCmyk` variant note, the precedence-inversion fix, and the HONEST_GAP markers.

### Gate sweep

| Gate | Status |
|---|---|
| `cargo fmt --all --check` | PASS |
| `cargo clippy --features rendering --all-targets --no-deps -- -D warnings` | PASS |
| `cargo clippy --features rendering,icc,test-support --all-targets --no-deps -- -D warnings` | PASS |
| `cargo check --lib --no-default-features` | PASS |
| `cargo test --features rendering --lib` | PASS |
| `cargo test --features rendering,icc --tests` | PASS |
| `cargo test --features rendering,icc,test-support --tests` | PASS |
| `cargo doc --no-deps --features rendering --document-private-items` | no new warnings |
| Forbidden-content scan (`fa7c801..HEAD`) | empty |
| Wave QA suites (`wave1`/`wave2`/`wave3`/`wave4`/`wave5`/`resolution`) | PASS |
| `test_separation_rendering` | PASS |
| `Cargo.toml` dependency-table audit (`git diff fa7c801..HEAD -- Cargo.toml`) | shows only `test-support = []` feature addition; no dependency edits |

### New tests

- **42 passing** in `tests/test_render_output_intent.rs` (3 ignored, all `HONEST_GAP_*`-marked: qcms intent dispatch, qcms BPC, real-corpus branding fixture).
- **3 cache-counter probes** behind `--features rendering,icc,test-support`: `output_intent_thousand_cmyk_paints_build_one_transform`, `qa_round3_cache_keys_include_rendering_intent`, `qa_round4_thousand_rgb_paints_through_default_rgb_build_one_transform`.

### CI lane recommendation

Add a `--features rendering,icc,test-support` lane so the cache-counter probes run. The existing `--features rendering,icc` lane silently filters them out via `#[cfg(feature = "test-support")]`.

- [x] Python bindings updated (if needed) — N/A, no PyO3 surface changes; the work is pure-Rust composite renderer plumbing
- [x] Python tests pass — verified locally; no Python-visible behavior changes
- [x] Python code formatted with `ruff format` — N/A, no Python diffs
- [x] Python code linted with `ruff check` — N/A, no Python diffs

## Checklist

- [x] My changes generate no new warnings (`cargo doc` warning count = baseline = 52)
- [x] The PR title follows conventional commits format (`feat(rendering):`)

## Documentation

- CHANGELOG `[Unreleased]` section covers the feature, the source-breaking `IccCmyk` variant, and the known limitations.
- Module docstring at `src/rendering/resolution/color.rs` documents the precedence rule.
- `src/rendering/resolution/context.rs` field-level docstrings cite the §-numbers for each borrow.
- `tests/fixtures/icc/README.md` catalogues the synthetic fixture design.
- HONEST_GAP markers (compile-time `&str` constants referenced from `#[ignore]`-marked tests) document every deferred capability so a future engineer auditing the suite sees exactly what's pending.

### How to review the 38 commits

The work landed in five rounds of design + QA + fix; reviewing by round is the easiest navigation:

- **Round 1** (`5e287ef`..`5e5d1f8`, 9 commits): foundation. `ResolutionContext` borrows, the `cmyk_to_rgb_via_intent` helper, the `Transform::convert_cmyk_pixel` route, the §10.3.5 byte-exact pin, the byte-exact qcms reference (replaces a pre-measurement `±10` hand-wave).
- **Round 2** (`fa1b947`..`fe4a3c1`, 6 commits): embedded-`/ICCBased N=4` precedence inversion fix + the `IccCmyk` dual-payload variant + N=4 fall-through probes.
- **Round 3** (`9e5f5c0`..`36122e4`, 6 commits): per-page transform cache, `/RI` dispatch in both renderers, ICC v4 verification, qcms intent HONEST_GAP.
- **Round 4** (`8bcce26`..`621be0a`, 12 commits): §8.6.5.6 `/Default<Family>` overrides with separate N=1/N=3/N=4 ICC arms in `resolve_iccbased`, the BPC HONEST_GAP, the green-mark branding sentry, the malformed-override fall-through.
- **Round 5** (`136295c`..HEAD, 4 commits): final edge-case probes (`/N=3` reader filter, missing `/DestOutputProfile`, multi-entry array, malformed stream), the `CmykTransformCache → IccTransformCache` rename, CHANGELOG, this PR description.

Within each round the pattern is `test (failing) → feat → style sweep`.

### Performance

The `IccTransformCache` lives on `PageRenderer` so the per-operator `ResolutionContext` (rebuilt fresh per paint dispatch) can borrow it without rebuilding. The 1 000-paint test verifies the amortisation: same-colour CMYK paints build a single transform across all 1 000 calls. The cache key is the profile's content hash plus the rendering intent — distinct profiles for Form XObjects coexist on the same cache without collision, and two interleaved `/RI` intents on the same page produce two distinct entries (pinned by `qa_round3_cache_keys_include_rendering_intent`).

### Honest design choices

- **`ResolvedColor::IccCmyk` is dual-payload, not Rgba-only.** A monolithic `Rgba` would zero out every separation plate (the per-plate router needs the original CMYK quadruple); per-plate-only would lose the ICC-converted composite RGB. The variant carries both. Source-breaking for downstream pattern matches, but the type was never `#[non_exhaustive]` and a single new arm is the clean fix.
- **The cache lives on `PageRenderer`, not on `ResolutionContext`.** The operator dispatcher builds a fresh `ResolutionContext` per paint operator (the resolver migration's design), so a context-local cache would amortise nothing. The `PageRenderer` lifetime spans the page, which is the amortisation scope that matters.
- **The cache key uses `profile.content_hash`, not `Arc` identity.** Multiple `Arc<IccProfile>` instances built from the same bytes (e.g. when a Form XObject re-parses the same stream) hash to the same key and share the compiled transform. Two profiles with byte-identical contents would collide on the cache key, but the resulting transform is identical so the collision is harmless. Distinct profiles never collide because the hash incorporates the colour-space signature (`'CMYK'` / `'RGB '` / `'GRAY'`) and the byte payload.

### Deferred items (HONEST_GAP markers)

- **qcms 0.3.0 silently ignores the CMYK rendering intent.** pdf_oxide's wiring is correct end-to-end — the intent reaches qcms — but qcms 0.3.0 declares the intent as `_intent` for CLUT-based CMYK conversion (`transform.rs:1283-1289`) and dispatches the same CLUT for every PDF intent. A qcms upgrade that honours the parameter, or a CMM swap, will surface intent-sensitive behaviour without further code changes. The test `qa_round3_qcms_030_treats_cmyk_intent_as_informational` is the upgrade gate. The intent-sensitive integration probe stays `#[ignore]`-marked with `HONEST_GAP_INTENT_SENSITIVE_FIXTURE`.
- **qcms 0.3.0 has no Black-Point Compensation** (`lib.rs:29-36` documents the choice as intentional). `qa_round4_bpc_paper_white_preservation_under_relative_colorimetric` is `#[ignore]`-marked with `HONEST_GAP_QCMS_030_NO_BPC`. A non-constant K-axis ramp LUT is built so a behaviour change surfaces rather than collapsing on a flat CLUT; under qcms 0.3.0 all four PDF intents produce byte-identical bytes for the deep-shadow input.
- **No real-corpus branding-logo regression fixture.** The synthetic green-mark probe (`qa_round4_branding_green_mark_routes_through_output_intent`) pins the press-target direction-of-shift through saturation collapse — saturated additive-clamp vs muted OutputIntent — but uses a constant-CLUT synthetic profile. A vendor-issued press profile plus a CIEDE2000 ΔE assertion against a commercial-viewer baseline would tighten the bound. `HONEST_GAP_NO_REAL_BRANDING_FIXTURE` documents the pending work.
- **RGB and Gray `/OutputIntents`** are out of scope for this PR (only the `/N=4` CMYK arm is read by `output_intent_cmyk_profile()`). The renderer writes sRGB PNG; the viewer's display profile is its responsibility.


yfedoseev (approved):
Approve with nits. ✅ Excellent work, @RayVR — the colour science is right and the failure handling is conservative.

I checked the conversion path against the spec and it holds up:
- **Direction & precedence are spec-correct.** Device CMYK → sRGB via the OutputIntent's `/DestOutputProfile` uses the profile's AToB ("to CIE") direction for screen preview, which is exactly what §14.11.5 sanctions. Embedded `/ICCBased N=4` (with a working CMM) is consumed directly as the conversion source rather than re-projected (§8.6.5.5), `/DefaultCMYK` overrides before device-CMYK reaches the OutputIntent (§8.6.5.6), and §10.3.5 additive-clamp is the terminal fallback. Picking the first N=4 output intent is a sound policy (PDF has no output-intent selector, §14.11.5 NOTE 1).
- **No render-path panics.** Every `.unwrap()/expect()` is in `#[cfg(test)]`; malformed / header-only / CMM-less / garbage profiles all degrade to the clamp fallback, and you have tests for each. The `test-support` cache-counter feature is a clean, genuinely zero-overhead test seam.
- **Tests are high-signal** — byte-exact pixel assertions against derived qcms references, the right edge cases (mixed `[RGB,CMYK]` array picks CMYK, N=3 filtered, missing `/DestOutputProfile`), honest `#[ignore]` gap markers, and documented fixture provenance.

**One trivial fix before merge:**
- The CHANGELOG/PR wording says the transform uses "from CIE (BToA)" — that's inverted. The code correctly builds a `new_to(src=OutputIntent, dst=sRGB)` transform, i.e. AToB/"to CIE". Please correct the wording so it doesn't mislead future auditors; the implementation itself is right.

**Non-blocking follow-ups (fine as a separate issue/PR):**
- **DRY:** the N=4/N=3/N=1 arms in `resolve_iccbased` (`src/rendering/resolution/color.rs`) are near-identical `decode → parse → get_or_build → has_cmm → convert` triplets — worth collapsing into one private helper so a future cache/clamp fix lands in one place.
- **Perf:** embedded `/ICCBased` spaces decode + `IccProfile::parse` on every paint operator (only the compiled `Transform` is cached, not the parse). Hoisting the parsed profile to the per-page context would close the embedded-space hot path; the 1000-paint test only exercises the device-CMYK→OutputIntent path, which has no per-paint parse.
- **Cache key:** `content_hash` uses `DefaultHasher` over the profile bytes; keying on `(content_hash, n_components, intent)` (or storing byte-len) hardens against a u64 collision returning the wrong transform.

None of the follow-ups block — happy to approve once the CHANGELOG wording is corrected. Thanks again, this is a meaningful press-quality improvement.

RayVR:
Thanks for the careful review!

**Wording fix landed in `f6cdd2e`.** Both `CHANGELOG.md` and the PR description now spell out the transform direction explicitly: `qcms::Transform::new_to(src = OutputIntent, dst = sRGB)` consults the OutputIntent profile's **AToB ("device-to-PCS") direction** into the CIE PCS, then the sRGB profile's PCS-to-device direction out — composite direction CMYK → CIE PCS → sRGB. The implementation direction itself was always right; the prior wording could be misread as a BToA inversion. The added clarifier closes that.

The literal strings `from CIE` / `BToA` / `B2A` don't actually appear anywhere in the diff — your quoted phrasing was a paraphrase of the implicit reading. The fix adds explicit direction language so no future auditor can misinterpret the existing description.

**The three non-blocking follow-ups are tracked** and will land in a separate PR after #652 merges:

- DRY the N=4/N=3/N=1 arms in `resolve_iccbased` — they're near-identical `decode → parse → get_or_build → has_cmm → convert` triplets that diverge only in emission shape. Consolidating means future cache/clamp fixes land in one place.
- Hoist embedded `/ICCBased` parse to the per-page context. The Transform cache covers the document `/OutputIntents` hot path, but embedded `/ICCBased` colour spaces still pay per-paint `IccProfile::parse` cost. New perf probe will pin one-parse-per-page invariant after the fix.
- Harden `IccTransformCache` key from `(content_hash, intent)` to `(content_hash, n_components, intent)`. Belt-and-braces against a theoretical u64 collision returning a wrong-shape Transform.

Ready to merge whenever you are. Thanks again — the spec-precedence walk you ran in your review note will be useful to future auditors reading the diff.

PR #674, opened 2026-06-08
feat(rendering): §11 transparency surface + IccBackend trait with optional lcms2 backend

RayVR:
## Description

Implements the §11 transparency surface end-to-end for press-grade separation output, with byte-exact spec correctness across the composite-then-separate path. A CMYK paint with non-trivial transparency, an overprint-active page, a Form XObject `/SMask`, a non-separable blend mode, a Pattern with a `/Separation` underlying, a DeviceN `/Process` paint with an embedded ICC profile differing from the OutputIntent — all render byte-exact against spec-derived references instead of the §10.3.5 additive-clamp approximation.

Architecture: composite all transparency math in the source colour space (CMYK + per-spot-ink sidecar lanes), then decompose to per-plate output for `render_separations`. Spec-correct per ISO 32000-1 §11.3 + §11.4 + §11.6.7 + §11.7.3 + §11.7.4; matches real prepress / packaging RIP workflows.

## Type of Change

- [x] New feature (non-breaking — extends rendering correctness; existing call sites unchanged)
- [x] Performance improvement (composite-path detection gate keeps detection-OFF pages on the existing per-plate walker)
- [ ] Bug fix
- [ ] Breaking change

## Related Issues

None tracked upstream. Driven by an internal roadmap for press-grade prepress rendering (composite-then-separate semantics for separation output, opt-in ICC retargeting).

### Composite-then-separate architecture

- New `CmykSidecar` struct (CMYK plane + per-spot-ink lanes). Spot ink set discovered via a pre-pass walking `/Resources/ColorSpace`. Sized once per page; spot lanes initialise to subtractive 0.0 per §11.7.3.
- `page_renderer.rs` composite path populates the sidecar through every paint operator; `separation_renderer.rs` routes detection-ON pages through the composite path then decomposes per-plate; detection-OFF (pure-overprint, no transparency) pages stay on the existing per-plate walker for byte-identical output.
- `page_declares_transparency` detection gate (narrower than `page_declares_transparency_or_overprint`) honours §11.7.4 OPM correctness on pure-overprint pages by keeping them on the per-plate walker.

### §11.3.5 blend modes

- Non-separable blend modes (`/Hue`, `/Saturation`, `/Color`, `/Luminosity`) implemented per §11.3.5.3 in HSL/HSY space via a new `src/rendering/blend_nonsep.rs` module. BT.601 luma weights, `SetLum`/`SetSat` per spec.
- The pre-existing `_ =>` arm in `page_renderer` that silently degraded non-separable modes to `SourceOver` is closed.

### §11.7.4.2 spot-lane Normal substitution

- `BlendModeClass::{SeparableWhitePreserving, SeparableNonWhitePreserving, NonSeparable}` classifier dispatches per §11.7.4.2: non-separable + non-white-preserving blend modes apply to process channels only; spot lanes substitute `/Normal` per the dispositive spec rule.
- §11.7.4.2 dispatch wired at every paint surface (path fills, strokes, combo FillStroke, text-showing operators, Image Do, ImageMask Do, shading sh).

### SMask handling

- Form XObject `/SMask /S /Alpha`, `/S /Luminosity`, `/BC` backdrop colour, `/TR` transfer function — previously documented as "intentionally ignored" — fully implemented with proper `SoftMaskForm` / `SoftMaskSubtype` plumbing through the graphics state. Image-attached SMask continues to work.
- `MAX_SMASK_DEPTH=32` cap on cyclic `/G` recursion (documented policy choice).
- SMask attenuates spot lanes per §11.3.3's single shape/opacity rule via per-pixel mask alpha applied uniformly across all lanes.
- `/BC` backdrop colour handled for n=1 (Gray), n=3 (RGB), n=4 (CMYK), AND n≥5 (DeviceN/NChannel) — the n≥5 path resolves the Form XObject's `/Group /CS` and evaluates the tint transform with the backdrop tints before projection.
- `/TR` transfer function supports all four PDF function types: Type 0 (sampled), Type 2 (exponential), Type 3 (stitching), Type 4 (PostScript calculator). The Type 4 path reuses the existing PostScript calculator at `src/functions/mod.rs` that already serves Separation / DeviceN tint transforms.

### Knockout groups `/K`

- §11.4.6.2 knockout composition: each element composes against the group's initial backdrop, not the previous element's result. Sidecar lanes (CMYK + spot) snapshot at group entry, restore before each replay, merge per-byte diffs into accumulators, install at group exit.
- Form XObject Do inside a `/K` group correctly preserves the inner Form's authoritative paint writes (the post-Do mirror is suppressed for Form XObjects since the Form's content stream already painted with its own gs).

### `/RI` rendering intent

- Previously silently dropped; now flows through to the colour stage and into `try_retarget_cmyk_via_embedded_profile` for ICC retargeting.

### §11.7.4.3 CompatibleOverprint (per-channel)

- Replaces the previous `(src + dst).min(1.0)` additive merge with spec-correct §11.7.4.3 Table 149 per-channel REPLACE / PRESERVE / SKIP logic.
- Three source-CS classes dispatched: `DeviceCmykDirect` (only class where §11.7.4.5 OPM=1 zero-source-preserve fires per spec), `OtherProcess` (DeviceGray/RGB/ICCBased/DeviceN /Process), `SeparationOrDeviceN` (spot inks preserve backdrop on process channels; spot mirror handles spot lanes).
- Stale `gs.fill_color_cmyk` cleared on `SetFillRgb` / `SetFillGray` / `SetStrokeRgb` / `SetStrokeGray` to prevent leakage across paint operations.
- Cross-path byte-identity verified: pure-overprint DeviceCMYK page renders byte-identical via per-plate walker (detection-OFF) and via composite-then-decompose (detection-ON synthetic trigger).

### RGB-source paint into CMYK sidecar (§11.3.4 single blend space)

- Adds an RGB→CMYK sidecar mirror so RGB-source paints contribute to the CMYK sidecar plane per §11.3.4's "one blend space, paint-time conversion" rule.
- Backend-dispatched: under `icc-lcms2`, conversion goes sRGB → Lab → OutputIntent CMYK with full ICC accuracy; under default `icc-qcms` or no-CMM builds, falls back to §10.3.5 additive-clamp inverse `(C, M, Y) = (1-R, 1-G, 1-B)`, `K = 0` (the K-loss is documented in-source).
- Wired at the Stroke and Fill arms; CMYK transparent paint over an RGB backdrop now composites against the converted backdrop instead of paper-white.

### DeviceN `/Process` source CMYK

- `SetFillColorN` / `SetStrokeColorN` for DeviceN sources with `/Process` attribution populate `gs.fill_color_cmyk` and `gs.fill_color_rgb` from the actual `scn` tint values via the `/Process /ColorSpace` tint transform — replacing the previous tint-blind path that produced constant `(1, 1, 1, 0)` source CMYK regardless of tints.
- Handles `/Process /ColorSpace` = `/DeviceCMYK` (tints as natural-form), `/DeviceRGB` / `/DeviceGray` (§10.3.5 additive-clamp inverse), `/ICCBased` (natural-form fallback under default features; full retargeting under `icc-lcms2`).
- Mixed DeviceN (process prefix + spot tail) routes process tints through the process classifier AND the spot tail through the existing spot mirror.
- §8.6.8 initial colour for DeviceN `/Process`: `cs /CS_N` populates a CMYK identity built from `extract_process_paint_cmyk(all-1.0_tints)` instead of leaving `fill_color_cmyk = None`.

### Pattern colour spaces

- Pattern colour spaces with `/Separation` or `/DeviceN` underlying CSes recurse correctly through ink discovery (sidecar + separation_renderer parity). Indirect-ref underlying CSes deref via `doc.resolve_object` (previously fell through to `ResolvedSpace::Unknown`).

### Real coverage rasterisation for text-show / Image Do / shading sh

- Replaces the previous binary "any RGB change" coverage approximation at text-showing operators (`Tj`, `TJ`, `'`, `"`), Image / ImageMask Do, and shading sh paint sites.
- New helpers re-run text / image / shading paint with a `coverage_only_gs` (forces opaque-black, no SMask, no transparency, Normal BM) into a scratch RGBA Pixmap; alpha channel extracted as coverage. AA-edge byte-exact correctness; identical-RGB-collision pixels no longer silently skipped.
- Invisible text (`3 Tr` per §9.3.6) correctly produces no spot-lane write (the coverage scratch honours the user's `render_mode` instead of overriding it to 0).
- ImageMask `/Decode` array honoured per §8.9.6.2 (default `[0 1]` = bit-0 paints).

### ICC backend trait + lcms2 backend

- `IccBackend` trait extracts the CMM surface pdf_oxide actually needs: profile parsing, sRGB transform building, CMYK→sRGB conversion (single + buffer), CMYK→CMYK retarget building, retarget application, sRGB→CMYK transform building (added for the RGB→CMYK mirror path), intent + flag dispatch.
- `QcmsBackend` (the existing pure-Rust default) implements every method qcms 0.3.0 supports; `build_cmyk_retarget` and `build_srgb_to_cmyk` return `None` (qcms 0.3.0 has no CMYK output pipeline).
- `Lcms2Backend` (new, opt-in behind `icc-lcms2` feature, FFI wrapper around Little CMS 2.x) implements the full surface including CMYK→CMYK retargeting with BPC + rendering intent dispatch. `DisallowCache` for `Sync` so `Arc<Transform>` can cross rayon workers under the `parallel` feature.
- Compile-time backend selection via `ActiveIccBackend` type alias; lcms2 wins when both features are enabled.
- `try_retarget_cmyk_via_embedded_profile` short-circuits when (a) the active backend can't retarget CMYK, (b) the document has no OutputIntent CMYK profile, (c) either profile fails to parse, (d) the embedded profile's `content_hash` matches the OutputIntent's (identity skip). Otherwise retargets `src.AToB → Lab → dst.BToA` with BPC for the press-default intent.
- Rendering intent threaded from `gs.rendering_intent` (set by the `/RI` operator) through `extract_process_paint_cmyk` to the lcms2 transform builder — `/Perceptual ri` followed by a /DeviceN /Process /ICCBased paint retargets with perceptual intent, `/Saturation ri` with saturation, etc. §8.6.5.8 default (RelativeColorimetric, the press standard) applies when no `/RI` is in effect.

### Byte-exact references throughout

Every probe pins byte-exact CMYK / spot-lane / per-plate values computed by hand from §-formulas or via independent CMM verification (lcms2 standalone for retargeting probes). No `±X` tolerances. Where the spec is genuinely silent, the chosen reading is pinned as a documented marker with both the chosen byte value and the alternate-reading rationale captured in-source.

### Mechanism-pinning probes (sensitivity-verified)

Where a fix addresses a specific underlying mechanism (e.g., Form Do post-mirror suppression for nested /K, identical-RGB collision in coverage extraction, ICC intent threading, Type 3 stitching dispatcher), probes are sensitivity-verified: temporarily reverting the fix in a git stash → probe fails with the predicted incorrect byte; restoring → probe passes. Documented in commit messages.

### Test surface

14+ integration test files covering:

- Spot ink discovery (Separation / DeviceN / `/All` / `/None` reserved-name exclusion / hex-escaped names / NChannel / `/Process` attribution / mismatched names)
- Sidecar storage + detection gating (per-plane allocation invariants, detection-OFF byte-identity)
- §11.7.4.2 dispatch (every BM × every paint surface × spot vs process lane)
- SMask alpha + luminosity + `/BC` n=1/3/4 + `/BC` n≥5 (DeviceN with tint-transform projection) + `/TR` Type 0 + Type 2 + Type 3 stitching + Type 4 PostScript + cyclic `/G` + nested `/K` + multi-glyph TJ + invisible text
- Knockout /K cumulative-replay across CMYK + spot lanes; nested /K through Form XObject Do
- §11.7.4.3 CompatibleOverprint (DeviceCMYK direct, DeviceGray, DeviceRGB, /Separation, /DeviceN, /DeviceN /Process /CMYK + /RGB + /Gray + /ICCBased + mixed-with-spot-tail)
- §11.7.4.4 OPM=0 and §11.7.4.5 OPM=1 across all source classes (verified OPM=1 zero-source-preserve binds DeviceCmykDirect only)
- §11.3.4 RGB-source paint into CMYK sidecar; CMYK transparent paint over RGB backdrop composites against converted backdrop, not paper-white. Both `icc-lcms2` ICC path AND the `icc-qcms` / no-CMM §10.3.5 fallback path probed
- Pattern with /Separation / /DeviceN underlying via inline arrays AND indirect refs
- Real coverage at text-show / Image Do / shading sh; AA-edge byte-exact; identical-RGB collision; rotated CTM glyph footprint; multi-glyph TJ with negative kern; empty Tj no-panic; invisible text no-lane-write
- ICC retargeting under all 4 feature combos: default qcms (preserves natural-form), `icc-lcms2` (real retargeting verified against independent lcms2 reference values), both features enabled (lcms2 wins), `--no-default-features` (no-CMM §10.3.5 fallback)
- Rendering intent dispatch (Perceptual, Saturation, AbsoluteColorimetric, RelativeColorimetric default, qcms-treats-intent-as-informational)
- Cross-path byte-identity (walker vs composite for pure-OP pages)

### Architectural deviation honestly surfaced

The CMYK sidecar on `PageRenderer` is used instead of routing every paint through `SeparationBackend`'s plate-emit shape, because the latter would require emitting `ResolvedPaintCmd` from every paint surface (currently only path fills / strokes). The sidecar achieves equivalent press-accurate output (composite-first transparency, post-composite plate decomposition) without the architectural rework. Probes verify the two architectures produce equivalent output on every probed edge case.

### Pre-merge verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy --all-targets --features "rendering icc test-support" -- -D warnings` — clean
- `cargo clippy --all-targets --features "rendering icc-lcms2 test-support" -- -D warnings` — clean
- `cargo check --lib --no-default-features` — clean
- `cargo test` (default features) — full suite green
- `cargo test --features "rendering icc-lcms2 test-support"` — green
- `cargo test --features "rendering icc icc-lcms2 test-support"` — green (both features)
- `cargo test --no-default-features --features "rendering test-support"` — green (no ICC)
- `cargo doc --no-deps --features "rendering icc-lcms2 test-support"` — clean

## Python Bindings

No binding changes. The existing `render_page` / `render_separations` Python API surface is unchanged. Users enabling `icc-lcms2` at build time get the press-grade CMM behaviour transparently through the same calls.

## Documentation

- Inline docstrings cite the §-numbers for every spec-derived behaviour
- All probe docstrings cite the spec section under test
- Documented spec-ambiguity markers describe the chosen reading and the alternate readings considered, with the byte value of the chosen reading pinned by sensitivity-verified probes

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have added byte-exact probes for every behaviour change
- [x] All existing tests pass with no byte-identity regressions across the prior suite
- [x] No new clippy warnings
- [x] No new `unsafe` blocks in production code (the optional `icc-lcms2` feature pulls a C dep through a safe Rust wrapper)
- [x] Compiles cleanly across all feature-flag combinations

### Surviving documentation markers (defensible-reading or policy choices, not implementation gaps)

The following constants document choices where ISO 32000-1 is genuinely silent on reader behaviour OR where a policy choice is required on spec-silent ground (e.g., bounded recursion, malformed-input handling). Each has a byte-exact probe pinning the chosen behaviour + the alternate readings noted in-source.

**Genuine spec ambiguities (6):**

- `HONEST_GAP_NONSEP_DEVICEN_GROUP` — §11.3.4 + §11.6.6 forbid `/CS /DeviceN` as a transparency group blend space; reader behaviour on malformed files declaring it is unspecified. Chosen: surface colorants as active spots through the pre-pass, fall back to the DeviceN alternate space.
- `HONEST_GAP_NONSEP_K_CHANNEL_FOR_NON_CMYK_FOUR_COMPONENT_ICC` — §11.3.5.3 K-channel rule is named only for `/DeviceCMYK` and calibrated CMYK; 4-component non-CMYK ICCBased blend spaces aren't covered. Chosen: apply the CMYK K-rule analogously.
- `HONEST_GAP_SPOT_LANE_UNSOURCED_PRESERVE_BACKDROP` — §11.7.3 strict source-expansion vs §11.7.4.3 CompatibleOverprint example disagree on whether unsourced spot lanes preserve backdrop or compose against source=0. Chosen: CompatibleOverprint preserve-backdrop semantics for implicit-not-named spots; explicit-zero-tint paints write tint 0.
- `HONEST_GAP_KNOCKOUT_DIFFERENT_INK_SPOT_INTERACTION` — §11.4.6.2 cross-ink knockout behaviour is underspecified when paints in a /K group target different spot inks. Chosen: per §11.7.4.3, preserve each ink's prior write across the group.
- `HONEST_GAP_DEVICEN_PROCESS_OVERPRINT_CLASS` — Table 149 row 6 admits broad ("DeviceN /Process attribution → process-class") and narrow ("any DeviceN source preserves backdrop on process channels") readings. Chosen: broad, consistent with §8.6.6.5 EXAMPLE 3.
- `HONEST_GAP_DEVICEN_PROCESS_MISMATCHED_NAMES` — §8.6.6.5 says /Components names must match /Names but doesn't specify reader behaviour on violation. Chosen: treat /Process as inert (preserves spot-tail intent over partial process attribution).

**Defensible policy choices (4):**

- `HONEST_GAP_KNOCKOUT_MERGE_BYTE_EQUALITY_SKIP` — /K accumulator skips per-byte merge when post == backdrop (observationally identical to "no paint"; documented as policy choice with byte-exact probe).
- `HONEST_GAP_SMASK_CYCLIC_G_UNBOUNDED_RECURSION` — `MAX_SMASK_DEPTH=32` cap to prevent OOM / stack-overflow on cyclic `/G` references. Spec doesn't require a depth limit; real implementations must have one.
- `HONEST_GAP_DEVICEN_PROCESS_ICC_PROFILE_MISMATCH` / `_R7` — three-state matrix: closes under `icc-lcms2` (real CMYK→CMYK retargeting through Lab PCS with BPC); remains a documented feature-level limitation under default `icc-qcms` (qcms 0.3.0 has no CMYK output pipeline); §10.3.5 fallback under no-ICC builds.

**Defensible malformed-input policy (1):**

- `HONEST_GAP_SMASK_BC_MALFORMED_ARITY` — when `/BC` array's component count doesn't match the declared blend space, falls back to black backdrop. Spec is silent on reader behaviour for malformed `/BC`.

### Optional `icc-lcms2` feature

Pure-Rust default behaviour (`icc-qcms`) is unchanged for existing users. WASM, C# AOT, and other pure-Rust-only targets continue to work without a C toolchain. Users needing PDF/X-3 / PDF/X-4 conformance with embedded-source-profile-to-OutputIntent retargeting opt in:

```bash
cargo build --features "rendering icc-lcms2"
```

The `icc-lcms2` feature pulls Little CMS 2 via the `lcms2` crate (well-maintained safe Rust wrapper, ~2.4M crates.io downloads). C build dep cost: comparable to the existing `fips` (AWS-LC) and `ocr` (ONNX Runtime) features in the project's opt-in surface.

### PR size

This is a large PR (~57k insertions across the §11 transparency surface plus the IccBackend trait extraction and lcms2 backend). Retroactively splitting after this much spec-correctness work has accumulated would mean substantial rebase risk for marginal review benefit — every spec area is interconnected through the sidecar and dispatch machinery. The change is structured as cohesive feature areas in this description so reviewers can focus on the surface they care about; the commit log preserves the bisect-ability of individual feature additions.


yfedoseev:
Thank you @RayVR — this is an impressive piece of work, and the non-separable blend implementation is spec-exact: I checked `src/rendering/blend_nonsep.rs` line-by-line against `docs/spec/pdf.md` §11.3.5.3 + Table 137 and `Lum` / `Sat` / `SetLum` / `ClipColor` / `SetSat` and all four mode formulas (Hue/Saturation/Color/Luminosity) match the spec exactly. 👏

Two things before merge:

**1. CI blocker (one-liner).** `Test (…, stable)` is red on ubuntu + macOS + windows. It is *not* a logic failure — it's a rustdoc broken-intra-doc-link, which fails the `cargo doc` step under `-D warnings`:

```
error: unresolved link to `Lcms2Backend`
  --> src/color/backend.rs:12
   |
12 | //!  - [`Lcms2Backend`] (`icc-lcms2`, opt-in): Little CMS via the …
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
```

`Lcms2Backend` isn't in scope at doc-build time (feature-gated), so the intra-doc link can't resolve. Either drop the brackets to a plain code span (`` `Lcms2Backend` ``) or gate/import the type so the link resolves under the `icc-lcms2` cfg. (The lone `PHP 8.5 on macos` failure is unrelated infra noise — git cleanup only in the log.)

**2. Minor spec caveat (non-blocking).** `compose_in_place` computes `out = αs·B + (1−αs)·Cb`, which is the §11.3.4 result only for an opaque backdrop (αb = 1). You already document this ("spec algorithm for an opaque backdrop … dest alpha already at 255"), and it's correct for the page-background-filled fixtures here — just flagging it so it's on the record for when transparent backdrops / nested groups come into play.

Once the rustdoc link is fixed and CI goes green, happy to take another pass.

RayVR:
@yfedoseev the docs have been fixed. The php CI job failed again.

yfedoseev (changes requested):
## Review: changes requested

Thorough multi-pass review of the §11 transparency surface + IccBackend work. **The engineering quality is high** — the spec work is careful, the non-separable blend math checks out against §11.3.5.3, the lcms2 FFI `Sync`/`DisallowCache` reasoning is sound, RAII cleanup is correct on all error paths, feature-gating across all four combos is correct, and the byte-exact tests are genuinely spec-derived (not snapshots). The `MAX_SMASK_DEPTH=32` cap, knockout /K snapshot-restore-merge, and spot-ink reserved-name handling all verified clean.

Requesting changes for the items below — all introduced by this PR. Two pre-existing issues found during review (CI doesn't run the rendering test tier; `output_intent_cmyk_profile` swallows profile errors) are filed separately as #711 and #712 and are **not** blockers for this PR.

> Note: because of #711, this PR's green CI does **not** exercise its own 260 tests. They pass locally under `--features rendering,test-support` (+ `icc-lcms2` subset), but please confirm, and the #711 fix should land so this suite is actually protected.

### 🔴 Blocker

**B2 — Unbounded recursion → stack overflow on malformed input.**
`src/document.rs:735` — the new `collect_inks_from_color_space` Pattern arm recurses into the underlying CS (`collect_inks_from_color_space(&underlying, doc, out)`) with **no depth bound and no visited-set**. A self-referential Pattern colour space (`5 0 obj [/Pattern 5 0 R]`) recurses until the process aborts. Reachable on every detection-ON render and every `render_separations` via `discover_page_spot_inks`. Please thread a depth counter or `HashSet<ObjectRef>` visited-set, mirroring `walk_form_xobject_tree_for_inks`.

### 🟠 High

**H1 — SMask form rendered at `Transform::identity()`.**
`page_renderer.rs` `apply_smask_after_paint_inner` renders the mask form at identity, ignoring `base_transform` (DPI scale + PDF→device y-flip). The mask is mis-scaled/y-flipped at any DPI ≠ 72; it only happens to work at the 100×100 / 72-DPI audit fixture. Per §11.6.5.2 the mask must render in the device space in effect. Thread `base_transform` into `render_form_xobject` instead of `Transform::identity()`.

**H2 — Per-pixel ICC profile hashing (perf).**
`apply_cmyk_compose_after_paint_with_coverage` and `apply_overprint_after_paint_with_coverage` call `icc_transform_cache.get_or_build(...)` **inside the per-pixel loop**, and `get_or_build` hashes the entire ICC profile byte blob each call → millions of full-profile hashes on a full-page transparency fill. The sibling diff-driven path already hoists the `Arc<Transform>` out of the loop — please do the same here.

**H3b — Silent K=0 press output.**
`page_renderer.rs` `resolve_rgb_paint_to_cmyk` turns a `None` from `output_intent_cmyk_profile()` into a silent `(1-R,1-G,1-B, K=0)` result. The math fallback itself is acceptable when no profile was requested, but combined with #712 it means "broken declared profile" silently degrades press output. The diagnostic belongs upstream (#712); please coordinate so the K=0 path only runs silently when color management was genuinely never requested.

### 🟡 Medium

**M1 — Premultiplied-alpha blend bug.**
`blend_nonsep.rs` `compose_in_place` reads tiny_skia bytes as straight RGBA, but tiny_skia stores **premultiplied** alpha. For partial-alpha non-separable paints (Hue/Sat/Color/Luminosity) colors are wrong; only correct when src+dst alpha = 255. Also assumes an opaque backdrop, so it's wrong inside isolated/transparent groups. Either un-premultiply before the §11.3.5 math, or explicitly gate to opaque-backdrop and document the limitation.

**M2 — Uncached CMYK retarget / OutputIntent re-parse per paint.**
On `icc-lcms2` builds each DeviceN /Process /ICCBased N=4 `scn` re-walks `/OutputIntents`, re-decodes + re-parses both profiles, and rebuilds the CLUT. Cache the retarget transform keyed by `(src_hash, dst_hash, intent)` and memoize the OutputIntent profile `Arc` (the `SrgbToCmyk` path already has a per-page cache to mirror).

**M3 — Detection gates miss indirect / nested-form transparency.**
`sidecar.rs` `ext_g_states_signal_transparency{,_only}` read `/CA /ca /SMask /BM` without dereferencing indirect refs (e.g. `/SMask 12 0 R`), and don't recurse into a Form XObject's own `/Resources/ExtGState`. Such a page silently routes to the per-plate walker and loses transparency in the separations. Deref the values; consider documenting or extending the nested-form boundary.

### 🟢 Low / docs
- Doc typo: "the embedded profile parses but the destination profile parses" → "fails to parse" (`color/mod.rs:381` and the mirrored comment in `sidecar.rs`).
- Stale test docstrings claim `#[ignore]` / "at HEAD this bypasses…" on tests that actually pass (`test_46_round6_qa_pass.rs:307`, `test_transparency_flattening_qa_round2.rs:884`, round1/round2 headers) — clean up so they don't misdescribe state.
- `test_46_round7_icc_retargeting.rs:429` `compute_retarget_reference` calls lcms2 with the same profiles/intent/flags as production — it's a same-engine round-trip, not an independent oracle. Rename, or add one hand-derived Lab anchor.
- A couple of adversarial probes assert only `alpha == 255` rather than the mechanism their docstring describes; tighten to honor the "byte-exact, no tolerances" promise.

---
Happy to pair on B2/H1 — they're localized. Architecture and approach look right; this is polish before merge.

RayVR:
@yfedoseev — thanks for the careful pass. All 8 findings addressed (commits `47209fe..eb79cce`), plus a parallel bug surfaced while pinning M3. Per-finding summary below.

| Sev | Ref | Approach | Sensitivity-verified |
|---|---|---|---|
| 🔴 B2 | `47209fe` | `collect_inks_from_color_space` Pattern arm now takes a depth counter + `HashSet<ObjectRef>` visited-set, mirroring `walk_form_xobject_tree_for_inks`. | self-referential `[/Pattern 5 0 R]` test stack-overflows without the fix |
| 🟠 H1 | `8d6eef4` | `base_transform` threaded through `apply_smask_after_paint{,_inner}` and all 12 call sites; mask renders in device space per §11.6.5.2. | DPI=144 asymmetric mask probe: pixel (150, 50) reads white pre-fix |
| 🟠 H2 | `a294be7` | `get_or_build` lookup hoisted out of 4 per-pixel loops — the two you named, plus two sibling diff-driven helpers that turned out to also walk per-pixel. | counting probe: 80,008 pre-fix lookups → ≤32 post-fix for 8 paints |
| 🟠 H3b | `c4693d4` | K=0 fallback now logs a warning when `/OutputIntents` is declared but the profile didn't yield. Coordinates with #712 — the upstream silent-swallow path is unchanged; this just surfaces the symptom locally. | warning-capture probe |
| 🟡 M1 | `cc79252` | Un-premultiply backdrop + source before §11.3.5, run spec math on straight colour, re-premultiply on write-back, full §11.3.4 result with αb < 1. | 4 partial-alpha byte-exact probes (Hue / Sat / Color / Lum) all fail pre-fix |
| 🟡 M2 | `52d1e19` | Per-page CMYK retarget cache keyed by `(src_hash, dst_hash, intent)` + OutputIntent profile `Arc` memoised; mirrors the SrgbToCmyk pattern. | counting probe: build_count == 1 across 6 paints |
| 🟡 M3 | `8d51c1e` | `resources_declare_transparency_or_overprint` resolves indirect refs and recurses into Form XObject `/Resources/ExtGState`. | 5 unit probes; 4 of 5 fail without each respective deref |
| 🟢 Low | `5199aac` | sidecar typo fix; 4 stale docstrings refreshed; `compute_retarget_reference` → `compute_retarget_self_check` + hand-derived Lab anchor added (not just a rename); 2 adversarial probes tightened to byte-exact. | — |

**Parallel finding surfaced while validating M3** — `eb79cce`: `parse_ext_g_state_inner` in `src/rendering/ext_gstate.rs` was reading `/ca`, `/CA`, `/BM`, `/OP`, `/op`, `/OPM`, and the `/SMask` sub-entries `/S`, `/BC`, `/TR` directly and applying typed accessors without dereferencing first. Per §7.3.10 any direct value can be replaced by an indirect reference; `.as_real()` / `.as_bool()` / `.as_integer()` / `.as_name()` / `.as_array()` all return `None` on `Object::Reference`, so a PDF emitting `/ca 3 0 R` silently dropped `fill_alpha` to default — meaning M3's correctly-routed transparency was then ignored by the receiver. Closing M3 without this left the wire intact but the receiver deaf. Fixed by threading `doc.resolve_object` through every entry plus `/BM`-array element resolution; 6 new unit tests pin each surface, all initially failing.

**Local verification** (CI re-run on `eb79cce` in flight): `cargo test --features rendering,test-support` is 5,665 lib + integration + 137 doctests green; the `icc-lcms2` variant is also green; both doc builds clean under `-D warnings`; `cargo check --lib --no-default-features` clean. The structural CI test-tier gap from #711 remains worth tracking — until that lands, local runs are the ground truth for this branch.

yfedoseev (approved):
## Approving — all requested changes verified addressed

Re-reviewed against the current head (`294b504a`). Every item from the earlier *changes requested* pass is resolved in the code (not just claimed — I checked each against the source):

**🔴 Blocker**
- **B2** (unbounded `collect_inks_from_color_space` recursion) — now threads a `visited` set + `depth` through the Pattern arm (`src/document.rs:722`, recursion at `:769`), with a self-referential-colour-space regression test that previously aborted. ✅

**🟠 High**
- **H1** — SMask form now renders at `base_transform` instead of `Transform::identity()` (`page_renderer.rs:apply_smask_after_paint_inner`), so the mask is correctly scaled/y-flipped at DPI ≠ 72 per §11.6.5.2. ✅
- **H2** — `icc_transform_cache.get_or_build(...)` is now hoisted **out of** the per-pixel loop in both `apply_cmyk_compose_after_paint_with_coverage` (`:4817`, loop at `:4820`) and `apply_overprint_after_paint_with_coverage` (`:5131`, loop at `:5137`). ✅
- **H3b** — the K=0 RGB→CMYK fallback now emits a `log::warn` when `/OutputIntents` is declared but the profile lookup returns `None` (`resolve_rgb_paint_to_cmyk`, `:5281`), so a broken declared profile no longer degrades press output silently. ✅

**🟡 Medium**
- **M1** — `compose_in_place` now un-premultiplies on input and re-premultiplies on output for the §11.3.5.3 non-separable math (`blend_nonsep.rs:60`). ✅
- **M2** — CMYK→CMYK retarget transform is now cached (`cmyk_retarget_build_count` counter; per-`(src,dst,intent)` memoization). ✅
- **M3** — `ext_g_states_signal_transparency*` now resolves indirect refs for `/CA /ca /SMask /BM` and recurses into a Form XObject's own `/Resources/ExtGState` (`sidecar.rs:514+`). ✅

**🟢 Low/docs** — the "fails to parse" typo is corrected in both `color/mod.rs` and the `sidecar.rs` mirror. ✅

The §11 transparency math (separable + non-separable blend modes, §11.3.4 compositing, soft-mask luminosity/alpha, group isolation/knockout) and the ICC `/N`-vs-profile-signature validation are spec-faithful.

**One optional follow-up (non-blocking, not a regression):** non-isolated groups don't perform §11.4.8 backdrop *removal* before compositing onto the page, which slightly double-counts the backdrop for partial-alpha non-isolated groups. Worth a separate issue rather than holding this PR.

**Separate from this approval (per our policy that CI/process ≠ review blockers):** the branch is behind `main` and needs a rebase (which should also clear the Windows `setup-ruby` failure, since `ruby/setup-ruby` 1.312.0 is now on main), and per #711 the green CI doesn't yet exercise the 260 rendering tests — please confirm those pass locally before merge. Great work.

RayVR:
@yfedoseev the rendering tests pass locally.

PR #678, opened 2026-06-08
docs: document new page-oriented Python API

### LINKED ISSUE #371: Pythonic API (UX Suggestion)
Instead of the current Python API:
```python
from pdf_oxide import PdfDocument

# path can be str or pathlib.Path; use with for scoped access
doc = PdfDocument("paper.pdf")
# or: with PdfDocument("paper.pdf") as doc: ...
text = doc.extract_text(0)
chars = doc.extract_chars(0)
markdown = doc.to_markdown(0, detect_headings=True)
```

I suggest the following IMO more Pythonic API:

```python
from pdf_oxide import PdfDocument

with PdfDocument("paper.pdf") as doc:
    print(len(doc))        # number of pages

    for page in doc:
        text = page.text          # computed on demand
        chars = page.chars        # computed on demand
        md = page.markdown(detect_headings=True)
```

## Design principles

1. Document = container
Represents the whole PDF
Provides sequence semantics (__len__, __getitem__, __iter__)
Does not eagerly load pages

2. Page = unit of work
Encapsulates all page-level operations
Created cheaply
Does not compute anything upfront

3. Lazy + cached extraction
Expensive operations are computed only when accessed
Results are cached per page
page.text   # triggers text extraction once
page.chars  # triggers char extraction once

### PdfDocument

```python
doc = PdfDocument(path)

len(doc)
doc[i]
for page in doc: ...
Page
page.text        # property (lazy)
page.chars       # property (lazy)
page.markdown()  # method (has options)
```

## Performance model

Iterating pages is cheap
Creating Page objects is cheap
Extraction (text, chars, markdown) is lazy and cached
Optional shared internal parsing to avoid duplicate work

SeanPedersen:
Closes #371

## Summary

- Updated the **Quick Start** Python snippet in `README.md` and `python/README.md` to use the new page-oriented idiom (`len(doc)`, `for page in doc`, `page.text`, `page.markdown()`)
- Added a **"Page objects"** section in `python/README.md` documenting the full `Page` surface: lazy properties (`text`, `chars`, `words`, `lines`, `spans`, `tables`, `images`, `annotations`, `paths`) and methods (`markdown()`, `html()`, `plain_text()`, `render()`, `search()`)
- Restructured the **"Python API"** section in `README.md` with a "Page-oriented API" subsection
- Kept document-level methods (`extract_text(0)` etc.) documented as the lower-level alternative for backward compatibility

## Test plan

- [ ] Verify code snippets match the published PyPI wheel API
- [ ] Confirm all `page.*` properties and methods shown are present in `src/python.rs` (`PyDocPage`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,17 @@ The fastest PDF library for text extraction, image extraction, and markdown conv
 ```python
 from pdf_oxide import PdfDocument
 
-# path can be str or pathlib.Path; use with for scoped access
+with PdfDocument("paper.pdf") as doc:
+    print(len(doc))                          # number of pages
+    for page in doc:
+        text = page.text                     # lazy property
+        chars = page.chars                   # lazy property
+        md = page.markdown(detect_headings=True)
+
+# Direct page access by index
 doc = PdfDocument("paper.pdf")
-# or: with PdfDocument("paper.pdf") as doc: ...
-text = doc.extract_text(0)
-chars = doc.extract_chars(0)
-markdown = doc.to_markdown(0, detect_headings=True)
+page = doc[0]
+text = page.text
 ```
 
 ```bash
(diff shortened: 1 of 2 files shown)


yfedoseev (approved):
Thanks for this, @SeanPedersen! 🙏 Really nice cleanup — the page-oriented idiom reads much better than the old document-level snippets, and the new "Page objects" section is a great addition.

I verified every documented surface against `src/python.rs` (`PyDocPage`) and it all checks out: `len(doc)`, `doc.version()`, `doc[0]`/`doc[-1]` negative indexing, iteration, all page getters, `markdown()`/`html()`/`plain_text()`/`render()`/`search()`, `within()`, region extraction, and the `ExtractionProfile` factories.

I pushed one small follow-up commit (71fe5a83) correcting four type-hint comments: `words` returns `PyWord` objects (no `Word` class), and `images`/`annotations`/`paths` return `list[dict]` rather than typed classes. Everything else was spot-on. 🚀

PR #728, opened 2026-06-10
fix(python): drop leaked Py<Self> receiver params from generated .pyi

norbusan:
## Description

rylai injects an implicit `self` for every #[pymethods] method but does not recognise a by-value receiver (`fn page(slf_handle: Py<Self>, ...)`), so it re-emits the receiver as an extra positional parameter. The stub then read e.g. `def page(self, slf_handle: DocumentBuilder, width, height)`, making `builder.page(w, h)` (correct at runtime) trip mypy's call-arg / arg-type checks and forcing downstream `# type: ignore`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Changes Made

Fix in two parts:
- Add #[pyo3(signature = (...))] to the affected ordinary methods (page, a4_page, letter_page, within). rylai honours the signature attribute, which never lists the receiver, so the leak disappears.
- pyo3 forbids `signature` on magic methods, so __iter__/__getitem__ still leak. Add scripts/fix_self_handle_stub.py, wired into rylai's `format` hook in rylai.toml, to strip any param named slf/slf_handle from the generated stub. It runs on every generation (CI included) and uses ast for docstring-safe, offset-based removal.

## Testing

Verified: regenerated stub has no receiver leaks; mypy --strict and basedpyright report 0 errors on the affected call sites; runtime behaviour and stub symbol parity unchanged.

- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] I have run `cargo fmt`

- [ ] Python tests pass

## Documentation

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to PDFOxide are documented here.
 
+## [Unreleased]
+
+### Fixed
+
+- **Fixed Python type stubs leaking the pyo3 `Py<Self>` receiver as a positional parameter** — methods implemented in Rust with a by-value receiver (`fn page(slf_handle: Py<Self>, …)` — the idiom pyo3 uses to hand a method an owned handle to its own instance) were emitted by the rylai stub generator with that receiver re-exposed *alongside* the injected `self`.
+
 ## [0.3.63] - 2026-06-09
 
 > CJK extraction-quality fixes — vertical-CJK (tategaki) reading order no longer mis-fires on horizontal Japanese text, CJK glyphs no longer surface as Kangxi-radical codepoints, and Korean number spacing is preserved — plus recovery of dropped inter-word spaces on tightly-typeset PDFs, and routine dependency updates.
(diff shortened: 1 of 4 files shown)


yfedoseev (approved):
Approve. The generator fix is correct: `#[pyo3(signature = …)]` on the four by-value-receiver methods plus a `fix_self_handle_stub.py` post-processor (wired into rylai's format hook) to strip residual `slf`/`slf_handle` from magic methods where PyO3 forbids the signature attr. Verified by running the script against the real stub — all 6 leak sites cleaned, every legitimate param preserved, idempotent.

No pdf.md surface — pure Python stub codegen, no PDF semantics touched.

Two follow-ups worth doing (non-blocking, but the fix isn't fully shipped without #1):
1. **Regenerate and commit `python/pdf_oxide/pdf_oxide.pyi`** — it's checked in but was NOT regenerated here, so it still contains the leaks (and is missing `pages`). Consumers won't see the fix until the artifact is regenerated.
2. Add a small pytest over a fixture stub (first-arg receiver, sole-arg receiver, magic-method, and a negative case) to lock the offset-math edge cases.

norbusan:
Concerning the first suggestion: the `pdf_oxide.pyi` file is .gitignore-d, and updated via CI. Should I update and commit it anyway?

Issue #711, opened 2026-06-09
CI: rendering / test-support integration tests never run in CI


yfedoseev:
## Summary
The entire `feature = "rendering"` integration-test tier is **never executed in CI**. This is pre-existing (independent of any single PR) and means rendering regressions can land on `main` undetected.

## Evidence
- `Cargo.toml:314` — `default = ["icc", "legacy-crypto"]`. `rendering` and `test-support` are **not** default features.
- The only full test step is `.github/workflows/ci.yml:162` — `cargo test --verbose` — runs under default features only, so every `#![cfg(all(feature = "rendering", ...))]` test file compiles to **0 tests**.
- `main` already has 10+ rendering-gated test files silently skipped: `test_render_output_intent`, `test_jbig2_rendering`, `test_pdfa_roundtrip`, `test_render_ocg_filtering`, `test_render_real_world_fixture`, `test_render_resolution_pipeline_qa_wave1/2`, `test_concurrent_document_reads`, `test_debug_visualization`, `test_render_region_and_fit`.
- `grep 'cargo test.*rendering' .github/workflows/ci.yml` → no matches. `test-support` appears nowhere under `.github/`.
- The "extended" build job that enables `rendering` only runs `cargo build`, never `cargo test`.

## Impact
A green CI does **not** prove the rendering test suite passes. Large new rendering test suites (e.g. the §11 transparency work in #674, ~19k lines / 260 tests) fall straight into this dead zone.

## Suggested fix
```yaml
- run: cargo test --features rendering,test-support
- run: cargo test --features rendering,test-support,icc-lcms2   # separate job; installs Little CMS 2
```

PR #730, opened 2026-06-10
feat(render): substitute Adobe predefined CIDFonts via bundled CJK fallback

### LINKED ISSUE #727: Rendering: predefined Adobe CIDFonts (Ryumin-Light / GothicBBB-Medium / STSong-Light / etc.) render blank — ISO 32000-2 §9.7.5.2 substitution missing
## Summary

Vertical Japanese / horizontal CJK PDFs that reference Adobe's predefined CIDFonts (Ryumin-Light, GothicBBB-Medium, STSong-Light, MHei-Medium, HYSMyeongJo-Medium, etc.) without embedding glyph outlines render with **zero CJK glyphs** at present. Pages come out either fully blank or carry only the embedded Latin subset, depending on whether the PDF mixes other fonts.

ISO 32000-2:2020 §9.7.5.2 requires the opposite: "A PDF processor **shall** support Adobe-CNS1-7, Adobe-GB1-5, Adobe-Japan1-7 and Adobe-KR-9 character collections." The renderer should substitute a covering font when the source PDF references a predefined CID font by name but supplies no outlines, instead of silently dropping every glyph.

## Reproducer

Two public-domain fixtures from the `pdfminer.six` MIT-licensed sample corpus:

- **`jo.pdf`** (3.4 KB) — Kenji Miyazawa, "Haru to Shura" preface, vertical Japanese typeset via pTeX. Single CID font: `Ryumin-Light-Identity-V`. No embedded outlines.
- **`kampo.pdf`** (28 KB) — Japan Official Gazette page in vertical Japanese. Mix of `Ryumin-Light-Identity-H`, `Ryumin-Light-Identity-V`, `GothicBBB-Medium-Identity-H`, plus one embedded `Type1C` Latin subset (`GMALPM+DFHSMincho-W3G014`).

Download:
- https://raw.githubusercontent.com/pdfminer/pdfminer.six/master/samples/jo.pdf
- https://raw.githubusercontent.com/pdfminer/pdfminer.six/master/samples/nonfree/kampo.pdf

Render at 200 DPI via the existing `render_pages` example:

```
cargo run --release --features rendering --example render_pages -- jo.pdf out/ 1
cargo run --release --features rendering --example render_pages -- kampo.pdf out/ 1
```

Current behaviour on `upstream/main` at `9f7786db` (v0.3.63):
- `jo.pdf` → fully blank PNG (0 non-white pixels of 3.74 M total).
- `kampo.pdf` → header shows only the Latin digits "20 4 25 4817 4"; the body shows empty bounding-box outlines where each kanji should be. ~21 k non-white pixels of 3.87 M (all from the embedded Latin subset).

A reference render via `pymupdf` at the same DPI shows full CJK content in both cases.

## Why the data is being lost

Text extraction is **not** affected — `extract_with_pipeline` recovers 98 % of CJK characters from `jo.pdf` and 96 % from `kampo.pdf`. So the CID → Unicode tables (`src/fonts/cid_mappings/adobe_*.rs`) are correct and complete. The break is purely in the page-render path: when a CID font has no `/FontFile{,2,3}` outlines in its `/FontDescriptor`, the renderer has nowhere to source the glyph shape from, and silently emits nothing — no `.notdef` rectangle, no diagnostic, no fallback.

## Impact

- Anything typeset by pTeX / LaTeX-CJK / Word-with-bundled-Acrobat-font / older InDesign that ships pre-PDF/X-1a-strict — i.e. most of the consumer Japanese PDF ecosystem from ~2000-2015 — renders blank.
- Japanese government PDFs (Official Gazette, MLIT, MOF) routinely reference predefined CID fonts without embedding outlines because Acrobat shipped them. Today's renderer drops all the text.
- Same story for Chinese (Adobe-GB1, Adobe-CNS1) and Korean (Adobe-Korea1).
- Extraction is fine, so this only surfaces when callers actually rasterise pages.

## Proposed direction

Substitute the bundled `DroidSansFallbackFull.ttf` (already in `src/fonts/assets/`, already wired for the form-flatten path via `src/fonts/form_fallback.rs::font_bytes(Fallback::Cjk)`) when:

1. The font is a Type 0 CIDFont.
2. The descendant `FontDescriptor` has no `/FontFile`, `/FontFile2`, or `/FontFile3`.
3. The `/BaseFont` name (after stripping the 6-char subset prefix and the predefined CMap suffix) matches a known Adobe predefined CIDFont.

Routing: CID → Unicode via the existing `adobe_*.rs` registry table → glyph_id via Droid Sans Fallback's `cmap` → paint via the existing TrueType outlines path. The PDF's `/W` / `/DW` advance widths are honoured; WMode-1 vertical advance routing is preserved.

Limits accepted by the substitution: Droid Sans Fallback is sans-serif, so when a Mincho/Song serif face is requested the rendered output is sans-serif. Vertical-form glyph variants (rotated punctuation) are not provided. These are documented trade-offs; the alternative (continue dropping all glyphs) is worse.

Implementation can be feature-gated (`cjk-render-fallback`) so users who don't need CJK fallback keep a slimmer binary. No new dependencies required — every piece of infrastructure already exists in-tree.

RayVR:
## Summary

Closes #727. When a Type 0 CIDFont references an Adobe predefined CIDFont name (Ryumin-Light, GothicBBB-Medium, STSong-Light, MHei-Medium, HYSMyeongJo-Medium, etc.) without embedding the actual glyph outlines, the page renderer can now substitute the bundled Droid Sans Fallback so the CJK content actually renders. Before this change, every glyph from such a font was silently dropped — `jo.pdf` rendered fully blank, `kampo.pdf` rendered only the Latin numerals while the kanji body stayed empty.

ISO 32000-2 §9.7.5.2 requires support for Adobe-CNS1-7, Adobe-GB1-5, Adobe-Japan1-7 and Adobe-KR-9 character collections — this PR brings the renderer in line with that requirement.

## Approach

Reuses existing in-tree infrastructure:

- `src/fonts/assets/DroidSansFallbackFull.ttf` was already bundled (used by the form-flatten path).
- `src/fonts/cid_mappings/adobe_japan1.rs`, `adobe_gb1.rs`, `adobe_cns1.rs`, `adobe_korea1.rs` already provide the CID → Unicode tables.
- `src/fonts/form_fallback.rs::font_bytes(Fallback::Cjk)` is the proven prior art for routing to the bundled CJK font in another path.

Three new pieces:

1. **`src/fonts/predefined_cidfont.rs`** — name→character-collection registry covering ~30 known Adobe predefined CIDFont names across all four character collections, with subset-prefix (`ABCDEF+`) and CMap-suffix (`-Identity-V`, `-Identity-H`, `-90ms-RKSJ-H`, etc.) stripping. Longest-match preference protects the legacy single-letter `-H` / `-V` suffixes from mis-stripping (e.g. `STSong-Light-GBK-EUC-H` correctly resolves to `STSong-Light`).
2. **Detection at font-load time** — `FontInfo` gets an `Option<CharacterCollection>` field populated when subtype is Type 0, `embedded_font_data.is_none()`, and the BaseFont matches the registry.
3. **`TextRasterizer::render_substituted_cjk`** (feature-gated on `cjk-render-fallback`) — walks CIDs from the existing `TextCharIter`, maps CID → Unicode via the registry → glyph_id via Droid Sans Fallback's `cmap` → paints through the existing tiny-skia outline path. Advance widths come from the PDF's own `/W` / `/DW`. Vertical writing mode is honoured (advance along y, `v_x`/`v_y` origin offsets applied). When the CID has no Unicode mapping OR Droid Sans has no glyph, paint is skipped but advance is preserved (so the layout doesn't collapse).

The form-flatten path's existing behaviour is unchanged — the new feature flag (`cjk-render-fallback`) is independent of `cjk-form-fonts`. Both gates can be enabled together, alone, or neither. No new dependencies; the bundled font and the CID mapping tables were already in-tree.

## Verification

Tests render the new `tests/fixtures/vertical_cjk/jo.pdf` and `kampo.pdf` (committed from the pdfminer.six MIT corpus — see `tests/fixtures/vertical_cjk/ATTRIBUTION.md`) and count non-white pixels in the pixmap as a coarse "is anything being drawn" check. Each test is **sensitivity-verified** by toggling the substitution path off and confirming the assertion fails.

| Test | Pre-fix non-white px | Post-fix | Δ |
|---|---:|---:|---:|
| Synthetic Adobe-Japan1 PDF | 0 | 7,340 | +7,340 |
| `jo.pdf` (vertical Japanese, Ryumin-Light-Identity-V only) | 0 | 119,186 | +119,186 |
| `kampo.pdf` (mixed Ryumin / GothicBBB-Medium / HeiseiKakuGo-W5 + Latin subset) | 21,212 (Latin only) | 277,397 | +256,185 |

Plus 12 unit tests in `predefined_cidfont.rs::tests` covering bare and suffixed names per collection, subset-prefix strip, suffix-strip boundary protection, and the CID → Unicode chain.

## Pre-push gates

| Gate | Result |
|---|---|
| `cargo fmt --check` | green |
| `cargo clippy --features rendering,test-support,cjk-render-fallback --all-targets -- -D warnings` | green |
| `cargo check --lib --no-default-features` | green |
| `cargo test --features rendering,test-support,cjk-render-fallback` | 288 test binaries, 0 failures |
| `cargo test --features rendering,test-support` (regression guard) | 0 failures |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` | green |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --no-default-features` | green |

Note: the rendering test tier still doesn't run in CI (per #711); the suite above is local on macOS.

## Documented trade-offs

- **Style fidelity**: Droid Sans Fallback is a sans-serif face. When a Mincho/Song serif typeface like Ryumin-Light is requested, the rendered output is sans-serif. The module header documents this explicitly. The alternative (continue dropping every glyph) is worse for every reasonable consumer.
- **Vertical-form glyph variants**: Droid Sans Fallback is a horizontal font. Layout advances vertically via the existing WMode-1 routing, but vertical-form glyph variants (rotated punctuation, vertical-baseline-shifted parens via `vert` / `vrt2` OpenType features) are not provided. The glyph shapes themselves are the horizontal form.
- **No system-font integration**: deliberate. Keeping the binary self-contained means no CoreText / DirectWrite / fontconfig dependency and deterministic output regardless of host. Trade-off: ~3.4 MB binary size when the feature is enabled.
- **Existing rendering paths unchanged when the feature is off**: the substitution check is a no-op without `cjk-render-fallback`. Documents that embed their own outlines never touch this code path.

## Why feature-gated

Binary-size sensitivity. The 3.4 MB Droid Sans Fallback only matters for consumers actually rendering CJK; embedded-pipeline and OCR-only consumers don't need it in their binary. The form-flatten path already follows the same opt-in pattern via `cjk-form-fonts`.

## Test plan

- [x] Unit tests for the predefined-CIDFont name registry (12 tests, all positive and negative cases)
- [x] Synthetic minimal Adobe-Japan1 PDF renders non-blank
- [x] Real-world `jo.pdf` (Miyazawa Kenji preface) renders non-blank
- [x] Real-world `kampo.pdf` (Japan Official Gazette) renders kanji content (not just Latin numerals)
- [x] All three integration tests sensitivity-verified by toggling the substitution off
- [x] Regression guard: full test suite passes with `cjk-render-fallback` disabled
- [x] Doc build clean under `-D warnings` for both default and `--no-default-features`


src/fonts/font_dict.rs
```diff
@@ -1170,6 +1187,35 @@ impl FontInfo {
...
+        //      the Type0 wrapper AND the CIDFont descendant have both been
+        //      checked for `/FontFile{,2,3}`. With outlines the document is
+        //      self-contained and we have nothing to substitute against.
+        //   3. The base font name (after subset-prefix + CMap-suffix strip)
+        //      matches one of the registered predefined names from
+        //      Technical Notes #5078 / #5079 / #5080 / #5093.
+        // When all three hold, the renderer routes the paint through the
+        // bundled covering font; otherwise we leave `cjk_substitution` at
+        // `None` and the existing render path runs unchanged.
+        let cjk_substitution = if subtype == "Type0" && embedded_font_data.is_none() {
```
yfedoseev: **🔴 Blocker — substitution masks a failed embedded-font decode.**

The trigger is `embedded_font_data.is_none()`, but `embedded_font_data` goes `None` *both* when there is no `/FontFile2` **and** when one is present but fails to load/decode — those errors are dropped by `.ok()` at lines 615/618 (FontFile2), 637/640 (FontFile3), and 667/670 (FontFile) with no log. So a document that *does* embed a real CJK (e.g. Mincho) outline, but whose font program fails to decrypt/parse, silently falls through to a sans-serif Droid Sans substitution — hiding the actual decode defect and degrading output with no signal.

Gate substitution on **descriptor `/FontFile{,2,3}` key absence** (compute a `has_font_program` bool from key presence), not on the post-extraction `Option`. If a key is present but extraction failed, log a warning (surfacing the error currently swallowed by `.ok()`) and do **not** substitute. Contrast `extract_embedded_font_from_descriptor` (~`:1797`), which at least `warn!`s before degrading.

src/rendering/text_rasterizer.rs
```diff
@@ -1478,6 +1539,191 @@ impl TextRasterizer {
...
+
+        let mut x_cursor: f32 = 0.0;
+        let mut y_cursor: f32 = 0.0;
+        let wmode = gs.text_wmode;
+
+        let mut glyphs_painted: usize = 0;
+        let mut glyphs_missing: usize = 0;
+
+        for (char_code, _) in TextCharIter::new(bytes, Some(font_info)) {
+            let cid = char_code;
```
yfedoseev, line 1621: **🔴 Blocker — raw char-code used as CID for non-Identity CMaps.**

`let cid = char_code;` then `collection.cid_to_unicode(cid)` below assumes `code == CID`, which only holds for `Identity-H`/`Identity-V`. But `KNOWN_CMAP_SUFFIXES` in `predefined_cidfont.rs` deliberately accepts dozens of non-Identity CMaps (`90ms-RKSJ-H`, `GBK-EUC-H`, `B5pc-H`, `KSC-EUC-H`, `UniJIS-UCS2-H`, …), so a font like `Ryumin-Light-90ms-RKSJ-H` (the module's own flagship example) is flagged for substitution and then mis-decoded: its Shift-JIS code is **not** an Adobe-Japan1 CID, so the lookup is a miss (blank, advance preserved → invisible gap) or a wrong-but-present mapping (wrong glyph, no diagnostic).

Route `char_code → CID` through the existing `lookup_predefined_cmap` (already used by the non-substituted path) before indexing the collection — or restrict substitution to Identity CMaps until that's wired. Only `Identity-H` is exercised by the tests, so this whole class is currently unverified.

yfedoseev (changes requested):
## Review: request changes

Great work — the intent is genuinely spec-motivated (ISO 32000-2 §9.7.5.2 requires Adobe-Japan1/GB1/CNS1/KR support), the BaseFont subset-prefix stripping matches §9.6.4, and substituting a covering font for a non-embedded CIDFont is a legitimate conforming-reader strategy. Registry longest-match / `-H`/`-V` boundary protection, feature-gating (clean no-op when the flag is off), and unwrap-safety all check out. Two blockers before merge (see inline comments), plus two lower-priority notes below.

### Lower-priority (non-blocking)

**Collection chosen from font *name*, not `/CIDSystemInfo`** (`font_dict.rs:1204`). The character collection — i.e. which CID→Unicode table is used — is derived purely from the base-font name. The descendant font's `/CIDSystemInfo` Registry/Ordering is parsed and available in the same function but never consulted. A producer that names a font `STSong-Light` (→ Adobe-GB1) but pairs it with an `Identity`-ordered descendant would be interpreted against the wrong table with no diagnostic. Suggest cross-checking `cid_system_info.ordering` and preferring it (or declining substitution) on mismatch.

**Returned advance omits horizontal scaling (Tz/Th)** (`text_rasterizer.rs`, success path vs the bundled-face-missing branch). The success path accumulates `x_cursor` without `h_scale`, but the load-failure fallback routes through `measure_text_bytes`, which *does* apply `h_scale`. Per §9.4.4, `tx = ((w0·Tfs)+Tc+Tw)·Th`, so Th belongs in the returned text-space advance — and the two branches of the same function should agree. (`render_cid_direct` shares the no-h_scale convention, so this is partly pre-existing, but the new function disagrees with itself.)

Nice touch on the sensitivity-verified tests — toggling the path off and confirming the assertion fails is exactly right. The main gap is that only `Identity-H` is exercised (see blocker #2).

src/fonts/font_dict.rs
RayVR: Fixed in 8d732de0. The gate now computes `has_font_program` from `/FontFile{,2,3}` key presence on **both** descriptors (the Type0 wrapper's, checked inline, and the descendant's — `extract_embedded_font_from_descriptor` and `parse_descendant_fonts` now return key presence alongside the extracted bytes), and substitution requires the key to be absent. The errors previously dropped by `.ok()` at the six wrapper-descriptor sites are now logged via `inspect_err` before the `.ok()`, and a dedicated warning names the font when a declared program fails to extract and substitution is declined.

Sensitivity-verified regression test: `cjk_substitution_declined_when_font_program_key_present_but_unextractable` builds a descriptor whose `/FontFile3` points at a non-existent object — it asserted `Some(AdobeJapan1)` pre-fix and `None` post-fix.

src/rendering/text_rasterizer.rs
RayVR: Fixed in 8d732de0, taking the restrict-to-Identity option. Routing through `lookup_predefined_cmap` doesn't close the gap: that function maps CID→Unicode, not charcode→CID, and the missing piece is bigger than glyph selection — `/W` advance widths are also CID-keyed, so a non-Identity font needs real charcode→CID CMap data (Adobe CMap resources we don't bundle) before either lookup is sound. That's now an explicit follow-up.

The load-time gate requires the `/Encoding` to resolve to `Encoding::Identity` (Identity-H/V names, or the Adobe-collection identity CMap streams `parse_encoding` already folds into that variant). Fonts like `Ryumin-Light-90ms-RKSJ-H` are still recognised by the name registry but skipped with an info log, and the paint loop documents the code == CID invariant the gate guarantees.

One observable consequence worth noting: kampo.pdf's `HeiseiKakuGo-W5-UniJIS-UCS2-H` font was previously substituted with its UCS-2 codes mis-read as CIDs — i.e. it painted wrong glyphs. It now renders blank-with-correct-advance until the CMap pass lands; the fixture's ink assertions still pass on the Identity-encoded fonts (277k non-white pixels vs the 5k gate).

Sensitivity-verified regression test: `cjk_substitution_requires_identity_cmap_encoding` (asserted `Some(AdobeJapan1)` pre-fix, `None` post-fix).

RayVR:
@yfedoseev all four findings addressed in `8d732de0..09396218`:

| Sev | Finding | Fix | Sensitivity-verified |
|---|---|---|---|
| 🔴 | Substitution masks failed embedded-font decode | Gate on `/FontFile{,2,3}` key presence across both descriptors; swallowed `.ok()` errors now logged; warn + decline when a declared program fails to extract (`8d732de0`) | unloadable-`/FontFile3` test flipped `Some→None` |
| 🔴 | Raw charcode used as CID for non-Identity CMaps | Substitution restricted to `Encoding::Identity`; non-Identity predefined CMaps recognised but skipped with a diagnostic until a charcode→CID CMap pass is wired — `/W` widths are CID-keyed too, so partial wiring can't be correct (`8d732de0`, detail in the inline thread) | `90ms-RKSJ-H` test flipped `Some→None` |
| 🟡 | Collection from font name, not `/CIDSystemInfo` | Known `/CIDSystemInfo` Ordering now overrides the name-derived collection (logged on mismatch); Identity/unknown orderings keep the name fallback (`8d732de0`) | `Ryumin-Light` + GB1 ordering test flipped `AdobeJapan1→AdobeGB1` |
| 🟡 | Returned advance omits Th | Horizontal advance returned by `render_substituted_cjk` now scales by Th per §9.4.4, matching its own `measure_text_bytes` fallback; vertical stays Th-free (`09396218`) | Tz=50% test: advance was 20.0 both ways pre-fix, now 10.0 |

Full suite (8547 tests, `rendering,cjk-render-fallback`), default-feature lib tests, clippy `-D warnings`, `--no-default-features` lib check, and rustdoc `-D warnings` are all green locally. The fixture renders still pass with margin: jo.pdf 119k and kampo.pdf 277k non-white pixels against the 2k/5k gates.

RayVR:
CI note: the single red check, `Java Lint (Spotless + SpotBugs)`, failed in its first step — Maven Central returned 403 Forbidden while fetching the Spotless plugin POM itself, before any lint ran ([job log](https://github.com/yfedoseev/pdf_oxide/actions/runs/27336808237/job/80763007735)). This PR touches no Java; a re-run of the failed job should clear it (I can't trigger re-runs from a fork). The other 178 checks are green.

yfedoseev (approved):
Approve — thanks @RayVR, all four findings are genuinely fixed (verified in code, not just description; tests run green locally). 🙏

**Blocker 1 (masked embedded-decode failure) — fixed.** Substitution now gates on `has_font_program` derived from `/FontFile{,2,3}` *key presence* across both the wrapper and descendant descriptors (`font_dict.rs:1294-1297`, merged at `:1127`), and warns + declines when a declared program key is present but extraction fails (`:1338-1344`); the previously-swallowed `.ok()` errors are now surfaced. The `...key_present_but_unextractable` test (dangling `/FontFile3`) pins `Some→None`.

**Blocker 2 (raw char-code as CID for non-Identity CMaps) — fixed.** Substitution restricted to `Encoding::Identity` (`font_dict.rs:1297`); `parse_encoding` maps only Identity-H/V there, so 90ms-RKSJ-H/GBK-EUC-H/B5pc-H fall to `Standard` and are recognized-but-skipped with a diagnostic (`:1345-1351`). The `requires_identity_cmap_encoding` test pins `Ryumin-Light-90ms-RKSJ-H → None`. The safe choice (decline until a charcode→CID pass is wired) is right — partial wiring would mis-key the CID-keyed `/W` widths too.

**Non-blocking (3)(4) — addressed.** `/CIDSystemInfo` Ordering now overrides the name-derived collection on mismatch (logged); horizontal advance scales by Th per §9.4.4 while vertical stays Th-free, matching the `measure_text_bytes` branch.

**pdf.md-aligned:** §9.4.4 tx=…×Th vs ty (no Th) (pdf.md:17442-48), §9.6.4 six-uppercase subset prefix (17852), §9.7.5.2 + Table 118 predefined CJK CMaps, and §9.7.4.2 (pdf.md:18676-81) which explicitly authorizes implementation-dependent glyph selection for a non-embedded predefined CIDFont — the substitution mechanism's spec basis.

One note for the record (no change needed): per pdf.md:19042-43 Identity-H "shall not be used with a non-embedded font", so the Ryumin-Light-Identity-H input is technically non-conforming — the PR is correctly being a lenient reader for these ubiquitous real-world files, which §9.7.4.2 permits. The bundled fallback stays opt-in behind `cjk-render-fallback`.

The lone red `Java Lint` check is a Maven Central 403 flake (PR touches no Java) — I'll re-run it.

PR #731, opened 2026-06-10
fix(color): don't abort OutputIntents scan on a broken entry; log damaged profiles (#712)

### LINKED ISSUE #712: color: output_intent_cmyk_profile silently swallows profile decode/parse errors and aborts on first bad entry
## Summary
`Document::output_intent_cmyk_profile` (`src/document.rs:3721`) collapses three distinct failure conditions into a silent `None`, making "declared-but-broken press profile" indistinguishable from "no profile requested." Pre-existing on `main`, but increasingly load-bearing as ICC retargeting / RGB→CMYK paths depend on it.

## Details
1. **First-bad-entry aborts the whole search** — `Object::Reference(r) => self.load_object(r).ok()?` (~`src/document.rs:3737`) propagates `None` out of the entire function. `[<broken-ref> <good-CMYK-profile-ref>]` loses color management because of the first entry. (The inner profile-stream load correctly `continue`s — inconsistent.)
2. **Corrupt profile stream → silent drop** — `decode_stream_data() Err => continue` (~`3764`) and `IccProfile::parse` failure (no else) drop the entry with no log.

## Impact
A prepress PDF that *declares* a CMYK OutputIntent but ships a damaged ICC profile silently renders with the K=0 RGB-inverse fallback — wrong plate values, no diagnostic.

## Suggested fix
- Change entry-load to `match … { Ok(o) => o, Err(_) => continue }` for parity with the inner load.
- `log::warn!` (naming the object) on decode failure and parse rejection. `sidecar.rs` `discover_page_spot_inks` already models this logging pattern.

mbeschastn0v:
Fixes #712.

`output_intent_cmyk_profile` returned `None` for the entire `/OutputIntents` array when a single entry's object failed to load — the per-entry `load_object(r).ok()?` propagated out of the whole function — so a valid CMYK `DestOutputProfile` declared after a broken entry was lost. A failing entry is now skipped, matching the inner profile-stream load and the §7.3.10 handling already applied to missing references (a missing ref is `Null`; this covers entries whose object exists but fails to parse).

Entries that fail to load, fail to decode, or are not a valid N=4 ICC profile (§8.6.5.5) now emit a `log::warn!` naming the object instead of being dropped silently, so a declared-but-damaged press profile is diagnosable rather than mis-rendering through the §10.3.5 fallback.

Tests in `tests/test_output_intent_resilience.rs` (default features, real PDFs via `from_bytes`):
- broken-entry regression (fails before this change)
- undecodable-stream and unparseable-ICC guards
- happy path, no-intent, RGB-only

No behaviour change on the happy or no-profile paths.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -3734,7 +3734,14 @@ impl PdfDocument {
 
         for entry in intents_arr {
             let entry = match entry {
-                Object::Reference(r) => self.load_object(r).ok()?,
+                // Skip a broken entry rather than aborting the whole array (§7.3.10).
+                Object::Reference(r) => match self.load_object(r) {
+                    Ok(o) => o,
+                    Err(e) => {
+                        log::warn!("OutputIntents entry {r} could not be loaded ({e:?}); skipping");
+                        continue;
+                    },
+                },
                 other => other,
             };
             let entry_dict = match entry.as_dict() {
@@ -3745,10 +3752,19 @@ impl PdfDocument {
                 Some(p) => p.clone(),
                 None => continue,
             };
+            let profile_label = match &profile_obj {
+                Object::Reference(r) => format!("DestOutputProfile {r}"),
+                _ => "inline DestOutputProfile".to_string(),
+            };
             let profile_stream = match profile_obj {
-                Object::Reference(r) => match self.load_object(r) {
-                    Ok(o) => o,
-                    Err(_) => continue,
+                Object::Reference(r) => {
+                    match self.load_object(r) {
+                        Ok(o) => o,
+                        Err(e) => {
+                            log::warn!("OutputIntent {profile_label} could not be loaded ({e:?}); skipping");
+                            continue;
+                        },
+                    }
                 },
                 other => other,
             };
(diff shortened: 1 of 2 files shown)


yfedoseev (approved):
Approve. Correct, well-scoped fix for a real silent-failure regression.

- The one-line behavioral change (`.ok()?` → `continue` at the per-entry load) stops a single broken `/OutputIntents` entry from aborting the whole lookup and hiding a valid later CMYK `DestOutputProfile`. Everything else is added `log::warn!` diagnostics on the four previously-silent skip points.
- pdf.md-aligned: §14.11.5 states output-intent data is informational and a conforming reader is "free to disregard it," so skipping a malformed entry is *more* spec-faithful than aborting; §7.3.10 (undefined ref → null), §8.6.5.5 (N must match) also check out.
- Ran the 6 hand-built-PDF tests locally — all pass; the regression test genuinely fails pre-fix.

Non-blocking: the pre-existing 'first N=4 profile, ignore /S subtype' simplification is unchanged and out of scope.

PR #737, opened 2026-06-14
fix(render): correct Lab/DeviceN spot colour and form-XObject compositing; byte-identical dirty-rect perf

RayVR:
## Description

Spot-colour rendering **correctness** fixes bundled with **byte-identical** render-path **performance** improvements. They were developed and validated together on a corpus of spot-colour production PDFs, using two validation modes matched to the kind of change:

- **Performance changes** — validated by **byte-identity** (output is bit-for-bit unchanged).
- **Correctness changes** — validated by **CIEDE2000 (ΔE2000)**, quantifying the perceptual colour shift toward the spec-correct result (ΔE2000 < 1 ≈ imperceptible).

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Code refactoring
- [x] Tests
- [ ] CI/CD changes

## Related Issues

None tracked upstream. Supersedes #736 (its overprint dirty-rect commit is the first performance commit here).

## Changes Made

**Correctness (intended output changes, oracle-validated):**

- CIE L\*a\*b\* colour space (§8.6.5.4): proper `Lab → XYZ → Bradford-adapt-to-D65 → linear sRGB → gamma` conversion with `/Range` clamping, replacing a grayscale approximation that rendered saturated spots near-white.
- Resolve **indirect** `/Alternate` colour-space references (e.g. `12 0 R`) before dispatch — previously they fell through to a grayscale approximation.
- Resolve single-active-ink `/DeviceN` through the `/Colorants` CIE `/Separation` definition rather than the bulk CMYK alternate, materially improving spot fidelity.
- Form XObjects (§8.10.1, §11.6.6): inherit the caller's constant alpha and blend mode; composite a `/Group << /S /Transparency >>` form as a unit. Fixes overlay content that clobbered the layer beneath instead of blending.
- Map Separation/DeviceN **image** samples through the tint transform into the alternate space (with the alternate's ICC profile, or the document `/OutputIntents` CMYK profile per §14.11.5).

**Performance (byte-identical — each bounds a per-paint full-page operation to the paint's device rect, where pixels outside are provably zero-coverage):**

- Overprint after-paint snapshot + scan bounded to the paint rect.
- Text-arm overprint snapshot bounded via a dry glyph pass.
- Sidecar mirror passes (spot/RGB/CMYK) bounded to the paint rect.
- Snapshot families (smask, cmyk-compose, spot-paint, cmyk-sidecar, smask-spot) and the fill/stroke coverage rasterisers bounded to the paint rect, reusing a scratch buffer cleared only within the dirty rect.
- Reuse the per-paint coverage *result* buffer across paints (it was zero-allocated full-page on every paint operator); now allocated at most once per page per fill/stroke lane, overwriting only the paint rect. On a large page with thousands of small vector fills this per-paint full-page memset was the dominant cost (~73% of a worst-case page); eliminating it drops that page from ~5.1s to ~1.35s.
- Precompute `IccProfile` content hash once at construction.
- Memoise `load_font_data` name resolution per rasterizer.

## Testing

- [x] I have run `cargo fmt`

Tests run under `--features rendering,icc-lcms2,test-support` (the project's required ICC backend): **8892 passed, 0 failed**. `--all-features` is not a valid combination for this crate — `Cargo.toml` requires FIPS to build without default features — so the equivalent full-feature set above was used instead. Clippy run with the same feature set and `--all-targets`.

New/strengthened tests:
- Form-XObject compositing: a `/BM /Multiply` case (lands at ~64 — half the value a dropped blend mode would give), an isolated `/Group` case (detects opaque compositing vs ~128), and a tightened constant-alpha case.
- Dirty-rect guardrails (overprint, sidecar mirror, snapshot families) assert tight upper bounds on scanned/snapshotted area while preserving the full-page fallback, so any bound that dropped a painted pixel would fail.

> Note for CI: these guardrails require `rendering` + `test-support` to execute. A bare `--features icc-lcms2` run compiles them but runs zero, which still reads as green — use the full feature set.

N/A — no Python-facing changes.

## Documentation

Code comments updated (colour-conversion math, dirty-rect invariants, 8-bpc sample assumption). No README/CHANGELOG/examples changes were required.

## Checklist

## Additional Notes

**Validation results.** Performance: byte-identical across the full corpus (0 divergent pages); the overprint pass is a multi-× corpus-aggregate win (far larger on overprint-heavy pages), and the snapshot/coverage bounding turns the worst single-page outliers from tens of seconds into low single digits. Correctness: perceptible movement of spot colours toward the spec-correct result (mean ΔE2000 ≈ 3.0 → ≈ 1.9) with **zero perceptible regressions**.

**Known pre-existing limitations (not introduced here, left as follow-ups):** no recursion guard on form `Do`; form `/BBox` not honored; non-isolated transparency groups under alpha/blend are rendered as isolated.


yfedoseev:
Thanks for this @RayVR — really nice work. I went through it across spec alignment, transparency compositing, and the byte-identity perf, and it holds up well. A few notes, mostly for a follow-up — nothing that should hold up the good stuff here.

**What checks out cleanly**
- The Lab → XYZ → Bradford-D65 → linear-sRGB → gamma path is spec-correct down to the constants (EPS=216/24389, KAPPA=24389/27, the IEC sRGB matrix + transfer with the 0.0031308 segment), with `/Range` clamping (§8.6.5.4). The previous grayscale approximation was genuinely wrong, and this is the spec-faithful fix.
- Indirect `/Alternate` resolution and single-ink `/DeviceN` via `/Colorants` are spec-aligned (§8.6.6.4/.5).
- Form-XObject / transparency-group alpha + blend inheritance looks right for the targeted cases (§8.10.1, §11.6.4.4, §11.6.6) — group composited once, no double-applied `ca`/`BM`.
- The dirty-rect byte-identity argument holds analytically: `device_paint_rect_for_path` is a conservative superset (corner-mapped AABB + 2px AA margin + Frobenius-bounded stroke outset; non-finite geometry falls back to full-page), and the text dry-pass is the identical code path with only `fill_path` gated out. Clean.

**One thing worth tightening before merge** — the Separation/DeviceN *image* expansion path in `src/extractors/images.rs`. The image is stored as `ImageData::Raw` with the still-bit-packed stream (~`images.rs:832`), but `expand_separation_to_alt` reads `raw[base + c]` as one byte per sample. So a 1/2/4/16-bpc Separation image gets decoded as garbage with no fallback, and `evaluate_tint_function` only handles tint FunctionType 2 & 4 — Type 0 (sampled) / Type 3 (stitching) silently fall back to the literal samples. Both are silent-wrong-output rather than a graceful fallback. Smallest fix: gate the expansion on `bits_per_component() == 8` (else return the legacy path + a one-time `log::warn!`), and reuse `color.rs`'s `evaluate_type2`/`evaluate_type4` instead of the hand-rolled copy — that also picks up Type 0/3 and removes the duplication. Happy to push that change if useful.

**Nice-to-haves (no rush)**
- DRY: the tint-eval + Domain/Range parsing is duplicated between `images.rs` and `color.rs:~745`.
- Hot loop: per-pixel `as_ref().expect(...)` re-borrow in the compose/overprint loops, and the DeviceN image path re-evaluates the Type-4 program per pixel (the single-ink path already LUTs it).
- The 5 tests are hand-built and solid, asserting the dirty-rect bound tightness — a couple that assert full-page pixel-equality (two overlapping paints with buffer reuse vs a fresh-alloc baseline; a rotated/fat-stroke fixture) would lock the byte-identity guarantee into the suite itself.

The disclosed limitations (non-isolated group under blend, form `/BBox` clip, recursion guard) are fine to track as follow-ups. Overall this is a solid improvement — thanks again.

RayVR:
Thanks for the thorough review. I've pushed the Separation/DeviceN image fix you flagged as the pre-merge item.

**What changed**

I hoisted the comprehensive tint-transform evaluator (FunctionType 0/2/3/4, §7.10) out of the rendering-gated page renderer into the non-rendering `functions` module, and routed both the composite `/BC` path and the image extractor through it. So the hand-rolled Type-2/4-only copy in `images.rs` is gone, and the image path now gets Type 0 (sampled) and Type 3 (stitching) for free.

One wrinkle worth noting: I initially went to reuse `color.rs`'s `evaluate_type2`/`evaluate_type4` as you suggested, but that separation path only handles Type 2/4 as well — so reusing it wouldn't have picked up 0/3. The only complete 0/2/3/4 evaluator was `evaluate_bc_tint_function` in `page_renderer.rs`, which is behind the `rendering` feature; `images.rs`'s expansion path compiles in the default render-free build, so it couldn't call it. Hoisting to `functions` (which already owns the Type 4 `Program`) resolves both: one evaluator, reachable from both builds. It uses the always-available `load_object` rather than the rendering-gated `resolve_object`, and `cargo check --lib --no-default-features` stays green.

**Guards**

- `bits_per_component() != 8` now skips expansion and keeps the raw samples (one-time `log::warn!`) instead of misreading the packed stream.
- An unsupported/malformed tint transform falls back to raw samples rather than failing the whole extraction.

**Tests** (TDD, red first):
- `separation_non_8bpc_image_skips_tint_expansion` — a 4-bpc Separation image is left untouched (previously expanded to garbage).
- `separation_type0_sampled_tint_expands` — Type 0 sampled tint now expands to the alternate space.
- `separation_unsupported_tint_falls_back_to_raw` — unsupported type → raw, no error.
- `functions`-module unit tests for the shared evaluator across Type 0/2/3/4 (Type 3 with inline subfunctions exercises the `doc = None` path).

**Still open from your notes** — I scoped these out of this change deliberately:
- The composite separation path in `color.rs:~500` still carries its own `evaluate_type2`/`evaluate_type4`; folding it onto the shared evaluator touches the `ResolvedColor`/per-plate projection, so I'd rather do that as its own change than mix it in here.
- The hot-loop items (per-pixel `as_ref().expect()` re-borrow; multi-input DeviceN re-evaluating the Type-4 program per pixel — the single-ink path still LUTs) are untouched.
- The full-page pixel-equality byte-identity tests.

Happy to take the `color.rs` consolidation next if you'd like it in this PR rather than a follow-up.

yfedoseev:
Thanks for this — the colour-conversion math (Lab → XYZ → Bradford→D65 → sRGB per §8.6.5.4, DeviceN/Colorants, and the OutputIntents CMYK fallback per §14.11.5) all checks out line-by-line against the spec, and the dirty-rect failure paths degrade safely to full-page. Nice work.

Could you rebase onto `main`? The branch is currently behind, so it needs updating before merge. A fuller review with a few follow-up items will come separately.

Appreciate the effort here! 🙏

RayVR:
Rebased onto `main` (now v0.3.65) — the branch is even with upstream, linear history, force-pushed. One conflict came up, in `page_renderer.rs`: upstream's `APC_MATERIALIZED` test counter and this branch's after-paint dirty-rect helpers were added at the same spot; resolved by keeping both. The other commits replayed cleanly.

Verified on the rebased tree: full lib suite green (5766 with `rendering`, 5594 default), `--no-default-features` builds, clippy `-D warnings` and `cargo fmt --check` clean.

yfedoseev:
Re-reviewed against the latest head (`d5878e6`). The **Rendering Tests** CI job is failing on a real compile error, not a flake — needs a one-line fix before this can merge:

```
error[E0061]: this method takes 9 arguments but 8 arguments were supplied
   --> src/rendering/text_rasterizer.rs:2400:18
note: argument #9 of type `Option<&mut PaintBounds>` is missing
note: method defined here --> src/rendering/text_rasterizer.rs:1673:8
```

This PR adds the trailing `paint_bounds: Option<&mut PaintBounds>` parameter to `render_substituted_cjk` (defined ~L1673), but the test helper call site at L2400 (the `advance_at` closure) wasn't updated and still passes 8 args. Adding `None` as the final argument after the existing trailing `None` should resolve it:

```rust
                    Transform::identity(),
                    &gs,
                    None,
                    None, // paint_bounds
                )
```

Note: this only surfaces in the `rendering,test-support` tier, which CI runs for this PR but the default `cargo test` doesn't compile. Once pushed, CI will re-run automatically. Everything else looks good from the prior review.

PR #747, opened 2026-06-16
feat(color): add pure-Rust tintbox ICC backend (icc-tintbox)

RayVR:
## Description

Adds an `icc-tintbox` color-management backend: a pure-Rust ICC CMM with the same press-grade capability surface as the existing `icc-lcms2` backend, but with no C dependency. It keeps the full color pipeline — CMYK→CMYK retargeting through the Lab PCS, black-point compensation, per-rendering-intent dispatch, and sRGB↔CMYK transforms — available on pure-Rust targets (WASM, AOT) that today fall back to qcms and degrade to the additive-clamp path.

The backend wraps [`tintbox`](https://crates.io/crates/tintbox), a from-scratch reimplementation of the Little CMS transform pipeline that is verified bit-identical to its lossless code path by differential testing.

## Type of Change

- [x] Performance improvement

## Related Issues

None tracked upstream.

## Changes Made

- Add `tintbox = "0.2.0"` as an optional dependency and an `icc-tintbox` feature; implement `TintboxBackend` against the existing `IccBackend` trait (sRGB build/convert, CMYK→CMYK retarget, sRGB→CMYK, gray/rgb/cmyk buffer + pixel paths).
- Set backend selection precedence to **tintbox > lcms2 > qcms > none**; `active_backend_name()` reports `"tintbox"` when the feature is on.
- Drive tintbox's lossless `AccurateFast` strategy. It is bit-identical to lcms2's lossless transform path; lcms2's default lossy curve optimizer is the one intentional behavioural difference, and the parity suite documents the resulting drift rather than masking it.
- Add a differential parity suite (gated on both `icc-lcms2` and `icc-tintbox`) asserting **byte-for-byte** equality across CMYK→sRGB, CMYK→CMYK, RGB→sRGB and sRGB→CMYK over dense input lattices. Tests skip with a printed notice when standard ICC profiles are absent, so CI without the profile pack stays green.
- When the `parallel` feature is on, split CMYK image buffers above a pixel threshold across rayon worker threads through the shared `Send + Sync` transform; the parallel result is bit-identical to the serial path, with a serial fallback if any slice fails.
- Extend the existing ICC feature gates in the rendering color paths to include `icc-tintbox`, and add a `tintbox` cargo-shear exemption (mirroring `lcms2`) since the dependency is feature-gated.
- Add an opt-in `icc-tintbox-simd` feature that forwards to tintbox's SIMD path — lossless, and pulls in `wide` only when explicitly enabled. Off by default.

## Testing

- [x] I have run `cargo fmt`

Verified locally:

- `cargo test --features icc-lcms2,icc-tintbox` — the tintbox/lcms2 parity suite passes (max Δ == 0 on every transform class).
- `cargo test --features icc-tintbox,parallel,rendering color::` — color + rendering color paths pass, including the parallel buffer-split.
- `cargo clippy --features icc-tintbox,parallel,rendering -- -D warnings` — clean.
- `cargo check --lib --no-default-features` — clean (no reliance on rendering/ICC-gated items from the default build).
- `cargo fmt --all -- --check` — clean.

`cargo test --all-features` was not run end-to-end locally (it pulls in the ML/OCR/ONNX stacks); the targeted feature combinations above cover the code this PR touches.

- N/A — no public API surface changes; the backend is selected at compile time by feature.

## Documentation

- [x] I have commented the new backend, the parity suite, and the feature semantics inline.
- [ ] CHANGELOG.md — left for the maintainer to curate at release time (no `[Unreleased]` section to target).

## Additional Notes

- **Default strategy is `AccurateFast`, deliberately.** It is differential-proven bit-identical to the per-pixel `Accurate` path and, after tintbox's batched-scratch fix, is never slower than `Accurate` at any chunk size and materially faster on CLUT-bound CMYK transforms — so there is no accuracy cost to it being the default. The parity suite is the regression guard that keeps it lossless.
- **SIMD is opt-in and off by default.** The win is concentrated on the matrix stage of CLUT transforms and is modest; gating it keeps `wide` out of the dependency graph unless a consumer asks for it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3069,6 +3069,7 @@ dependencies = [
  "taffy",
  "tempfile",
  "thiserror 2.0.18",
+ "tintbox",
  "tiny-skia",
  "tokenizers",
  "tract-onnx",
@@ -4639,6 +4640,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "tintbox"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5f689ac8b49271264a1435c179cad6ac97e43ec7017b456059021e14efd4cd16"
+dependencies = [
+ "wide",
+]
+
+[[package]]
 name = "tiny-skia"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5292,9 +5302,9 @@ checksum = "d4ca08e5ef825b65b056d9efbd95c8750683f0a6d0466d02e96dc2e4e360f3d2"
 
 [[package]]
 name = "wide"
-version = "1.4.0"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a7714cd0430a663154667c74da5d09325c2387695bee18b3f7f72825aa3693a"
+checksum = "dfdfe6a32973f2d1b268b8895845a8a96cac2f0191e72c27cc929036060dbf89"
 dependencies = [
  "bytemuck",
  "safe_arch",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,9 @@ ignored = [
     # WASM-only runtime deps (gated on `wasm` feature)
     "getrandom_02",
     "web-sys",
-    # Press-grade ICC backend (gated on `icc-lcms2` feature)
+    # Press-grade ICC backends (gated on `icc-lcms2` / `icc-tintbox` features)
     "lcms2",
+    "tintbox",
 ]
 ignored-paths = [
     # Runnable examples — not declared as [[example]] targets because they use
@@ -291,6 +292,7 @@ office_oxide = { version = "0.1.2", default-features = false }
 x509-tsp = { version = "0.1", optional = true }
 ureq = { version = "3", default-features = false, features = ["rustls"], optional = true }
 cmpv2 = { version = "0.2", optional = true }
+tintbox = { version = "0.2.0", optional = true }
 
 # `getrandom`'s `wasm_js` backend is only meaningful when targeting
 # `wasm32-unknown-unknown` (browser / WASI runtimes that route entropy
@@ -363,7 +365,7 @@ fips = ["dep:aws-lc-rs", "aws-lc-rs/fips"]
 
 # Real ICC colour management.
 #
-# Three modes:
+# Four modes:
 #
 #  - `icc-qcms` (default, via the `icc` alias): pure-Rust Firefox-derived
 #    colour engine. CMYK→RGB through the embedded profile, no CMYK→CMYK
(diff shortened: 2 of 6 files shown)


RayVR:
## Performance analysis

Measured on a local merge of this branch with #737 — they compose cleanly (no conflicts), and merging is what lets tintbox actually drive the Separation/DeviceN/Lab rendering paths #737 adds. 12-core machine, release profile (fat LTO).

### CMM kernel throughput

2,000,000 px × 20 iters, MPx/s (higher = faster):

| Transform | tintbox AccurateFast | tintbox Accurate | lcms2 lossless | lcms2 default (lossy) |
|---|--:|--:|--:|--:|
| CMYK→sRGB | **5.89** | 4.36 | 5.70 | 43.48 |
| CMYK→CMYK | **4.57** | 2.99 | 3.40 | 37.16 |

On a like-for-like **lossless** basis tintbox is the fastest — it edges out lcms2's own lossless path. lcms2's ~43/37 MPx/s is its **lossy** default CLUT optimizer (the source of the colour drift the parity suite documents), not a lossless comparison.

### End-to-end render

Representative CMYK page (Separation/DeviceN spot inks + an embedded CMYK image), single page, median ms:

| DPI | lcms2 (lossy default) | tintbox serial (lossless) | tintbox parallel (lossless) | parallel speedup |
|---:|--:|--:|--:|--:|
| 150 | 377 | 1417 | **344** | 4.12× |
| 300 | 612 | 1649 | **578** | 2.85× |
| 600 | 1398 | 2466 | **1399** | 1.76× |
| 1200 | 4507 | 5648 | **4513** | 1.25× |

### Reading

- **Parallel tintbox ≈ lcms2's lossy default at every DPI**, and faster at low DPI — i.e. fully lossless, accurate colour at the speed of the lossy C backend, in pure Rust.
- **Serial tintbox carries a near-constant ~1.0–1.1 s CMM overhead** on this page regardless of DPI (1417−377 ≈ 1040 ms @150; 5648−4507 ≈ 1141 ms @1200). That constant absolute cost is the tell that CMYK→RGB is **fixed at the source-image resolution** — it doesn't grow with output DPI. The `parallel`-gated buffer split recovers it.
- **Parallel speedup decays 4.12× → 1.25× as DPI climbs.** Resample (bicubic) + PNG encode are serial, colour-agnostic, and scale with *output* pixels, so they increasingly dominate; at high DPI all three configs converge on that same floor (@1200: lcms2 4507 vs parallel 4513, within noise). So the CMM-side parallelism is a low-to-mid-DPI win and effectively a no-op at high DPI — where the bottleneck is resample + encode.

The design question of where parallelism should live (CMM buffer split vs resample/encode vs page-level) is in #749.

yfedoseev:
## We can't accept the `tintbox` dependency as it currently stands

Thanks for the work on the pure-Rust CMM — the direction is appealing. But we can't merge a dependency on `tintbox` 0.2.0 in its current form, and I want to be upfront about why.

**The blocker is supply-chain risk, not the code quality:**

- **~177 total downloads, first published 2026-06-15** (2 days ago), 2 versions both released this week — zero real-world exposure.
- **Sole maintainer is the PR author**, published from a personal crates.io account (`github.com/RayVR/tintbox`). That's bus-factor 1: nobody independent has reviewed, used, or fuzzed it, and a self-authored dependency carries the same review burden as the PR itself.
- **It parses untrusted input on a security-sensitive surface**: a from-scratch reimplementation of lcms2 that decodes ICC profiles embedded in arbitrary PDFs. That surface has produced multiple CVEs in lcms2 over the years. A brand-new parser has had no fuzzing in the wild, and the "bit-identical to lcms2 / differentially tested" claim is currently self-attested with no third-party verification.

For a library that other projects depend on to process untrusted documents, taking a days-old, single-author, ~100-download crate as the engine for untrusted-input parsing isn't something we can sign off on, even gated behind an opt-in feature.

**What would make this acceptable:**

1. **Vendor the source into this repo** (path dependency or pinned submodule) so the color-management code is reviewed, fuzzed, and maintained here under the project — not pulled from a personal crates.io namespace; **or**
2. Let `tintbox` **mature independently** — co-owned under the project org, exact-pinned (`=0.2.0`, not `^0.2.0`), with a fuzz target on the ICC-parsing path and meaningful third-party adoption before we depend on it.

Until one of those is in place, `icc-tintbox` can't land. If you'd like to go the vendoring route, I'm happy to help set up a fuzz harness for the ICC parser. (Separately, the `Rendering Tests` compile error noted above still needs the one-line fix.)

yfedoseev:
Want to add some context to the above, because I don't want the ask to land the wrong way.

To be clear: **this isn't about trust in you.** Your work across the rendering and colour stack — separations, overprint, OutputIntents, the CMYK pipeline, and now a from-scratch CMM — has been consistently expert-level and carefully done, and our confidence in your judgement here is high. If this were a question of whether *you* know colour management, there'd be nothing to discuss.

The concern is purely **structural, and it's a policy we'd apply to any author — including maintainers**: the project can't take a hard dependency on a crate that is days old, single-maintainer, and barely downloaded when its job is parsing untrusted input (ICC profiles out of arbitrary PDFs). That's not a statement about the code's quality — it's about **bus factor, downstream consumers who inherit the dependency, and what happens to this project if the crate ever goes unmaintained or out of sync.** Those risks exist regardless of how good the author is.

And honestly, the fix also protects *you*: vendoring `tintbox` into this repo as a workspace sub-crate (MIT, your attribution) means the colour engine is reviewed and maintained *with* the project rather than resting on one person's solo release cadence. That's the outcome we'd love — your expertise, in-tree, under shared maintenance — and it removes the supply-chain objection entirely. `icc-tintbox` stays opt-in either way.

So: no reservation about you, full reservation about depending on a brand-new external crate. Happy to help with the move and a fuzz target for the parser whenever you're ready.

Issue #724, opened 2026-06-09
[Bug]: Inter-word spaces dropped on tightly-typeset PDFs (glyph-per-Tj + Td gaps, no space glyph)


regularkevvv:
## Bug Description

When a PDF draws text one glyph per `Tj` with incremental `Td` moves and **no explicit space glyph** — inter-word gaps are just slightly larger `Td` offsets (~0.3–0.8 × the glyph advance) — `to_plain_text` concatenates words:

- `"Master of Science"` → `"MasterofScience"`

The characters are correct; only the **word boundaries** are lost. This is a spacing/segmentation issue, not character corruption. It hurts search/indexing (you can't match `science` inside `masterofscience`). Many résumé generators typeset exactly this way, so it shows up broadly on real-world résumés.

PyMuPDF and poppler `pdftotext` both infer these spaces correctly, so oxide is the outlier against its own calibration reference.

## Steps to Reproduce

1. Take a PDF that positions words via `Td` gaps with no space glyph and tight inter-word spacing. A 100%-fabricated reproduction (no PII), generated with reportlab, is here:
   - PDF: https://github.com/regularkevvv/pdf_oxide/blob/3086f9ebe76cb545e8ffad9d4d2cb17d846c1cf7/tests/fixtures/word_spacing/resume_tight_spacing.pdf
   - generator (provenance): https://github.com/regularkevvv/pdf_oxide/blob/3086f9ebe76cb545e8ffad9d4d2cb17d846c1cf7/tests/fixtures/word_spacing/generate_resume.py
2. Extract page 0 with `to_plain_text`.
3. Compare against `pdftotext` / PyMuPDF on the same file.

```text
poppler pdftotext : "... Results driven software engineer with expertise in distributed data systems, ... Master of Science in Information Systems ..."   (105 word tokens)
pdf_oxide (before) : "... Resultsdrivensoftwareengineerwithexpertiseindistributeddatasystems, ... Masterof Sciencein Information Systems ..."                       (45 word tokens)
```

## Expected Behavior

Tight inter-word gaps that PyMuPDF / poppler treat as word breaks should be recovered as spaces, e.g. `Results driven software engineer ... Master of Science in Information Systems`.

## Actual Behavior

Lowercase→lowercase word runs are glued together; only boundaries adjacent to an uppercase letter / digit / punctuation get a space. On the fixture this drops the page from 105 word tokens to 45.

## PDF Information (if applicable)

[...]

Issue #487, opened 2026-05-09
fix(html): to_html produces lower Jaccard scores than to_markdown for table-heavy PDFs


yfedoseev:
## Problem

`to_html` scores consistently lower than `to_markdown` and `extract_text` when comparing against kreuzberg ground-truth files for PDFs with complex tables (`pdfa_036`, `nougat_026`, `nougat_040`, `nougat_018`).

## Context

All four PDFs below were introduced as quality-gate fixtures in #484. Their download URLs and ground-truth files are listed in `tests/test_corpus_extraction_quality.rs`. All text-extraction quality gates pass; the failures are specific to the HTML output path.

## Quality gate evidence (kreuzberg GT, Jaccard similarity)

| PDF | threshold | oxide_text | oxide_md | oxide_html | pdftotext | pdfium |
|---|---|---|---|---|---|---|
| pdfa_036 (fixture: #484) | 0.78 | P 0.943 | P 0.937 | **F 0.715** | P 0.986 | P 0.986 |
| nougat_026 (fixture: #484) | 0.87 | P 0.874 | P 0.882 | **F 0.813** | P 0.929 | P 0.926 |
| nougat_040 (fixture: #484) | 0.35 | P 0.398 | P 0.444 | **F 0.324** | P 0.461 | P 0.462 |
| nougat_018 (fixture: #484) | 0.95 | P 0.952 | F 0.639 | **F 0.884** | P 0.964 | P 0.975 |

## Suspected causes

1. **HTML table cell wrapping** — table cells are wrapped in `<td>` / `<th>` tags; when stripped for text comparison, surrounding whitespace collapses tokens differently than in the markdown pipe-table format.
2. **Escaped entities** — special characters are encoded as HTML entities (`&quot;`, `&amp;`, `&#x2019;`). After `html.unescape()` the tokens should match, but some entity sequences may not round-trip cleanly.
3. **Structural tokens in GT** — the kreuzberg GT (Nougat ML output) contains markdown-specific tokens (`---`, `##`) that match the markdown path but not the HTML path. Some score gap is inherent to the GT format.
4. **Table border/header rows** — HTML renders `<thead>` rows that may duplicate header text, inflating token counts and reducing Jaccard.

## Fix direction

- Audit `src/converters/html.rs` table cell text assembly for spurious whitespace insertion.
- Compare the token set produced by `to_html` vs `to_markdown` for the same table to locate the specific cells that differ.
[...]

PR #759, opened 2026-06-22
fix(text): preserve same-row span order when grouping splits a line

### LINKED ISSUE #752: [Bug]: `to_plain_text` reorders and splits same-row spans, corrupting form-field values
### Bug Description

When one logical line of text is extracted as several spans at the **same Y** in **different reading-order groups**, with adjacent spans **overlapping** by a fraction of a point, `to_plain_text` both reorders and splits the line — a value comes out with a fragment hoisted to the front:

```
ALPHA BETA GAMMA DELTA EPSILON   ->   A EPSILON ALPHA BETA GAMMA DELT
```

`pdftotext` reads the same file correctly. Every glyph is present — this is a reading-order/spacing defect, not character corruption.

### Steps to Reproduce

1. Open the attached [repro_cid.pdf](https://github.com/user-attachments/files/29140301/repro_cid.pdf).
2. Extract page 0 with `to_plain_text` (see `Code Sample`).
3. The value line `ALPHA BETA GAMMA DELTA EPSILON` comes out reordered and split.

### Expected Behavior

```
Header ALPHA BETA GAMMA DELTA EPSILON RIGHTVAL
```

### Actual Behavior

```
Header
A EPSILON ALPHA BETA GAMMA DELT RIGHTVAL
```

`A EPSILON` (left edge x=191) is emitted ahead of `ALPHA BETA GAMMA DELT` (left edge x=35), even though both are on the same row and the latter has the smaller X.

- Single A4 page, unencrypted, ~90 KB, untagged
- Font: **embedded CID TrueType, Identity-H** encoding, with `/ToUnicode`
- The value line is drawn as a `Td`-chain of runs; pdf_oxide extracts them as same-Y spans in different reading-order groups whose boxes overlap by ~0.5pt
- `pdftotext` extracts the page correctly

### Environment

- **OS**: Ubuntu 22.04
- **Rust version**: rustc 1.92.0 (ded5c06cf 2025-12-08)
- **pdf_oxide version**: 0.3.66 (reproduces on current `main`)
- **Python version** (if using Python bindings): 3.x (also reproduces via the Rust API)

### Code Sample

```rust
use pdf_oxide::{converters::ConversionOptions, PdfDocument};

let doc = PdfDocument::open("repro_cid.pdf").unwrap();
let txt = doc.to_plain_text(0, &ConversionOptions::default()).unwrap();
println!("{txt}");
// actual (0.3.66): "Header\nA EPSILON ALPHA BETA GAMMA DELT RIGHTVAL\n..."
// expected:        "Header ALPHA BETA GAMMA DELTA EPSILON RIGHTVAL\n..."
```

### Error Messages / Logs

```shell
No error or panic — extraction succeeds and returns reordered/split text.
```

### Checklist

schelip:
## Description

`to_plain_text` orders spans by reading-order group, then `reorder_columnar_groups`
interleaves side-by-side column groups row-by-row and `merge_same_y_across_groups`
joins same-Y spans from different groups onto one line. When a producer draws one
logical line as a `Td`-chain whose runs land in **different reading-order groups at the
same Y with boxes overlapping by ~0.5pt**, two failures compound:

1. the overlapping group (≥2 spans sharing a Y) is mistaken for a 2-row vertical column
   and interleaved, hoisting a fragment to the front of the line; and
2. the `cross_group_same_y_space` rule forces a space between the overlapping fragments,
   splitting a word.

Net: `ALPHA BETA GAMMA DELTA EPSILON` → `A EPSILON ALPHA BETA GAMMA DELT`. Every glyph is
present — a reading-order/spacing defect, not character corruption.

Text show operators position glyphs purely geometrically; a `Td` offset encodes no space
character (ISO 32000-1:2008 §9.4.3), so adjacency must be inferred from box geometry — and
overlapping boxes mean one continuous run, never a column boundary or a label/value gap.

### Approach

- `reorder_columnar_groups` — a column candidate must occupy distinct Y rows (one cell per
  row). New `run_has_distinct_rows` helper; a group whose spans share a Y is excluded from
  columnar detection at both the base-group and column-extension checks.
- render spacing — remove `cross_group_same_y_space`. It bypassed `has_horizontal_gap` to
  always space same-Y cross-group pairs; since #487 removed that helper's upper gap bound,
  genuine label/value gaps are already covered by `needs_gap_space`, so the branch was
  redundant and, on overlapping fragments, wrong. Plain-text now uses the same gap
  threshold as the Markdown/HTML converters.
- orphan recovery — same-line orphan fragments are spaced only on a real horizontal gap
  (`has_horizontal_gap`), mirroring the main render path.

Self-gating: clean columns (distinct Ys) and genuine label/value pairs (positive gap) are
unchanged; only the ambiguous overlap case is touched.

## Type of Change

## Related Issues

Fixes #752 · Complements #724 (inter-word spaces on tightly-typeset PDFs) and #487 (the
`has_horizontal_gap` upper-bound removal this builds on)

## Changes Made

- `src/pipeline/converters/plain_text.rs` — (1) new `run_has_distinct_rows` + guards in
  `reorder_columnar_groups` (base group and column-extension) so a same-Y group is not a
  column candidate; (2) remove the `cross_group_same_y_space` branch from the per-span
  spacing decision; (3) orphan-line builder spaces fragments only on a real
  `has_horizontal_gap`. Adds four in-module regression tests:
  `test_split_line_not_misdetected_as_column`,
  `test_overlapping_cross_group_spans_no_spurious_space`,
  `test_orphan_overlapping_fragments_no_spurious_space`,
  `test_orphan_gapped_fragments_keep_space`.
- `CHANGELOG.md` — `[Unreleased]` entry.

## Testing

- [x] I have added tests that prove the fix (all fail on `main`, pass here; each pins a
      distinct hunk — columnar reorder, render spacing, orphan path).
- [x] All new and existing tests pass locally (`cargo test --lib` green; the columnar /
      two-column / converter-structure / extraction-API integration suites green).
- [ ] I have run `cargo test --all-features` (enables `fips`, mutually exclusive with the
      default `legacy-crypto` — `compile_error!`; CI doesn't run it either).
- [x] I have run `cargo clippy --all-targets -- -D warnings`
- [x] I have run `cargo fmt`

N/A — core Rust converter change, no binding API change; flows through to all bindings.

## Documentation

- [ ] Examples (not applicable)

## Checklist

## Additional Notes

Regression tests build spans in code (synthetic, in-memory — no binary fixtures). The
issue's reproducer extracts `Header ALPHA BETA GAMMA DELTA EPSILON RIGHTVAL` as expected.

Tradeoff: removing `cross_group_same_y_space` makes the plain-text converter apply the same
`has_horizontal_gap` threshold as the Markdown/HTML converters — a cross-group same-Y pair
whose boxes touch or overlap (gap ≤ 0.15·font) no longer gets a forced space. This is the
consistent, geometry-based behavior and the safe default for the reported corruption.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to PDFOxide are documented here.
 
+## [Unreleased]
+
+### Fixed
+
+- **Same-row spans no longer reordered or split in plain-text output (#752)** — when a producer emits one logical line as several spans at the same Y in different reading-order groups whose boxes overlap by a fraction of a point, `to_plain_text` interleaved the overlapping group as if it were a vertical column (hoisting a fragment to the front) and forced a space between the overlapping fragments (splitting a word). A group whose spans share a Y row is now excluded from columnar detection, and the cross-group same-Y space special case is removed in favour of the standard `has_horizontal_gap` threshold the other converters already use — so overlapping fragments join. Fixes e.g. `ALPHA BETA GAMMA DELTA EPSILON` extracted as `A EPSILON ALPHA BETA GAMMA DELT`.
+
 ## [0.3.67] - 2026-06-20
 
 > Reading-order quality release for untagged scientific papers — manuscript-line-number rails lifted out of the body, dense two-column bodies kept apart at a measured gutter, comma-bearing statistic subscripts rejoined, repeated journal pagination footers suppressed, whole-page placed-PDF article bodies recovered, and single-column pages protected from false column detection.
--- a/src/pipeline/converters/plain_text.rs
+++ b/src/pipeline/converters/plain_text.rs
@@ -72,6 +72,23 @@ impl PlainTextConverter {
         table.render_text()
     }
 
+    /// Whether a run of spans occupies distinct vertical rows — a real table
+    /// column has one cell per row. A group whose spans share a Y row (e.g. a
+    /// name split into adjacent same-line fragments) is a broken-up line, not a
+    /// column; treating it as one makes the row-interleave hoist same-row spans
+    /// out of left-to-right order.
+    fn run_has_distinct_rows(spans: &[&OrderedTextSpan]) -> bool {
+        let mut ys: Vec<f32> = spans.iter().map(|s| s.span.bbox.y).collect();
+        ys.sort_by(|a, b| crate::utils::safe_float_cmp(*a, *b));
+        let tol = spans
+            .iter()
+            .map(|s| s.span.font_size)
+            .fold(0.0_f32, f32::max)
+            .max(1.0)
+            * 0.5;
+        ys.windows(2).all(|w| (w[1] - w[0]).abs() >= tol)
+    }
+
     /// Detect consecutive groups that form a columnar layout and reorder
     /// their spans row-by-row instead of column-by-column.
     ///
@@ -126,8 +143,12 @@ impl PlainTextConverter {
 
         while i < runs.len() {
             let base_count = runs[i].2;
-            // Need at least 2 rows per column (header + data) and at least 2 columns.
-            if base_count < 2 {
+            // Need at least 2 rows per column (header + data) and at least 2
+            // columns. A column must also occupy distinct Y rows (one cell per
+            // row); a group whose spans share a row is a split line, not a column.
+            if base_count < 2
+                || !Self::run_has_distinct_rows(&sorted[runs[i].1..runs[i].1 + runs[i].2])
+            {
                 i += 1;
                 continue;
             }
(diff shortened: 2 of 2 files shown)


yfedoseev (approved):
Code and spec alignment look good. `run_has_distinct_rows` correctly gates columnar detection per §9.4.3 (Td encodes no space — adjacency is geometric); removal of `cross_group_same_y_space` is sound since `has_horizontal_gap` already covers label/value gaps. Orphan recovery now consistently uses the same threshold. Tests cover both paths. CI build failures are transient registry fetches; security advisory (RUSTSEC-2026-0186/memmap2) is pre-existing on main and will be cleared with the upcoming v0.3.68 merge.

Issue #773, opened 2026-06-24
[Bug]: Embedded Type0 fonts decode CIDs missing from `/ToUnicode` to wrong, per-file-varying characters


schelip:
### Bug Description

For embedded **Type0 / Identity-H** CID fonts, any glyph whose CID is **not
covered by the `/ToUnicode` CMap** is extracted as the **CID read directly as a
Latin-1 character** rather than recognised as having no Unicode mapping. The
result is a plausible-but-wrong character that **varies from file to file**,
because each subset assigns glyphs to different CID slots.

This is not specific to one glyph — it affects every CID absent from
`/ToUnicode` whose value happens to fall in a printable range. In the documents
examined, the uncovered CIDs were ligature glyphs (which have no single Unicode
codepoint): e.g. a `ti` ligature comes out as a stray character, so the word
`notificacao` extracts as `no:ficacao`.

### Steps to Reproduce

A self-contained reproducer is attached — `type0_tounicode_gap.pdf` (~1.6 KB, no
real data). It draws the word `notificacao` whose `ti` ligature is a single glyph
at CID `0x003A`, deliberately omitted from `/ToUnicode`.

1. Extract page 0 text (`to_plain_text` / `extract_text`).
2. The uncovered CID (`ti`) comes out as a stray `:`.

### Expected Behavior

The `ti` glyph has no Unicode mapping anywhere in the file (no `/ToUnicode` entry,
no `post` name, no GSUB), so its letters are not recoverable. The key point is that
it must **not** be replaced with a wrong, content-like character. It should be
treated as unmapped — which way depends on the unmapped-glyph handling:

```
noficacao          # default (unmapped glyphs dropped): the "ti" glyph simply omitted
no�ficacao         # only if unmapped-glyph preservation is enabled: marked U+FFFD
```

i.e. the `�` appears only when dropping of unmapped glyphs is turned off; by
default the glyph is dropped and no replacement char is emitted.

### Actual Behavior

The uncovered CID is emitted as itself-read-as-a-character:

```
no:ficacao         # CID 0x3A → ':'
```

The substituted character is the CID slot value, so it **varies per file** — in
a 98-file corpus it was one of `:` `;` `<` `=` `B` `C` `D` `E` `F` `G` `H` `K`
`M` `N` (0x3A–0x4E) for the same source word, never the correct letters.

### PDF Information (if applicable)

- Affected font: **embedded CIDFontType2**, `Encoding /Identity-H`, Symbolic.
[...]

PR #774, opened 2026-06-24
fix(fonts): emit U+FFFD for Type0 CIDs missing from /ToUnicode instead of guessing

schelip:
## Description

For embedded **Type0 / Identity-H** fonts, any CID **not covered by the
`/ToUnicode` CMap** was decoded by treating a number as a Unicode value — the GID
via the standard glyph-name table → AGL, or the CID itself (`char::from_u32`).
Because the CID is an arbitrary subset slot, this produced a plausible-but-wrong,
content-like character that **varies per file** (e.g. a `ti` ligature glyph → `:` /
`D` / `;` …), so `notificacao` extracts as `no:ficacao`. The glyph has no Unicode
identity anywhere in the file (no `/ToUnicode` entry, no `post` name, no GSUB), so
the letters are genuinely unrecoverable by any tool — but emitting a wrong,
per-file-varying character is silent corruption. (`pdftotext` drops it to a space;
PyMuPDF substitutes the same `:`, so the wrong-char behaviour is not unique to
pdf_oxide — but `U+FFFD` is the honest result.)

The fix gates the two numeric *guess* fallbacks on the font having **no usable
`/ToUnicode`**. When a `/ToUnicode` is present but a CID is absent from it, the
glyph is genuinely unmapped and decodes to `U+FFFD` (detectable) instead of being
guessed. It is purely subtractive: a font with no `/ToUnicode` still uses the
CID-as-Unicode heuristic exactly as before, and the authoritative embedded-`cmap`
reverse and `post`/charset glyph-name lookups (which run first) are unchanged — so
only the previously-corrupting path changes behaviour.

(`U+FFFD` is dropped from the high-level accessors by default via
`PRESERVE_UNMAPPED_GLYPHS`, so the visible result is that the unmapped glyph is
omitted; enabling unmapped-glyph preservation surfaces it as `�`. Either way the
wrong character is gone.)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Refs #773

## Changes Made

- `src/fonts/font_dict.rs`: in the `Encoding::Identity` Type0 decode arm
  (`char_to_unicode_uncached`), gate the two numeric fallbacks — gid → standard
  glyph name → AGL, and CID-as-Unicode (`char::from_u32`) — on
  `self.to_unicode` having no usable CMap. When a `/ToUnicode` is present but the
  CID is absent from it, return `U+FFFD` instead of guessing. The authoritative
  embedded-`cmap` reverse and `post`/charset glyph-name lookups, which run first,
  are unchanged.
- `src/fonts/font_dict.rs` (tests): inline `#[cfg(test)]` regression alongside the
  existing Fix A / Fix B tests — `make_type0_font` with a present-but-incomplete
  `/ToUnicode` asserts `char_to_unicode` returns `U+FFFD` for two uncovered CIDs
  (`0x0100`, the CID-as-char path; `0x003A` = `colon`, the gid→glyph-name path),
  and a sibling test confirms a font *without* `/ToUnicode` still uses the
  CID-as-Unicode heuristic (no regression).
- `CHANGELOG.md`: `[Unreleased] → Fixed` entry.

## Testing

- [ ] I have run `cargo test --all-features` *(N/A for this crate: `--all-features` enables `fips`, which is a compile-time error alongside the default `legacy-crypto` and needs a CMake/C toolchain to build `aws-lc-fips-sys`; CI uses default-feature `cargo test`)*
- [x] I have run `cargo fmt`

- [ ] Python bindings updated (if needed) — N/A, internal decode change, no API surface change
- [ ] Python tests pass

## Documentation

- [ ] I have updated the documentation (README, docs/, code comments) — rationale comment added at the change site

## Checklist


yfedoseev (approved):
The fix is correct and well-targeted — thank you @schelip!

**Spec compliance** ✅  
ISO 32000-1 §9.10.2 leaves the result undefined for CIDs absent from `/ToUnicode`; U+FFFD is exactly the right output. The no-`/ToUnicode` heuristic path is correctly left unchanged.

**Two minor nits (non-blocking)**

1. **DRY**: `!has_usable_tounicode` appears twice (AGL path + CID-as-Unicode path). Both guards could be merged into a single wrapper block to avoid repeating the condition.

2. **Issue refs in test comments**: Our project policy is no `#NNN` references in code (they belong in commit messages and PR descriptions, where they don't rot). The two test doc-comments `/// #773:` and `/// #773 (no regression):` should describe the *behaviour* being tested rather than referencing the issue number. Happy to fix these up on merge if you'd prefer not to push another commit.

Tests use synthetic in-code builders (no committed third-party PDFs) ✅

schelip:
Thanks for the review @yfedoseev! Appreciate all your work in the lib. I'll get started on the fixes

yfedoseev (changes requested):
There is a real test regression that needs a fix before this can merge.

**Failing test**: `test_cff_font_with_embedded_tounicode` (tests/test_cff_font_support.rs:511)

```
assertion `left == right` failed
  left:  Some("") ← U+FFFD (new behaviour)
 right: Some(" ") ← space (expected)
```

**Root cause**

`has_usable_tounicode` gates *both* fallback paths (AGL and CID-as-Unicode). That is correct for the AGL path (always a guess), but **CID-as-Unicode is definitional for Identity-H encoding** — it is not a guess. The PDF spec (ISO 32000-1 §9.7.6) defines Identity ordering as CID = Unicode codepoint. When a ToUnicode CMap exists but does not cover a CID, the Identity fallback is still the authoritative mapping for that CID.

The test font has `ordering: "Identity"` and a ToUnicode that maps only one character. CID 0x0020 (space) is not in the ToUnicode, but CID = 0x0020 = U+0020 (space) is correct by Identity definition.

Felipe's own new test (`test_type0_tounicode_gap_returns_fffd_not_guess`) uses `ordering: "UCS"` — the fix is correct for non-Identity fonts but too broad for Identity ones.

**Suggested fix**

Add an Identity-ordering guard around the CID-as-Unicode fallback (the AGL fallback can stay gated on `!has_usable_tounicode` — that IS a guess):

```rust
let is_identity_ordered = self.cid_system_info.as_ref()
    .map(|csi| csi.ordering == "Identity")
    .unwrap_or(false);

// AGL fallback: always a guess → block when ToUnicode is present
if !has_usable_tounicode {
    // ... existing AGL code ...
}

// CID-as-Unicode: definitional for Identity-H, heuristic for everything else
if !has_usable_tounicode || is_identity_ordered {
    if let Some(unicode_char) = char::from_u32(char_code) {
        if !unicode_char.is_control() || unicode_char == ' ' {
            return Some(unicode_char.to_string());
        }
    }
}
```

This preserves the #773 fix for non-Identity fonts while restoring correct behaviour for Identity-H fonts. The `test_type0_tounicode_gap_returns_fffd_not_guess` test passes because it uses `ordering: "UCS"`.

Note: the #773 ligature case (`ti` → `:`) for Identity-H fonts is a harder problem — the CID *is* 0x003A by the font's own assignment, and without decomposition information there's no clean way to distinguish a private ligature slot from a genuine colon. That could be a follow-up. This PR's fix is a clear improvement for non-Identity fonts either way.

yfedoseev:
One more housekeeping item: could you remove the `This matches MuPDF behavior` reference from the comment on the CID-as-Unicode fallback? We avoid naming competing or third-party products in code comments — the comment already explains the *why* well enough without the attribution.

Just drop that clause and the comment reads cleanly on its own.

yfedoseev (approved):
Re-review: the CID-as-Unicode regression I flagged is fixed — `is_identity_ordered` now guards the CID-as-Unicode fallback so Identity-ordered fonts (incl. the space-glyph case in `test_cff_font_with_embedded_tounicode`) keep their mapping, while non-Identity fonts correctly emit U+FFFD. Both earlier nits addressed (single merged guard; no issue refs in test comments). Verified locally: `test_cff_font_with_embedded_tounicode`, `test_type0_tounicode_gap_returns_fffd_not_guess`, `test_type0_no_tounicode_keeps_cid_as_unicode`, and the full `test_optimized_cmap_parsing` suite all pass.

Note for tracking: this resolves the **non-Identity-ordered** case. The Identity-ordered reproducer from the linked bug still decodes via CID-as-Unicode by design, so that issue should stay open and be handled as the proper default (not an opt-in flag) in a follow-up.

yfedoseev:
## Regression sweep — clean ✅

Ran our version-over-version corpus regression sweep (process: `docs/releases/process/corpus-regression-sweep.md`) on this PR vs the current baseline (v0.3.68).

- **Corpus:** 380 real-world PDFs, deterministically sampled and weighted toward font/CID/encoding-heavy collections (pdf.js, PDFBox, olmOCR-math, curated multilingual/academic) with a broad no-regression tail.
- **Method:** extract `text` / `md` / `html` on both versions, byte-diff every pair, human-classify each change.

**Results (1,128 doc×format pairs):**

| Metric | Result |
|---|---|
| Byte-identical | **1,125 / 1,128** |
| Changed | 3 pairs = **1 document** |
| New U+FFFD anywhere | **0** |
| Regressions | **0** |
| Table guard (`google_doc_document.pdf` md) | byte-identical ✅ |

The single changed document is an **improvement**, not a regression — `PDFBOX-4322-Empty-ToUnicode` (all three formats):

```
baseline:  J)'(i#      ← wrong AGL glyph-name guess
this PR:   Justin      ← correct
```

Gating the AGL fallback behind "no usable /ToUnicode" lets the correct CID-as-Unicode mapping win on this empty-ToUnicode font instead of a wrong glyph-name guess.

Notably the **U+FFFD delta is zero** across the whole corpus — the non-Identity-ordered + gappy-`/ToUnicode` surface this PR touches is genuinely rare in the wild, so the blast radius is tiny and strictly positive here. Good to merge once CI is green.

yfedoseev:
Confirming final state ahead of merge:

- **Review:** approved (spec-compliant — ISO 32000-1 §9.10.2 leaves CIDs absent from `/ToUnicode` undefined, so `U+FFFD` is the honest output; the earlier `test_cff_font_with_embedded_tounicode` regression was fixed via the `is_identity_ordered` guard).
- **CI:** all checks green (180 pass / 0 fail).
- **Regression sweep:** clean — 380-PDF font/CID/encoding-weighted corpus, 0 regressions, no `U+FFFD`-delta noise; verified again on a large local PDF subset.
- **Housekeeping:** the `This matches MuPDF behavior` third-party reference was dropped from the comment as requested.

Squash-merging into `main` now. Thanks again @schelip — the follow-up Identity-ordered case is tracked in #776.

Issue #775, opened 2026-06-24
[Feature]: Opt-in flag to emit U+FFFD instead of CID-as-Unicode for Identity-ordered Type0 fonts


schelip:
## Problem Statement

For embedded Type0 fonts whose descendant CIDFont declares `CIDSystemInfo` **`Ordering` = `Identity`** (Adobe-Identity-0), a glyph whose CID is **not covered by the `/ToUnicode` CMap** is decoded as the **CID read directly as a Unicode codepoint**. For slots holding a glyph with no single Unicode value — ligatures, private-use forms — this emits a plausible-but-wrong, content-like character that **varies from file to file** (e.g. a `ti` ligature glyph at CID `0x003A` → `:`, so `notificacao` extracts as `no:ficacao`).

The fix proposed in #774 routes uncovered CIDs to `U+FFFD` for **non-Identity-ordered** fonts but intentionally retains the CID-as-Unicode mapping for Identity ordering (matching MuPDF), leaving this case as a known limitation. ISO 32000-1 §9.10.2 leaves the result for these CIDs undefined ("a conforming reader may choose a character code of their choosing"), so the behaviour is spec-conformant — it is just not the right choice for every consumer.

The affected glyphs can be genuinely unrecoverable: when the font carries no embedded `cmap`, `post` glyph names, GSUB, or `/ActualText`, there is no Unicode source anywhere in the file, so the substituted character is simply wrong. For text-extraction pipelines this is silent corruption — a wrong, content-like character indistinguishable from real text.

## Proposed Solution

An **opt-in flag** — a sibling to the existing `set_preserve_unmapped_glyphs` — that selects how an uncovered CID in an Identity-ordered Type0 font is decoded, **without changing the current default**:

- **default (unchanged):** CID-as-Unicode — matches MuPDF; current output and tests are untouched.
- **strict (opt-in):** the uncovered CID decodes to `U+FFFD` instead of a guessed character — matches poppler and PDFBox.

It composes with `set_preserve_unmapped_glyphs`: strict + drop-unmapped (the default) omits the glyph entirely; strict + preserve surfaces `U+FFFD`. Best-guess rendering stays the default; consumers who need extraction fidelity can opt out of the guess.

## Feature Area

Text Extraction

## Priority

Would be helpful

## Use Case

[...]

PR #776, opened 2026-06-24
fix(fonts): emit U+FFFD for uncovered Identity-ordered Type0 CIDs, keep whitespace

### LINKED ISSUE #773: [Bug]: Embedded Type0 fonts decode CIDs missing from `/ToUnicode` to wrong, per-file-varying characters

schelip:
> **Stacked on #774** — opened as a draft. The diff includes #774's commits until it
> merges; once #774 lands in `main`, this reduces to the single Identity-ordered commit.

## Description

Completes #773 for the **Identity-ordered** case left open by #774. For an embedded
Type0 font with `CIDSystemInfo` ordering `Identity` (Adobe-Identity-0) whose
`/ToUnicode` omits some drawn CIDs, the CID-as-Unicode fallback is now restricted to
**whitespace** (`U+0020` → space, which producers routinely omit and is reliably
`U+0020`; dropping it would wreck word boundaries). Any other uncovered CID decodes
to **`U+FFFD`** — a detectable gap — instead of the CID read directly as a
plausible-but-wrong, per-file-varying character (`notificacao` → `no:ficacao`).

This is the default cascade settled in #775 (closed in favour of making this the
default rather than an opt-in flag): `/ToUnicode` → embedded `cmap`/`post` (both
already run first on this path) → structural-whitespace carve-out → otherwise
`U+FFFD`. The engine decides from the font, not the caller. Composes with
`set_preserve_unmapped_glyphs` (drop vs surface the `U+FFFD`).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Fixes #773

## Changes Made

- `src/fonts/font_dict.rs`: in the `Encoding::Identity` Type0 decode arm, the
  CID-as-Unicode fallback for Identity-ordered fonts is restricted to whitespace
  (`char_code == 0x20`); any other uncovered CID falls through to `U+FFFD`. The
  authoritative `/ToUnicode` and embedded `cmap`/`post` lookups, which run first, are
  unchanged — only the final numeric guess changes.
- `src/fonts/font_dict.rs` (tests): `test_type0_identity_uncovered_cid_is_fffd_keeps_space`.
- `tests/test_optimized_cmap_parsing.rs`: updated the Identity-ordered sibling to
  assert the new behaviour (uncovered → `U+FFFD`, whitespace retained).
- `CHANGELOG.md`: extend the `[Unreleased]` #773 entry to cover the Identity-ordered case.

## Testing

- [ ] I have run `cargo test --all-features` *(N/A for this crate: `--all-features` enables `fips`, which is a compile-time error alongside the default `legacy-crypto` and needs a CMake/C toolchain to build `aws-lc-fips-sys`; CI uses default-feature `cargo test`)*
- [x] I have run `cargo fmt`

- [ ] Python bindings updated (if needed) — N/A, internal decode change, no API surface change
- [ ] Python tests pass

## Documentation

- [ ] I have updated the documentation (README, docs/, code comments) — rationale comment added at the change site

## Checklist


yfedoseev (approved):
## Review — approved (code) ✅

Completes the **Identity-ordered** case that #774 deliberately left open (#773), and lands the cascade we settled on after closing the opt-in flag (#775) in favour of a correct default. The engine decides from the font, not the caller — right call.

**The logic is sound.** For an Identity-ordered Type0 font with a present-but-incomplete `/ToUnicode`:
- `identity_whitespace = is_identity_ordered && char_code == 0x20` keeps **only** the structural space (CID `0x20` → `U+0020`) on the CID-as-Unicode path. Producers routinely omit space from `/ToUnicode` and it is reliably `U+0020`; dropping it would wreck word boundaries (`no ficacao` → `noficacao`).
- Every other uncovered CID falls through to **`U+FFFD`** — a detectable gap — instead of a plausible-but-wrong, per-file-varying glyph.
- The authoritative `/ToUnicode` and embedded `cmap`/`post` lookups still run first and are untouched; a font with no `/ToUnicode` keeps the old heuristic. Purely subtractive on the corrupting path.

**Tests** cover all three branches: non-space Identity CID → `U+FFFD`, whitespace retained, and no-`/ToUnicode` keeps CID-as-Unicode. Good coverage of the decision boundary.

### CI — the two red checks are transient infra, not this change
- **Code Coverage**: `Token required - not valid tokenless upload` — a Codecov upload-token issue, unrelated to coverage or this diff.
- **Ruby 3.2 (windows)**: `download of aws-lc-rs failed … schannel: server closed abruptly` — a Windows crate-download network flake. A re-run clears it.

### Merge sequencing
This is stacked on #774. I'm squash-merging #774 into `main` now. Once it lands, please **rebase this branch onto `main`** — the diff will reduce to the single Identity-ordered commit (+ tests), and the stale merge commits drop out. After the rebase + a green re-run, this is good to go.

Thanks @schelip — this closes out #773 cleanly.

yfedoseev (changes requested):
## ⚠️ Regression sweep found one regression — empty `/ToUnicode` on an Identity-ordered font

Ran the version-over-version sweep (`docs/releases/process/corpus-regression-sweep.md`) for this PR: **baseline v0.3.68 vs #776 head (`2982c35c`)** over the **419-PDF font/CID-weighted manifest**, byte-diffing text/md/html and reading every change.

**Result: 1248/1257 (doc,fmt) pairs byte-identical. 3 docs changed — 2 benign, 1 regression.**

### 🔴 Regression — `PDFBOX-4322-Empty-ToUnicode-reduced.pdf`

| | text | md |
|---|---|---|
| v0.3.68 | `Justin` | `Justin` |
| #776 | *(empty)* | *(empty)* |

The whole document content — the word **`Justin`** — is **dropped**. Reference engines confirm `Justin` is the correct text:
- `pdftotext` → `Justin`
- PyMuPDF → `'Justin\n'`

**The font (from the PDF):**
```
/Encoding /Identity-H  /Subtype /Type0  /ToUnicode 10 0 R   (BaseFont TCDRYK+Calibri)
/CIDSystemInfo << /Ordering (Identity) /Registry (Adobe) /Supplement 1 >>
```
…and object `10 0 R` (`/ToUnicode`) is **empty** — **0** `beginbfchar`, **0** `beginbfrange`. It is *present* but maps nothing.

### Root cause

Two things combine, in `src/fonts/font_dict.rs`:

1. `has_usable_tounicode` treats a *present-but-empty* CMap as authoritative:
   ```rust
   let has_usable_tounicode =
       self.to_unicode.as_ref().and_then(|c| c.get()).is_some();   // empty CMap → Some(..) → true
   ```
2. This PR narrows the Identity fallback to whitespace only:
   ```rust
   let identity_whitespace = is_identity_ordered && char_code == 0x20;
   if !has_usable_tounicode || identity_whitespace { /* CID-as-Unicode */ }
   ```

For this file `has_usable_tounicode == true` (empty CMap still parses) and the `J/u/s/t/i/n` codes are non-space, so `identity_whitespace == false` → the CID-as-Unicode fallback is skipped → every glyph becomes `U+FFFD` → dropped → empty output.

**This is specific to #776.** #774 (now on `main`) used `if !has_usable_tounicode || is_identity_ordered`, so an Identity-ordered font kept full CID-as-Unicode and still recovered `Justin`. The whitespace-only narrowing removed that recovery.

### Suggested direction (your call)

The cleanest fix is to make an **empty** `/ToUnicode` count as *not usable* — require ≥1 actual mapping:
```rust
let has_usable_tounicode = self.to_unicode.as_ref()
    .and_then(|c| c.get())
    .is_some_and(|cmap| !cmap.is_empty());   // empty CMap ⇒ treat as absent
```
Then `!has_usable_tounicode` re-enables CID-as-Unicode for *both* arms (recovering `Justin`), while preserving #776's intent for a genuinely-present-but-*incomplete* `/ToUnicode` (real gaps → `U+FFFD`). A unit test on a Type0 font with a present-but-empty `/ToUnicode` (assert real CIDs still decode, not `U+FFFD`) would lock it in. Exact shape is up to you — flagging the root cause, not prescribing the patch.

### The other two changed docs — both benign (not this PR)

- **`iliad.pdf`** — baseline `<<TIMEOUT_OR_CRASH>>` in all three formats (v0.3.68 hit the 60 s sweep timeout); #776 extracts the full, correct text. Improvement / baseline flake, not a font-decode change.
- **`lyell_geology.pdf`** — title-page **reading-order** reflow (identical words, reordered); a `main` change unrelated to this font-only PR. Lateral.

Everything else is byte-identical. Once the empty-`/ToUnicode` case is handled, I'm happy to re-run the sweep.

yfedoseev:
## Re-review after `027b3925` — isolated sweep vs `main`

Built `main` (`c3ee58f2`, = #774) and #776 head (`027b3925`) and ran the **419-PDF font/CID manifest** through both, so this measures **#776's effect alone** (not main-vs-0.3.68 noise).

**Result: 1254/1257 (doc,fmt) pairs byte-identical. Exactly one document changes — and it's a regression. Zero documents improve.**

### The empty-`/ToUnicode` fix doesn't recover the text — it changes the failure

`PDFBOX-4322-Empty-ToUnicode-reduced.pdf` (Identity-H, `Ordering Identity`, empty `/ToUnicode`):

| | text |
|---|---|
| v0.3.68 | `Justin` ✅ |
| `main` (#774) | `Justin` ✅ |
| #776 `2982c35c` | *(empty)* ❌ |
| #776 `027b3925` | `J)’(i#` ❌ |

`pdftotext` and PyMuPDF both give `Justin`, so that's ground truth.

**Why `027b3925` still fails:** the new `is_empty()` check makes `has_usable_tounicode == false`, which **re-enables the GID→standard-glyph-name→AGL fallback** (`if !has_usable_tounicode { … gid_to_standard_glyph_name … }`). For a *subset* font the embedded GIDs don't follow the standard Mac glyph ordering, so that path yields wrong names (`)’(#`), shadowing the CID-as-Unicode mapping that `main`/#774 used to get `Justin`. The new unit test passes only because it uses a synthetic font where `char_code == Unicode`; the real subset font goes through CIDToGIDMap and doesn't.

### The bigger question: does #776 improve anything?

Across all 419 PDFs, **#776 adds `U+FFFD` to zero documents vs `main`** — the "wrong-char → honest gap" benefit it's built for doesn't trigger on a single corpus doc. Meanwhile it regresses the one Identity+empty-`/ToUnicode` file it touches.

The root tension: for **Identity** ordering, CID-as-Unicode is the *conventional correct* mapping (subset fonts routinely set CID == Unicode). Narrowing it to whitespace-only trades that common correct case for a partial-`/ToUnicode` scenario that isn't represented here. On this corpus the trade is purely negative.

I'd hold #776 until there's (a) a concrete PDF it demonstrably *improves*, and (b) no regression on the CID==Unicode case — these may be in tension with the whitespace-only approach. Flagging the data; the call is yours / the maintainer's. Happy to re-run anytime.

yfedoseev:
@schelip — could you share an example PDF where #776 **improves** extraction?

Our 419-PDF font/CID regression corpus shows **0 documents improved** by #776 vs `main` (and one regressed), so we don't currently have a case in-corpus that exercises the benefit. To validate the change before merging, it'd help a lot to have a concrete repro:

- ideally a small/reduced PDF (or the original `#773` document, or a minimised version of it) with an **Identity-ordered Type0 font** whose `/ToUnicode` is **present but incomplete**, where
- `main` decodes an uncovered CID to a **wrong, content-like character** (the `notificacao → no:ficacao` class), and
- #776 instead yields `U+FFFD`.

A few lines of "before (main) / after (#776) / expected" extraction alongside it would be perfect. With that we can add it to the corpus, confirm the improvement, and weigh it against the CID==Unicode regression (`PDFBOX-4322`: `Justin → J)'(i#`) — which is the trade-off we need to understand to make the call.

No rush, and thanks again for pushing on this — the underlying issue (#773) is real; we just want to land a fix that's a net win across real documents.

schelip:
Thanks @yfedoseev, your root-cause was exactly right, and it's fixed: for Identity-ordered fonts the GID→glyph-name guess can never be right, so it's now skipped there in favour of CID-as-Unicode. `Justin` decodes again, non-Identity fonts are untouched, and I replaced the unfaithful test with a real subset remap that fails without the fix.

Attached ([stripped.pdf](https://github.com/user-attachments/files/29381227/stripped.pdf)) is a real affected document reduced to the single affected word - `Identity-H`, `/Ordering Identity`, a present `/ToUnicode` that omits the `ti` ligature CID; fonts and `/ToUnicode` kept byte-identical, all other text/images removed. Verified extraction:

| Engine | Extraction | The "ti" ligature |
|---|---|---|
| `main` (#774) | `no:ﬁcação` | → `:` — wrong, reads as content |
| `#776` (default) | `noﬁcação` | → `U+FFFD`, dropped |
| `#776` + `preserve_unmapped_glyphs` | `no�ﬁcação` | → `U+FFFD`, surfaced |

So on a real file #776 turns the silent, content-like : into a detectable gap. Happy for you to re-run the sweep.

yfedoseev:
## Re-verification after `0252564b` — 419-PDF sweep vs `main`

Thanks @schelip — rebuilt this head and re-ran the full **419-PDF font/CID corpus** against `main` (`c3ee58f2`). Both earlier blockers are resolved and the net effect on the corpus is now **positive**.

**Result: 1251/1257 (doc, fmt) pairs byte-identical. Two documents change vs `main` — both are improvements. Zero regressions.**

Ground truth = pdftotext (poppler) + PyMuPDF; content similarity is whitespace-stripped char ratio vs PyMuPDF:

| document | `main` vs GT | `#776` vs GT | |
|---|--:|--:|---|
| `sample_fonts_solidconvertor.pdf` | 0.963 | **1.000** | ✅ exact match — recovers Czech `ž` and the en-dash (`…češtinou " ěščř ýáíé` → `…češtinou – ěščřžýáíé`) |
| `complex_ttf_font.pdf` | 0.035 | **0.479** | ✅ glyph mapping corrected toward GT (`:}lǥƹ:…` → `̽ଋҨǥƹ̽…`) |

### The two earlier concerns — both fixed

- **`Justin` regression is gone.** `PDFBOX-4322-Empty-ToUnicode-reduced.pdf` now extracts `Justin` again (== `main`); the `J)'(i#` from `027b3925` is fixed by skipping the GID→glyph-name guess for Identity-ordered fonts.
- **There is now a demonstrable in-corpus improvement** — `sample_fonts_solidconvertor.pdf` goes from 0.963 → an exact 1.000 match with poppler/PyMuPDF. At `027b3925` the sweep showed 0 improved / 1 regressed; at `0252564b` it's 2 improved / 0 regressed.

### Your `stripped.pdf` repro confirmed

| | extraction | the `ti` ligature |
|---|---|---|
| `main` | `no:ﬁcação` | `:` — wrong, reads as content |
| `#776` (default) | `noﬁcação` | U+FFFD, dropped (poppler GT: `no ﬁcação`) |

### Not from this PR

`iliad.pdf` and `lyell_geology.pdf` differ only vs the 0.3.68 baseline, not vs `main` (iliad = a 0.3.68 sweep-timeout flake; lyell = a reading-order reflow already in `main`) — lateral, unrelated to this font change.

LGTM on the data. The Identity root-cause analysis was spot on.

Issue #794, opened 2026-07-01
[Bug]: remove_artifacts / remove_footer doesn't remove all of the footer elements


ultrasaurus:
### Bug Description

When there is an odd/even pattern and/or elements with three letters, it seems to leave them behind.

pdf_oxide = "0.3.69"

### Steps to Reproduce

[1965-nelson.pdf](https://github.com/user-attachments/files/29560280/1965-nelson.pdf)

1. put the attached PDF in 'data' directory
2. run this rust code:

```
use pdf_oxide::{PdfDocument, Error};
use pdf_oxide::converters::ConversionOptions;

fn print_markdown(filename: &str) -> Result<usize, Error> {
    match PdfDocument::open(filename) {
        Err(e) => {
            println!("Error opening {}", filename);
            return Err(e);
        },
        Ok(doc) => {
            doc.remove_artifacts(0.2)?;
            let options = ConversionOptions { 
                detect_headings: true, 
                ..Default::default() 
            };
            let pages = doc.page_count()?;
            for page_index in 0..pages {
                let text = doc.to_markdown(page_index, &options)?;
                println!("--- Page {} ---", page_index + 1);
                println!("{}\n", text);
            }
            Ok(pages)
        }
    }
}

fn main() {
    let filename = "data/1965-nelson.pdf";
    match print_markdown(filename) {
        Err(e) => println!("Error: {:?}", e),
        Ok(num_pages) => println!("\n--- {} pages \n", num_pages)
    }
}
```

### Expected Behavior

All footer elements should be removed

### Actual Behavior

footer text left behind:
```
***84 - ACM***
- ***85***
***86 - ACM***
- ***87***
***88 - ACM***
```
etc.

### PDF Information (if applicable)

PDF attached above with repro steps

### Environment

- MacOS 15.7.3
- rustc 1.95.0 (59807616e 2026-04-14)

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Checklist

- [x] I have searched existing issues to make sure this is not a duplicate
- [x] I have provided all required information above
- [x] I am using the latest version of pdf_oxide

ultrasaurus:
@yfedoseev could you assign this to me, since I've been working on it (and now PRs require linked issues that are assigned).   https://github.com/yfedoseev/pdf_oxide/pull/976 needs to be merged first, then I can fix the other parts as separate PRs

yfedoseev:
[...]

PR #795, opened 2026-07-01
fix(794): partial fix for remove_artifacts / remove_footer

ultrasaurus:
## Description

improve artifact removal by recognizing page numbers and recto/verso layouts, while ensuring that content (that should not be removed) is not caught by building per-page signature set from whole lines.

Fixes the false positive where a single recurring word (e.g. "the," the last word of a different sentence on every page) got flagged as running-footer chrome just because it landed at the same page position.  This required grouping text spans on a line; however, that introduced regressions, so now words are clustered (as was done elsewhere for column detection).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [X] Tests
- [ ] CI/CD changes

## Related Issues

Fixes # https://github.com/yfedoseev/pdf_oxide/issues/794

## Changes Made

* Recognize bare page numbers even when they share a line with other footer text, instead of only when isolated alone on a line.
* Let a digit-only span's normalized signature ("84" → "#") register at all — a length filter had been silently discarding it.
* Stopped a spec-tag match from short-circuiting the whole detector, so the exact-text heuristic still runs to catch the surrounding constant chrome (brand names, etc.) that digit-detection alone doesn't cover.  Added
* detection for constant, digit-less fragments (like "com" or "ACM") that only repeat on every other page (recto/verso layouts), which the existing 50%-of-all-pages threshold was missing.  Fixed the two heuristic passes to
* share one snapshot of spans instead of each other's partial erasures, so neither pass under-counts what the other already touched.

## Testing

I've added new tests: 

**`tests/test_remove_footers.rs`** 
- `remove_footers_strips_combined_page_number_and_brand_span` — page number + brand fragment in one span (`"1 com"`); confirms fail-before/pass-after the #794 fix.
- `remove_footers_strips_parity_alternating_brand_fragment` — brand fragment on alternating (odd) pages only; confirms the legitimate recto/verso detection still works.
- `remove_footers_preserves_footnote_sharing_baseline_with_page_number` — footnote and page-number stamp sharing a baseline, far apart horizontally; pins the column-gap merge fix (`simplified_cluster_words_into_lines`).

**`tests/test_remove_footers_parity_both_sides.rs`**
- `remove_footers_preserves_phrase_present_on_both_parities` — a constant phrase present on both parities (80% even / 40% odd); pins the "other parity must be empty" fix for the digit-less-parity gate (the real Form 709 regression).

**`tests/test_remove_footer_preserves_content.rs`** (merged separately via PR #800) `remove_footers_preserves_real_prose_overlapping_band`, `remove_footers_preserves_common_word_across_unique_sentences`) 

Also created some common shared helpers in **`tests/common/mod.rs`** 

For manual verification:
`tools/probes/probe_remove_artifacts_chrome_only.rs`

- [X] I have run `cargo test` 
- [X] I have run `cargo fmt`

## Documentation

- [X] I have added/updated examples (if applicable)
- [ ] I have updated CHANGELOG.md -- needed?

## Checklist


ultrasaurus:
I did a bit more testing to make sure wanted text is not removed and fine for what I tested.

yfedoseev (commented):
Thanks @ultrasaurus — this is a careful, well-reasoned improvement, and the inline comments explaining *why* each change is needed make it easy to follow. Review below, including a check against the spec (`docs/spec/pdf.md`, ISO 32000-1 §14.8.2.2).

## Spec alignment (`docs/spec/pdf.md` §14.8.2.2 "Real Content and Artifacts")

- The spec defines **Pagination** artifacts as "ancillary page features such as running heads and folios (page numbers)" with standard `Subtype` values **Header / Footer / Watermark** (§14.8.2.2.2). `PaginationSubtype::Other` / `PageNumber` are **not** spec subtypes — they're this crate's classifications for *untagged* PDFs (geometry heuristic), so inferring header-vs-footer for them from a top/bottom band is a reasonable extension the spec doesn't cover, and treating them as a *partial* catch that must not suppress the heuristic pass is the right call. ✔️
- **One spec gap to consider:** for genuinely-tagged artifacts the spec provides the `Attached` array (Top/Bottom/Left/Right, §14.8.2.2.2) that states exactly which edge an artifact belongs to. Where that's present it's authoritative and should win over the 12% geometry band. Not a blocker (untagged spans have no `Attached`), but worth a follow-up so a tagged footer attached to `Top` isn't mis-binned by pixel position.
- **Band constant drift:** the new tagged/geometry pass uses `page_height * 0.12` while the heuristic pass just below still uses `height * 0.15`. Two different "footer zone" definitions in the same function will disagree on edge cases — please unify (or comment why they differ).

## What's done well
- The **single pre-erasure span snapshot** shared by both passes is a genuine correctness fix — `erase_region` clears the span cache, so the old code let pass 2 undercount what pass 1 had already thinned. Good catch and good explanation.
- Separating `tagged_removed` from `removed_count` so only real Header/Footer tags short-circuit the heuristic is correct.
- Recto/verso parity tracking and the length-1 signature fix (`"84"` → `"#"`) both map directly to the reported failure. Unit tests are **synthetic `from_bytes` `%PDF` buffers** — exactly the right approach.

## Things to address before merge
1. **Please don't commit `data/1965-nelson.pdf`.** It's a real-world reporter PDF; the repo convention keeps committed fixtures tiny/synthetic and `.gitignore` already excludes root `*.pdf` and `tests/fixtures/real/*.pdf`. The regression tests are synthetic (great), so the binary isn't needed for CI — suggest dropping it and having `examples/clean-pdf2markdown.rs` take the path as an arg (or documenting that the user supplies it).
2. **False-positive risk of the new digit-less "≥80% of one parity" chrome rule.** Stripping constant, digit-less fragments that merely repeat on alternating pages is the highest-risk change here — it can erase legitimate recurring body text (a section label, a short recurring caption) that happens to sit in the footer band. The ≥2-chars/one-alphabetic guard helps but doesn't bound it. This needs a run against the full regression corpus (the 400+ PDF sweep) to confirm it doesn't remove real content elsewhere before merging.
3. **`removed_count` can double-count.** `erase_region` just appends the rect (idempotent for output), but a span matched by *both* the geometry pass and the text-signature pass increments the count twice. Output is fine; the returned/logged count over-reports. Consider de-duplicating erased rects per page.
4. **Scope:** the title says "partial" — worth spelling out in the PR which cases still leak (the alternating `"# - ACM"` / `"- #"` forms and their `***…***` bold-wrapped variants) so #794 can either be fully closed here or the remainder tracked.
5. **Housekeeping:** a `CHANGELOG.md` entry (this changes what `remove_artifacts` strips, which is user-visible) and a `cargo test --all-features` / full `cargo test` run.

Net: the approach and reasoning are sound and spec-consistent; the merge-blockers are the committed real PDF (#1) and validating the digit-less heuristic against the corpus (#2). Happy to help with the regression-sweep validation.

yfedoseev:
## Corpus validation (the regression-sweep I offered)

Ran `doc.remove_artifacts(0.2)` + `to_markdown_all()` over the 419-PDF regression corpus, **main (911270ad) vs main+this-PR**, and diffed the output. This surfaces exactly the false-positive risk flagged in point #2 above.

**Scope:** output changed on **51 / 419** docs; removed-artifact count changed on **118 / 419** (the PR removes strictly more).

**Good news first:**
- Target #794 (`1965-nelson.pdf`): improves — `removed` 22 → 61, the "ACM" leftovers drop 12 → 4 and the bare page-number lines are gone. (Still partial: 4 "ACM" remain.)
- Pure-folio / running-head docs are handled correctly, e.g. `lyell_geology` removes only page numbers and `Ch. II.] … COSMOGONY.` running heads. 👍

**But it also removes real content on a range of documents** (verified by word-frequency delta, i.e. tokens whose count actually dropped — not reflow):

| Doc | net words | Real content removed (not folios/heads) |
|---|---|---|
| `war_peace` | **−30,890** | 307 `###` headings, body function words (`the` −3924, `and` −2501, `of` −1771), TOC entries |
| `alice_old` | −481 | 24 ` ``` ` code-fence lines, body words, character names (`Mock`, `Gryphon.`) |
| `tracemonkey…` | −48 | section numbers (`5.1`, `6.5`), inline subscripts (`T16`, `t1`, `rooted2`) |
| `pocorgtfo08` | −101 | 32 table pipes `|` + `\|---\|` separators (a whole table), binary/code listing tokens |
| `fw2` / `fw2_2024` / IRS 1120/941… | −22/−21/… | **form box numbers `1 2 3 4`** and the `22222` control number |

**Root cause** is the same aggressiveness the review flagged: the new signals — length-1 `#` signatures, the recto/verso parity thresholds, and the digit-less "≥80% of one parity" chrome rule — match anything that recurs with a digit or short token near the band: section numbers, TOC folios *inside* content, form field numbers, table digits, and dense two-column body lines whose first/last line falls in the 12–15% band. The geometry-band `Other`/`PageNumber` match compounds it on multi-column pages.

**Suggested direction** (happy to pair on this): tighten each new signal so it can't fire on body/structure — e.g. never erase a span the layout classifies as a heading / list / table cell / code; require page-number-shaped tokens (short, numeric, isolated in the band) rather than any recurring digit; and cap removal to spans actually inside the header/footer band (not signature-matched text elsewhere). Then re-run this sweep — the target should stay fixed while `war_peace`/forms/`tracemonkey` go back to zero real-content loss.

<details><summary>method</summary>main = `pdfmain_wt@911270ad` (this PR's base); candidate = this PR built from the same base; identical `remove_artifacts(0.2)`+`to_markdown_all` over the shared 419-PDF manifest; "real content removed" = per-token count drops, so it excludes pure reflow/re-wrapping.</details>

ultrasaurus:
Thanks so much for the quick turnaround and thorough feedback.  I submitted the PR an after thought -- since was trying to fix for my own use case.  I decided to add some tests to pin down working behavior that I had broken.  Since I found no test coverage in this area (correct me if I missed something!), I started with the basics and found an unrelated failure:

see https://github.com/yfedoseev/pdf_oxide/pull/798

feedback on that PR would be much appreciated!  I'm a PDF novice so took me a while to build the test helper I needed, but it was a good learning experience :)

ultrasaurus:
we need a different kind of fix, a new way to recognize "the" as not a footer element.  I wonder if we could merge it to with its neighbors (if it has any) to learn that its not actually the same text on every line? 

Is there elsewhere that the code groups text into logical lines/paragraphs?  If we were evaluating the whole line (or paragraph / column) of text, then it would never be classified as a footer.

yfedoseev:
Great instinct — yes, that's the right direction, and the infrastructure already exists.

The root cause of the `"the"` false positive is that the artifact detector computes its repetition signature **per span**. `ensure_running_artifact_signatures` normalizes and counts each span in isolation, so a lone word that happens to recur in the band ("the", "of", "com") gets a constant signature and looks like chrome — exactly your case. A genuine footer, by contrast, is a whole repeated *line* ("Chapter II.]  COSMOGONY.", "84  •  ACM"), so moving the signature up to the **line** level fixes the false positive *without* weakening real-footer detection — if anything it strengthens it, because body words never form a repeated whole-line signature while running heads/folios do.

There is existing line-grouping you can reuse rather than build:

- **`OrderedSpans::group_into_lines(tolerance)`** — `src/pipeline/ordered_span.rs:329`. Groups spans into logical lines by baseline proximity; this is the one I'd reach for first.
- **`cluster_words_into_lines(words, epsilon_y)`** — `src/layout/clustering.rs`. Same idea at the word/TextBlock level.

Sketch of the change: in the band, group the spans into lines first, build the repeated-signature from the *line's* normalized text (with page-number tokens normalized to `#`, which the code already does), and only then mark spans for erasure if their **line** signature is the repeated one. `"…and gave it to the"` and `"…walked past the"` produce different line signatures → never flagged; `"Chapter II  84"` / `"Chapter II  85"` normalize to the same `"chapter ii #"` → correctly flagged on every page.

For reference, the automatic markdown running-head/footer detection (used by `to_markdown`) already works this way — it groups the header/footer band into lines and normalizes page numbers before deciding, which is why it doesn't hit the `"the"` problem. Aligning `remove_artifacts` with that line-level approach is the consolidation worth doing.

One caveat to keep the precision you want: still bound erasure to spans actually inside the header/footer band (don't erase a signature-matched line that appears mid-body), and keep the existing guard that never removes spans the layout classifies as heading/list/table/code. With line-level signatures + those guards, the corpus over-strip cases from the earlier sweep (war_peace, the IRS forms, tracemonkey section numbers) should go back to zero real-content loss while `1965-nelson` stays fixed.

ultrasaurus:
Note: I looked at reusing:

- **`OrderedSpans::group_into_lines(tolerance)`** — `src/pipeline/ordered_span.rs:329`. Groups spans into logical lines by baseline proximity; this is the one I'd reach for first.
- **`cluster_words_into_lines(words, epsilon_y)`** — `src/layout/clustering.rs`. Same idea at the word/TextBlock level.

However the difference across types would have created a more heavy weight solution.  I didn't want to do a refactor bundled into this PR, but I considered a function using traits that could be used by the now three implementations that run the same kind of process with different types and output.

Similarly it also didn't feel right to not use a constant for `.font_size.max(6.0)` in `group_band_lines`, but followed the pattern I saw elsewhere.

yfedoseev (changes requested):
## 🔴 Changes requested — validation gate + negative test still outstanding

@ultrasaurus the line-level rework (`group_band_lines`) is exactly the direction we discussed, and the committed-PDF and `≥80%`-parity guards are addressed. Two blockers remain from the 07-02 review:

1. **The 419-corpus re-validation gate hasn't been posted.** The digit-less `≥80%`-parity signal can still fire on any recurring ≥2-char band line. To reproduce the over-strip concern:
   ```sh
   # run remove_artifacts(0.8) then to_markdown, diff vs plain to_markdown; removed lines must be chrome only
   ```
   on `war_peace.pdf`, the IRS forms, and `tracemonkey` — and confirm **zero body lines removed**. (I can run this sweep and post the deltas if useful.)
2. **No negative test.** The two new tests only assert chrome *is* removed. Please add one asserting a body/heading line whose last word recurs in the band is **preserved** — the exact regression the line-level design exists to prevent.

Quick cleanups also still open: `removed_count` double-counts (`:6555` + `:6624`); the footer zone is `0.12` in pass 1 (`:6529`) vs `0.15` in pass 2 (`:6600`).

ultrasaurus:
@yfedoseev apologies for not re-reading all of your very good feedback and not mentioning the regression test that I added in https://github.com/yfedoseev/pdf_oxide/pull/800
-- [‎tests/test_remove_footer_preserves_content.rs](https://github.com/yfedoseev/pdf_oxide/pull/800/changes#diff-85db24425b5ad05ede96bf8ff183b9c502fbdc8bd3d64aef0925a23b2430bb4e) used to fail on this branch, now passes

Re: "The 419-corpus re-validation gate hasn't been posted" I didn't realize that was something I had access to!  I see scripts that refer to `~/projects/pdf_oxide_tests/` -- is there another repo where I can find `war_peace.pdf`, the IRS forms, and `tracemonkey` — and confirm zero body lines removed?  would be great to confirm above test actually works!

I did find https://github.com/yfedoseev/pdf_oxide/issues/575 which mentions `war_peace.pdf`is from the Internet Archive and appears to be this PDF:
https://dn760008.eu.archive.org/0/items/warandpeace030164mbp/warandpeace030164mbp.pdf
I will try to run `remove_artifacts(0.8) then to_markdown, diff vs plain to_markdown; removed lines must be chrome only` and see if it works with my changes

ultrasaurus:
I created a little tool for checking what was removed and it seems the code works on https://dn760008.eu.archive.org/0/items/warandpeace030164mbp/warandpeace030164mbp.pdf

```
--- per-page span diff (what `remove_artifacts` actually erased) ---
page 43: 735 spans -> 734 spans
  ERASED: "24 "  bbox=(x=40.1, y=628.5, w=11.8, h=7.0)
page 56: 791 spans -> 790 spans
  ERASED: "37 "  bbox=(x=403.0, y=632.4, w=12.1, h=9.0)
page 64: 835 spans -> 834 spans
  ERASED: "45 "  bbox=(x=403.1, y=631.9, w=12.1, h=10.0)
page 66: 828 spans -> 827 spans
  ERASED: "47 "  bbox=(x=403.6, y=631.9, w=12.1, h=10.0)
page 68: 1008 spans -> 1007 spans
  ERASED: "49 "  bbox=(x=402.4, y=631.9, w=12.4, h=11.0)
page 76: 833 spans -> 832 spans
  ERASED: "57 "  bbox=(x=403.4, y=631.7, w=12.0, h=10.0)
page 83: 731 spans -> 730 spans
  ERASED: "64 "  bbox=(x=41.3, y=630.2, w=13.7, h=9.0)
page 103: 814 spans -> 813 spans
  ERASED: "84 "  bbox=(x=41.1, y=629.5, w=13.7, h=10.0)
page 143: 775 spans -> 774 spans
  ERASED: "124 "  bbox=(x=41.2, y=623.6, w=15.3, h=7.0)
page 408: 679 spans -> 678 spans
  ERASED: "389 "  bbox=(x=218.6, y=29.2, w=17.8, h=19.0)
page 476: 768 spans -> 767 spans
  ERASED: "457 "  bbox=(x=395.8, y=628.3, w=16.1, h=11.0)
page 566: 828 spans -> 827 spans
  ERASED: "547 "  bbox=(x=397.6, y=626.4, w=16.6, h=11.0)
page 582: 692 spans -> 691 spans
  ERASED: "563 "  bbox=(x=219.0, y=34.8, w=17.5, h=10.0)
```

I also found `scripts/download_irs_forms.py` so will test those as well.  Still can't find `tracemonkey`

@yfedoseev let me know if you don't want verification tool checked in https://github.com/yfedoseev/pdf_oxide/pull/795/changes#diff-99767657450531e5e3f2ff7c603dff53db6cc318f3246647fd2719e05400ea60 or if that duplicates some other verification tool that I didn't find.  Thanks!

ultrasaurus:
Tested all the IRS forms (downloaded with `scripts/download_irs_forms.py`) -- verified all do not removed form content, except 3 which also fail on main (see https://github.com/yfedoseev/pdf_oxide/issues/879). 

One regression was found that required clustering words so if a constant footer or page number was on one side of the page, it would not be evaluated together with a footnote or unrelated text span on the other side of the page. 

`cluster_words_into_lines(words, epsilon_y)` in `src/layout/clustering.rs` was exactly what was needed, just needed a refactor so the simplified version was not feature flagged.

also fixed:
- no longer double-count a span erased by both passes
- inconsistent margin band: new const RUNNING_ARTIFACT_BAND_FRACTION
- added new tests (see top description with full list)

happy to verify with `tracemonkey` just don't know where to find it

ultrasaurus:
rebased on main, old and new tests pass -- some manual merges but no real confilcts.  

ready for review @yfedoseev

ultrasaurus:
- - -
   - - -

recheck

ultrasaurus:
@yfedoseev attempted to sign the CLA with no luck.  

guessed at the fix(): format, would be good for that to be documented in https://github.com/yfedoseev/pdf_oxide/blob/main/CONTRIBUTING.md#commits-dco-and-review

yfedoseev:
@ultrasaurus thanks for flagging it — I dug into why it didn't take, and it's a small formatting thing, easy to fix.

Your commits are fine: all 8 are `Sarah Allen <sarah@ultrasaurus.com>` → mapped to your GitHub account, exactly as they should be. The signature just didn't register because the sign comment carried extra decoration around the phrase.

When you signed, you copied the bot's message verbatim:

```
   - - -
   - - -

recheck
```

Those `- - -` lines and the indentation are only the bot's *visual* separators, and the trailing `recheck` is a separate command — bundling them into the same comment as the phrase kept the bot from recording it as a clean signature. (For contrast, a signature that went through on another PR was just the bare line, nothing else.)

**The fix:** post one new comment containing *only* this line — no indentation, no `- - -`, no `recheck` after it:

```
```

That's it — the bot will record you and flip the CLA check green. If for some reason it's still red a minute later, *then* drop a separate one-word `recheck` comment to re-trigger it.

Fair point that the exact format isn't spelled out in CONTRIBUTING.md — if this trips up more people we'll document it properly; for now the single-line phrase above is all it needs.

(Separately, the review feedback on the fix itself still stands — this is just to get the CLA gate unstuck.)

ultrasaurus:

ultrasaurus:
recheck

ultrasaurus:
FYI: Note about use of AI

I use Claude as assistive technology.  I have a print disability, and also much scrolling and visual context switching can trigger vestibular migraines.  Reading long prose (as is often included in technical documentation) is not practical; however, reading and writing code is unaffected.  I used Claude to learn PDF format, write first draft of tests, the probe tool, and some of the code. Whatever code I didn't write myself, I have reviewed and revised as needed.  

I also use speech-to-text and am working on a text-to-speech (TTS) tool that I can use for listening to long-format text.  The TTS tool has a PDF import feature which led me to attempt to fix this bug.

yfedoseev:
Thanks for squashing and adding the DCO sign-off — that's resolved. Ran a correctness review plus our native corpus-signature sweep (`corpus_sig`: text hash + word-length + page-rotation per PDF) against a 419-PDF regression set, diffed directly against current `main` since this branch is now rebased on top of it.

**Corpus result**: 40/419 files changed. Word-max-length and page rotation are identical across all 40 — no word-fusion, no crashes. 38 of the 40 are small char-count *decreases* (more chrome correctly stripped, consistent with intent); 2 are increases (`259.pdf`, `292.pdf` — the same `prefilled_f1040.pdf` also touched by #798), meaning this branch actually *stops* some prior over-stripping on that file relative to `main`. Category breakdown of the 40: 27 in "reading-order 2-col/multi-region" (academic papers with running headers/footers), 8 "broad ripple tail", 4 "tables guard", 1 "marginalia" — all plausibly in-scope for header/footer heuristics, nothing in an unrelated category.

**One open code-level question**: `group_band_lines` now signs a whole clustered *line* rather than each span, specifically to stop a coincidentally-recurring lone word from being misflagged as chrome. That creates a plausible mirror-image risk: if genuine page-specific content (the tail of a long line, an incidental word/digit) falls in the margin band and clusters onto the *same line* as real repeating chrome on one particular page — within the ~30pt/3×font-size column-gap threshold this uses — the concatenated per-page literal becomes unique to that page, so that page's signature won't hit the cross-page recurrence threshold and the genuine chrome silently survives un-stripped, just on that one page. None of the three new tests cover this (all keep the unrelated content far enough apart to land on a separate line). Might be a non-issue in practice, but worth a synthetic test with content and chrome sharing a line to rule out.

**Heads up on the other open PR in this area**: #798 also restructures `ensure_running_artifact_signatures` (changes the signature map's value type and the first-occurrence exemption). Since this branch is rebased on `main` (which doesn't include #798), whichever of the two lands second will need a real reconciliation pass against the other's changes here, not a mechanical rebase — flagging now so it doesn't surprise whoever goes second.

No corpus-regression sweep was documented in the PR body despite this touching layout/extraction heuristics (rule 2 of CONTRIBUTING.md) — the above should cover that gap for the 419-PDF set.

yfedoseev:
@ultrasaurus Thanks for the disclosure, and for the context — appreciated on both counts.

To be clear: the AI policy isn't "don't use AI." I use it heavily on this project myself. What it's actually about is understanding and responsibility — you need to be able to explain every line you submit, and stand behind its correctness, regardless of what tool helped produce it. Assistive use like yours (learning the format, drafting tests, getting a first pass on code you then review and revise) is exactly the kind of responsible use the policy is written to allow, not exclude.

The reason the policy exists at all: we've had PRs on other repos that were pure slop — no tests, broken implementations, functionality regressed, clearly never run against real input. Reviewing a PR properly (reading the diff, checking the reasoning, running it against real documents) takes me on the order of 30 minutes minimum. If low-effort AI-generated volume grows, that stops being sustainable and it's real contributors like you who lose out, since your PRs end up waiting behind noise. That's the failure mode we're guarding against, not AI use itself.

Your PRs here have consistently had the thing that matters: real regression tests that fail before the fix, a synthetic reproducer, and revisions when issues come up in review. That's the bar, and you're clearing it. Thank you for the contribution, and for taking the time to write this up.

ultrasaurus:
Regarding open code-level question: 
> `group_band_lines` now signs a whole clustered line rather than each span, specifically to stop a coincidentally-recurring lone word from being misflagged as chrome. That creates a plausible mirror-image risk: if genuine page-specific content (the tail of a long line, an incidental word/digit) falls in the margin band and clusters onto the same line as real repeating chrome on one particular page — within the ~30pt/3×font-size column-gap threshold this uses — the concatenated per-page literal becomes unique to that page, so that page's signature won't hit the cross-page recurrence threshold and the genuine chrome silently survives un-stripped, just on that one page. None of the three new tests cover this (all keep the unrelated content far enough apart to land on a separate line). Might be a non-issue in practice, but worth a synthetic test with content and chrome sharing a line to rule out.

It is an issue in practice!  Wrote two tests, both pass on main, one fails in this branch.  Will look into fixing

yfedoseev:
Draft, so no approval yet — but I ran it end to end today and it's in good shape.

- All 14 tests across the five files pass.
- Revert-checked: restoring `src/document.rs` and `src/layout/clustering.rs` from `main` turns 4 of the 5 tests in `test_remove_footers.rs` red, so they genuinely pin the behaviour.
- **419-PDF regression sweep, this branch merged onto current `main` vs `main`: zero documents change.** That was the one I most wanted to see given the `clustering.rs` edits, since those sit under default extraction rather than the opt-in removal API.
- No PDF binaries in the diff — the earlier fixture-policy concern is resolved.

One coordination issue worth catching before either lands: **this PR and #967 both add a file named `tests/test_remove_footers.rs`**, with different contents. They'll collide on whichever merges second. #967 is a small persistence fix in `DocumentEditor::from_document`; yours is the detection/parity work. Simplest fix is for one of them to rename — since yours is the larger set and closer to the `remove_footers` detection logic, keeping the generic name here and renaming the other by defect class (something like `erase_regions_persist_through_editor`) seems the less disruptive split, but that's @yfedoseev's call.

The branch is behind `main` (still mergeable). Worth a rebase when you take it out of draft.

ultrasaurus:
Unfortunately, my corpus testing revealed yet another failure.  I'm working on fixing from yet another new test case.  Specifically failed IRS_Form_1120_2024.pdf in fundamental way that I want to fix in isolation

ultrasaurus:
closing in favor of https://github.com/yfedoseev/pdf_oxide/pull/976 -- parity work from this branch will be a relatively simply change if that PR is merged

PR #799, opened 2026-07-02
fix: preserve add_text() overlays through destructive redaction on the same page

thomnico:
## Summary

Three related bugs surfaced together when combining `add_text()` with `apply_redactions_destructive()` on the same page — a common pattern for PII redaction (redact a sensitive value, then insert a replacement token at the same position).

1. **`/Contents` overwritten by destructive redaction.** The destructive-redaction block replaced `/Contents` wholesale with a single reference to the redacted stream, discarding the `[original, overlay]` array already built by an earlier `add_text()` merge step. The overlay stream object was still written to the file but became unreferenced by any page — silently dropped, no error.
2. **Added text's font never registered in `/Resources/Font`.** Text added via `add_text()` on an existing page referenced its Base-14 font (e.g. `/Helvetica`) as a `Tf` resource name, but nothing declared that name in the page's `/Resources/Font` dict — which, for a page loaded from an existing document, is very commonly an *indirect reference*, not an inline dictionary. Poppler's `pdftotext` rejected the resulting stream (`Unknown font tag`); pdf_oxide's own extractor tolerated it inconsistently.
3. **`save_page()` replaced instead of accumulated overlay text.** `self.overlay_additions.insert(page_index, added)` discarded whatever a prior `save_page()` call had staged for the same page, so only the last of several `add_text()` + `save_page_from_editor()` round-trips survived.

## Fix

- Bug 1: replace only the first `/Contents` array entry (the original content reference) when it's already an array, instead of overwriting the whole field.
- Bug 2: resolve `/Resources` and `/Resources/Font` through indirect references where needed, and register a minimal Type1 Font object for any Base-14 name used by overlay text, preserving pre-existing font entries (e.g. `/F1`, `/F3` from the original document).
- Bug 3: accumulate via `.entry(page_index).or_default().extend(added)` instead of `.insert()`.

## Testing

All three fixes were found and validated together against a real PDF (pandoc/xelatex output, embedded font) containing PII, redacted destructively and replaced with tokens, then verified **independently with Poppler's `pdftotext`** — not just pdf_oxide's own extractor — to confirm both that the redacted values are gone and the replacement tokens are present and correctly rendered.

Two regression tests added in `editor::document_editor::tests`:
- `destructive_redaction_preserves_add_text_overlay` (bugs 1 & 2)
- `multiple_add_text_calls_on_same_page_all_survive` (bug 3)

```
cargo test --lib
test result: ok. 5681 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
```

`cargo clippy --lib` clean on the changed code (no new warnings).

## Context

Found while building a PII-redaction pipeline (detect via regex, redact destructively, insert a `{{TOKEN}}` placeholder at the same position) — this is exactly the workflow the `redaction` module's docs describe (ISO 32000-1 §12.5.6.23, "shall remove all traces"), so I wanted to make sure it worked end-to-end before adopting pdf_oxide for that use case. Happy to adjust the approach if you'd prefer a different fix shape (e.g. tracking the original content object ID explicitly instead of assuming array position 0 — noted as an alternative in the code comment).

### DIFF
--- a/src/editor/document_editor.rs
+++ b/src/editor/document_editor.rs
@@ -2459,6 +2459,38 @@ impl DocumentEditor {
                                         None
                                     };
 
+                                // Any Base-14 standard font name referenced by a `TextContent`
+                                // in `overlay_additions` that isn't already a resource on this
+                                // page needs its own Font object allocated, so the `/BaseFont`
+                                // name used by the `Tf` operator resolves to a real
+                                // `/Resources/Font` entry rather than a dangling resource name.
+                                const BASE14_NAMES: &[&str] = &[
+                                    "Helvetica", "Helvetica-Bold", "Helvetica-Oblique",
+                                    "Helvetica-BoldOblique", "Times-Roman", "Times-Bold",
+                                    "Times-Italic", "Times-BoldItalic", "Courier",
+                                    "Courier-Bold", "Courier-Oblique", "Courier-BoldOblique",
+                                    "Symbol", "ZapfDingbats",
+                                ];
+                                let mut overlay_font_ids: HashMap<String, u32> = HashMap::new();
+                                if let Some(added) = self.overlay_additions.get(&source_page_index) {
+                                    let mut needed: Vec<String> = added
+                                        .iter()
+                                        .filter_map(|el| match el {
+                                            ContentElement::Text(t)
+                                                if BASE14_NAMES.contains(&t.font.name.as_str()) =>
+                                            {
+                                                Some(t.font.name.clone())
+                                            },
+                                            _ => None,
+                                        })
+                                        .collect();
+                                    needed.sort();
+                                    needed.dedup();
+                                    for name in needed {
+                                        overlay_font_ids.insert(name, self.allocate_object_id());
+                                    }
+                                }
+
                                 // Apply page property modifications if any
                                 let mut final_page_obj = if let Some(props) =
                                     self.modified_page_props.get(&source_page_index)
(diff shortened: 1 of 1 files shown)


yfedoseev (changes requested):
Thanks @thomnico — really solid diagnosis. All three bugs are real, the root-cause write-up is accurate, and validating with Poppler's `pdftotext` independently (not just pdf_oxide's own reader) is exactly the right instinct. I also confirmed the important safety property holds: **the "remove all traces" redaction guarantee (§12.5.6.23) is preserved** — it rests on `redacted_orphan_ids` + the unconditional write-skip, which this PR doesn't touch, so no redacted content becomes reachable. Bug-3's `.entry().or_default().extend()` accumulation is correct as-is.

Requesting changes on two points before we merge — as written the fix is incomplete/produces malformed output for common cases, so I'd like these addressed first:

**1. (must fix) Bug-1: `/Contents` entry-0 replacement breaks multi-stream pages.** Keeping `arr[1..]` assumes the original `/Contents` was a single stream. A page whose original `/Contents` is already `[c1, c2]` (very common) becomes `[redacted, c2]`, where `c2` is an *original* stream. The orphan-drop backstop physically drops `c2` (so no leak — good), but `/Contents[1]` is then a **dangling reference**: lenient readers render fine, strict validators (veraPDF/Poppler in strict mode) will flag a broken array. Your own suggested alternative is the right shape — track the original content object id(s) and drop *all* originals, keeping only genuine overlay/additions refs.

**2. (must fix) Bug-2: the font is registered under the raw name, but `Tf` emits the `map_font_name`-transformed name.** Registration keys on `t.font.name`, while the overlay stream writes `map_font_name(&t.font.name, is_bold())`. These diverge exactly in the cases the fix is meant to cover, so the "font not registered" bug is still live for:
- **bold/italic overlays** — `Helvetica` + bold → stream emits `/Helvetica-Bold Tf`, but only `/Helvetica` gets registered → dangling `Tf`;
- **Symbol / ZapfDingbats** — in `BASE14_NAMES`, but `map_font_name` routes them to Helvetica → an unused font object is written while the referenced `/Helvetica` dangles;
- **generic names** ("Arial", "sans-serif") — not Base-14, so unregistered, but the stream still emits `/Helvetica Tf`.

Keying registration off the `map_font_name(...)` output (the name actually emitted) closes all three. Symbol/ZapfDingbats can also be dropped from `BASE14_NAMES` since the writer never emits them.

**3. (nice to have) Bug-2: inherited `/Resources` shadowing.** When a page has no own `/Resources` key (legal — inheritable from the `/Pages` chain per §7.7.3.4), building a fresh font-only `/Resources` shadows the inherited dict and drops the page's original fonts/XObjects. The redaction code's own `build_page_font_metrics` already walks `/Parent` for resources — reusing that pattern here keeps it consistent.

**Tests:** the two regression tests are good but only cover single-stream `/Contents`, one upright Base-14 font, and inline (non-indirect) `/Resources` — so the indirection path (this PR's headline complexity) and the three cases above aren't exercised. Editor round-trip tests for multi-stream `/Contents`, indirect `/Resources`, and a styled font would lock those down, and promoting your manual Poppler cross-check into CI would be great if it's easy.

No corpus regression sweep is needed here — this is confined to the editor/redaction write path, which our extraction sweep never touches, so unit + editor round-trip tests are the right gate. Happy to re-review as soon as #1 and #2 are in.

yfedoseev (commented):
## 🔴 Still blocked — two must-fixes from the 07-03 review are unaddressed

@thomnico head is still `c47c16a` (the reviewed commit, no new pushes). Both must-fixes are verified still present:

1. **Multi-stream `/Contents` (`document_editor.rs:2727`):** the code rewrites only `arr[0]` and keeps `arr[1..]`. Reproduce with a page whose original `/Contents` is an **array** `[c1 c2]` (legal, §7.7.3.3): the result is `[redacted, c2]` where `c2` is the original unredacted stream, leaving a dangling `/Contents[1]` after the orphan-drop backstop removes `c2`. (No live leak today — the backstop saves it — but strict validators flag the dangling ref.) Fix: track and drop **all** original content object ids.
2. **Font name mismatch (`:2478` vs `content_stream.rs:861`):** the overlay font is registered under the **raw** name, but the `Tf` operator emits `map_font_name(name, is_bold())`. Reproduce with an overlay whose `TextStyle` is **bold/italic**, a **generic** family (`"Arial"`), or **Symbol/ZapfDingbats** → the emitted `/Tf` name is undeclared in `/Resources` (the exact "Unknown font tag" this PR set out to fix). Fix: key registration off the `map_font_name` output and drop Symbol/ZapfDingbats from `BASE14_NAMES`.

Tests currently pass only because they exercise single-stream `/Contents` + an unstyled Base-14 font. Happy to re-review as soon as these land.

yfedoseev:
Closing this out to keep the review queue current — not a rejection of the work.

To recap where it stands: your diagnosis here was genuinely strong (all three bugs real, accurate root-cause write-up, and validating against Poppler `pdftotext` independently was exactly right). The blocker is that the two must-fixes from the [07-03 review](https://github.com/yfedoseev/pdf_oxide/pull/799#pullrequestreview) — the multi-stream `/Contents` array case (`document_editor.rs:2727` rewrites only `arr[0]`) and the second item — were confirmed still unaddressed in the [07-14 follow-up](https://github.com/yfedoseev/pdf_oxide/pull/799#issuecomment), with no new pushes since (head is still `c47c16a`). It's been three weeks with no response, so I'm closing rather than leaving it open indefinitely.

This is fully reversible and the door's wide open:

1. **Open a tracking issue** for the redaction-overlay bugs — they're real and worth fixing; anchoring the work to an accepted issue is our rule-1 baseline.
2. When the two must-fixes land, **push and reopen** (or open a fresh PR) and I'll pick the review straight back up — the hard diagnostic work is already done, so it should be a short path from there.

Thanks again for the careful investigation — hope to see it come back.

thomnico:
Reopened as #943 — the two must-fixes from your review are in, rebased onto `main` (v0.3.75), with the requested tests (multi-stream `/Contents`, indirect `/Resources`, styled font) and independent Poppler validation. Tracked by #941. This PR couldn't be reopened directly after the rebase/force-push, hence the fresh one.

Issue #804, opened 2026-07-04
[Bug]: `extract_words` / `extract_spans` / `extract_text_lines` garble text on 90°/270° rotated pages


ankursri494:
### Bug Description

On pages with **/Rotate** of `90` or `270`, the spatial text-extraction methods (`extract_words, extract_spans, extract_text_lines`) do not apply the page rotation. Grouping is done in raw (unrotated) coordinate space, so it clusters along the wrong axis and fuses unrelated cells into giant tokens, `extract_words` can return a single "word" containing an entire column.
(180° pages are unaffected, since horizontal text stays horizontal.)

### Steps to Reproduce

1. Open a PDF with a page rotated 90° or 270° e.g. [tests/pdfs/nics-background-checks-2015-11-rotated.pdf](https://github.com/jsvine/pdfplumber/blob/stable/tests/pdfs/nics-background-checks-2015-11-rotated.pdf) from pdfplumber's public test suite (/Rotate 90).
2. Call `extract_words(0)` on the rotated page.
3. Observe a single token containing an entire column of text.
4. Compare with `pdfminer`/`pdfplumber` **extract_words()** on the same page, which returns normal words.

### Expected Behavior

On a **90°/270°** page, extract_words / extract_spans / extract_text_lines should group along the visual text direction and return normal words/lines, matching what the same page yields via `pdfminer/pdfplumber` (and matching a non-rotated equivalent).

### Actual Behavior

Whole columns are concatenated into single tokens, and cells from different visual rows are fused into one line. `extract_words(0)[0]` is a 1235-character token; `extract_text_lines(0)` fuses separate rows into one line.

### PDF Information (if applicable)

Public, PII-free reproducer from pdfplumber's MIT-licensed test suite: [tests/pdfs/nics-background-checks-2015-11-rotated.pdf](https://github.com/jsvine/pdfplumber/blob/stable/tests/pdfs/nics-background-checks-2015-11-rotated.pdf) /Rotate 90.
[...]

PR #814, opened 2026-07-06
fix: recover fragmented words, detect stroke-encoded table rules, order rotated-text pages (#811, #812, #813)

### LINKED ISSUE #811: [Bug]: Extracted words come apart in the middle ("module" becomes "m odu le")
### Bug Description

`extract_words` splits words that the PDF draws as **single string literals**, producing fragments like `m` / `odu` / `le` for the word `module`. The glyphs inside the literal get positioned with phantom inter-character gaps (~0.3em) that don't exist in the rendered output, and the word-boundary detector then splits at those gaps.

The document draws this line (page 195 of the attached manual):

```
[(control)-332.9(\(EA)73.9(TC\))-332.9(module)-332.9(\(vehicles)]TJ
```

`module` is one atom — no internal TJ offsets, no space glyphs (the document separates words purely with ~-333 thousandths-of-em TJ offsets). Rendering shows a normal `module`. But `extract_words` returns:

```
"m"   bbox=(367.81, 570.12, w=5.18,  h=9.98)  font=10.0
"odu" bbox=(375.94, 570.12, w=13.31, h=9.98)  font=10.0
"le"  bbox=(392.22, 570.12, w=5.08,  h=9.98)  font=10.0
```

Note the geometry: `m` ends at x=372.99 but `odu` starts at 375.94 — a 2.95pt gap at 10pt font *inside a single string literal*. A lowercase `m` at 10pt in this face should advance ~7pt+, but the reported ink width is 5.18pt, so successive glyph positions appear to be accumulating more than the width used for the bbox — the difference shows up as phantom gaps at certain glyph boundaries, and any gap-based boundary check then fires mid-word.

The same page contains many affected words, always splitting the same way: `on ly)`, `W indow`, `s igna ls`, `Ad justab le`, `H eated`, `D eck lid`. Real inter-word separation on the same lines is only marginally larger (TJ -332.9 ⇒ ~3.3pt), so downstream consumers can't repair this from geometry — the split must not happen in the first place.

### Steps to Reproduce

1. Download `cv-manual.pdf` (2011 Ford Crown Victoria owner's manual, 2.3 MB) from https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91 — sha256 `22b666e3a143567f54e5890216785f1720c3e2bda461749d2f4c09956de382eb`.
2. Extract words for page index 194 (PDF page 195).
3. Look for `module` — it comes back as three words `m`, `odu`, `le`.

### Expected Behavior

A word drawn as one string literal with no internal positioning operators and no space glyphs is one `Word`: `module`.

### Actual Behavior

Three `Word`s — `m`, `odu`, `le` — with ~2.9pt gaps between their bboxes that don't exist in the rendered page.

### Environment

- `pdf_oxide` **0.3.73** (also reproduced on 0.3.70 and 0.3.72), Rust crate API
- macOS arm64 and Linux x86_64, rustc 1.95

### Code Sample

```rust
let bytes = std::fs::read("cv-manual.pdf")?;
let doc = pdf_oxide::PdfDocument::from_bytes(bytes)?;
let words = doc.extract_words(194)?; // PDF page 195
for w in &words {
    if ["m", "odu", "le"].contains(&w.text.as_str()) {
        println!("{:?} bbox=({:.2},{:.2} w={:.2}) font={:.1}",
            w.text, w.bbox.x, w.bbox.y, w.bbox.width, w.avg_font_size);
    }
}
```

### Error Messages / Logs

No error — silently wrong segmentation. Word dump on page index 194:

```
"m"   bbox=(367.81, 570.12, w=5.18)  ← gap to next: 2.95pt
"odu" bbox=(375.94, 570.12, w=13.31) ← gap to next: 2.97pt
"le"  bbox=(392.22, 570.12, w=5.08)
```

versus the true content stream for that run:

```
[(control)-332.9(\(EA)73.9(TC\))-332.9(module)-332.9(\(vehicles)]TJ
```

### Additional Context

- The font is an embedded subset (print-era FrameMaker-style typesetting). The phantom gaps look like the per-glyph advance used during char positioning disagrees with the width used for the char/word bboxes — either a misread `/Widths` array, a `MissingWidth` default, or kerning applied to positions but not widths. The split points are deterministic per word (`module` always breaks `m|odu|le`).
- `extract_spans` / plain-text extraction returns `module` intact on the same page, so the span path and the word path disagree about the same glyph run — whichever metric the span path uses is the correct one.
- Consequence downstream: table extraction (which consumes `extract_words` via `extract_tables_with_config`-style pipelines) emits fragmented cell text — `Electronic automatic temperature control (EATC) m odu le (vehicles equipped with EATC on ly)` — which breaks exact-match search and tokenization over table content. Documents that separate words with TJ offsets only (no space glyphs) and embedded subset fonts are the affected class.
- Happy to test a fix against the full manual; the fuse-chart pages (195–200) contain all ten affected words.

### LINKED ISSUE #812: [Bug]: Tables with visible grid lines aren't detected in some print-era PDFs
### Bug Description

`PathContent.bbox` covers only the path's geometric points, ignoring `stroke_width` — so a path's bbox can be radically smaller than what the renderer draws. The extreme (and real) case: print-era generators draw a table's vertical rule as a ~1pt horizontal segment stroked with a line width equal to the table height:

```
430.05 w
q 1 0 0 1 236.154 402.163 cm
0 0 m
.998 0 l
S
```

Rendered: a 1×430pt vertical bar. Extracted: `bbox = (236.2, 402.2, w=1.0, h=0.0)` with `stroke_width = 430.05`. Every bbox consumer — including `is_table_primitive()` and `detect_tables_with_lines` — sees a speck, so the table's vertical rulings are invisible and the whole grid goes undetected (this was the root cause of xberg-io/xberg#1213; we ship a downstream workaround that re-derives rendered geometry from `stroke_width`, but the right home for render-true geometry is the extractor).

### Steps to Reproduce

1. Download `cv-manual.pdf` from https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91 (2011 Crown Victoria owner's manual; sha256 `22b666e3a143567f54e5890216785f1720c3e2bda461749d2f4c09956de382eb`).
2. `extract_paths(197)` (PDF page 198).
3. Observe four paths with `bbox` ≈ `(x, 402, 1.0, 0.0)` and `stroke_width = 430.05` — the four column rules of a fully ruled 34-row fuse chart.

### Expected Behavior

Either `bbox` reflects rendered extents (geometric bbox inflated by the stroke, at least for stroked paths), or `PathContent` exposes a `rendered_bbox()` helper so consumers like `is_table_primitive` / the spatial table detector can classify by what the reader sees.

### Actual Behavior

`bbox` is geometry-only; `extract_tables_with_config` misses fully ruled tables whose rules are stroke-width-encoded, and their text extracts column-major (row associations destroyed).

### Environment

- `pdf_oxide` 0.3.70–0.3.73, Rust crate API, macOS arm64 / Linux x86_64

### Code Sample

```rust
let doc = pdf_oxide::PdfDocument::from_bytes(std::fs::read("cv-manual.pdf")?)?;
for p in doc.extract_paths(197)? {
    if p.stroke_width > 100.0 {
        println!("bbox=({:.1},{:.1},{:.1},{:.1}) sw={:.1} prim={}",
            p.bbox.x, p.bbox.y, p.bbox.width, p.bbox.height,
            p.stroke_width, p.is_table_primitive()); // prim=false for real rules
    }
}
```

### Additional Context

Butt-cap stroking extends perpendicular to the segment by `stroke_width/2` each side; a conservative render-true bbox for line-like paths is the geometric bbox inflated on the minor axis. The affected document class is FrameMaker/Quark-era print PDFs — the same lineage as the word-splitting report (#811).

### LINKED ISSUE #813: [Bug]: Pages with sideways (rotated) text extract as scrambled text, and words don't report their rotation
### Bug Description

Pages whose text is rotated 90° (landscape tables typeset on portrait pages) extract as interleaved word salad: `extract_words`/text extraction returns the spans in portrait reading order and exposes **no rotation metadata on words** (`Word`/`TextBlock` has no rotation field; `TextSpan.rotation_degrees` exists but doesn't reach the word path). Consumers can neither read the text in the right order nor detect that they should handle the page differently.

Concrete case: the scheduled-maintenance charts in the manual below (PDF pages 312–313) — fully ruled grids rotated 90°, service items reading bottom-to-top. Extracted text interleaves fragments of every row:

```
Whichever * if and tread abnormal (if brake for for lubricate level suspension,
measure drums, shields dealer fluid (recommended) concentration and joints, ...
```

Each fragment is internally coherent (span text is fine); only the ordering/grouping runs in the wrong frame.

### Steps to Reproduce

1. Download `cv-manual.pdf` from https://gist.github.com/tobocop2/0d30905e2375e09b153ce50296316f91.
2. Extract words/text for page index 311 (PDF page 312).
3. Observe portrait-order interleaving of the rotated rows, and no rotation info on the returned words.

### Expected Behavior

Either (a) dominant-rotation pages are detected and reading order runs in the rotated frame, or (b) at minimum, words expose their rotation (e.g. `rotation_degrees` propagated from spans) so consumers can transform coordinates and re-run ordering themselves.

### Actual Behavior

Portrait-order word salad, no rotation metadata on the word type — the information needed to recover is dropped inside the library.

### Environment

- `pdf_oxide` 0.3.70–0.3.73, Rust crate API, macOS arm64 / Linux x86_64

### Additional Context

Affected class: any landscape table on a portrait page (financial appendices, wide data tables, maintenance charts). Option (b) is a small, non-breaking addition (populate an existing-typed field during word assembly) and unblocks downstream handling even before (a) exists.

tobocop2:
## Description

Three separate bugs, all hitting the same family of real-world documents — print-era PDFs generated by FrameMaker/Quark-style tools (owner's manuals, spec sheets, financial appendices). Each was reproduced on a 334-page owner's manual (the repro attached to the issues) and is pinned by a synthetic in-code regression test.

**Words came apart in the middle (#811).** Text extraction returned fragments like `m` / `odu` / `le` for a word the PDF draws as one literal (`module`), silently corrupting search, tokenization, and table cells. Each glyph's accurate position was being paired with the *previous* glyph's width whenever a synthetic space entered a merged run out of order, opening phantom gaps inside words that the word splitter then honored. Widths now stay aligned to their glyphs no matter how runs merge, and documents that were fine before stay byte-identical (see the fixture sweep below).

**Tables with visible grid lines went undetected (#812).** These tools draw a table's vertical grid lines in an unusual way — a dot-sized line segment whose stroke is as wide as the table is tall. A person sees a normal full-height line; extraction saw only the dot, so whole tables were never detected and their text came out column-by-column with the rows scrambled. Paths now expose the extents a reader actually sees (`rendered_bbox()`), and everything that classifies rulings judges the drawn line.

**Sideways pages read as scrambled text (#813).** A landscape table typeset on a portrait page (rotated text, no `/Rotate` key) extracted with every row interleaved into gibberish, and nothing told the caller the text was rotated. Predominantly-rotated pages now read in the orientation a person turns the page to (coordinates still reported in true page space), and every word carries its rotation. Lone rotated stamps no longer disturb the main flow; the `/Rotate`-page handling from #804 is untouched.

The two new accessors are threaded through every language binding (go, ruby, php, swift, csharp, dart, elixir, zig, julia, r, objc, cpp, node, and the JVM Java/Kotlin/Scala/Clojure), matching how `sequence` was surfaced across the bindings in v0.3.70.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Tests
- [ ] CI/CD changes

## Related Issues

Fixes #811
Fixes #812
Fixes #813

## Changes Made

- **#811** — TJ-offset space spans (ISO 32000-1 §9.4.4) carry their advance from creation, and the span merge maintains the `char_widths`/text lockstep positionally so per-glyph decomposition never pairs a glyph with its neighbour's width.
- **#812** — `stroke_width` is CTM-scaled at extraction (§8.4.3.2, so it shares the geometry's coordinate space) and a new `PathContent::rendered_bbox()` exposes stroke-inflated extents (exact perpendicular + cap inflation for straight segments, conservative outset otherwise). Table detection, line orientation, the table-primitive classifier (with a 2:1 elongation gate so fat-cap dots don't qualify), line clustering, row/column separator checks, the ruled-table fallback's Y-banding, and every path region query judge rendered extents; geometric `bbox` is unchanged for all other consumers.
- **#813** — a dominant-rotation vote (mirroring the existing tategaki vote) orders a rotated page in its reading frame and restores coordinates afterward; minority rotated runs are ordered per rotation group and appended after the horizontal flow; ±90° runs no longer merge across rotated lines. `rotation_degrees` flows span → char → `Word`.
- New surfaces: `Word.rotation_degrees` and `PathContent::rendered_bbox()` on the Rust API; `pdf_oxide_word_get_rotation` / `pdf_oxide_path_get_rendered_bbox` C accessors (header regenerated with the CI-pinned cbindgen 0.29.2); `rotation_degrees` on `PyWord` and the WASM word JSON; `rendered_bbox` in the Python/WASM path dicts.
- Threaded both accessors through the out-of-crate language bindings (go, ruby, php, swift, csharp, dart, elixir, zig, julia, r, objc, cpp, node) and the JVM `TextWord` (Java + JNI), each following that binding's existing `sequence`/`bbox` pattern with api-coverage tests. Kotlin resolves the Java artifact `mavenLocal`-first so a freshly `mvn install`-ed build wins over the released one of the same version.

## Testing

- [x] I have run `cargo fmt`

Regression tests are synthetic in-code PDF builders (no committed binary fixtures), one per issue. Each fails on pre-fix `main` — reverting only `src/` and re-running makes the `#811`/`#813` word/order assertions and the `#812` table-detection assertion fail. `cargo test --all-features` is left unchecked because it is not a buildable configuration for this crate — `fips` and `legacy-crypto` are mutually exclusive (FIPS 140-3 forbids MD5), enforced by a `compile_error!` in `src/lib.rs`. The project's canonical `make check` (`cargo fmt --check` + `cargo clippy --features python -- -D warnings` + `cargo test --features python`) passes clean in one pass.

- [x] Python tests pass

## Documentation

- [x] I have added/updated examples (if applicable)

## Checklist

## Additional Notes

**Fixture regression sweep (this branch vs `main`, `extract_text` on every page of every committed text fixture):**

| Fixture | Result |
|---|---|
| `simple.pdf`, `hello_structure.pdf`, `outline.pdf`, `multi_column_table.pdf` | byte-identical |
| `1008.3918v2.pdf` (topology paper) | changed — #811 word-rejoins (`p`+`resent`→`present`, `exam`+`p`+`le`→`example`; word counts only ever drop, never rise) plus a pre-existing `decimal_merge` interaction on math subscripts: `P_{1,0}` renders `1.0` instead of `1 0`. That merge fires **identically on `main`** (28×/page 0); on `main` an inconsistent `char_widths` accidentally re-split it back, and the #811 lockstep fix removes that accidental re-split. Left as-is rather than retuning the `decimal_merge` gap window, which serves real split-box dollar amounts and can't be revalidated without the full corpus. |
| `1.pdf` (CJK financial) | changed — table rows regroup (表3 stock-position rows land together instead of scattered with blank lines); a reading-order improvement from the #811 cell grouping. |

**On the owner's-manual repro (`cv-manual.pdf`, full 333-page sweep):** fragmented words rejoin across ~270 pages; the p. 198 fuse chart detects as a 3-column, 34-row table (was undetected); the rotated maintenance charts (pp. 264–265, 312–313) read line-by-line with rotation reported on words.

**CI (full matrix run on my fork against this commit — live status table below):** every build, test, clippy, and lint job passes on Linux/macOS/Windows. The three dependency-advisory jobs that fail — `Security Audit` and `Dependency Check (cargo-deny)` in the CI workflow, and Ruby Bindings' OSV-Scanner — are not caused by this PR: they all flag the same [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) invalid-pointer-deref advisory in `crossbeam-epoch` 0.9.18, a transitive dependency via `rayon`. It is present identically on `main` (this PR makes no change to `Cargo.lock`) and clears with a dependency bump, which is out of scope for a text-extraction fix.

## Before / after on the real corpus documents

Both review-flagged articles are CC BY in the [PMC Open Access subset](https://www.ncbi.nlm.nih.gov/pmc/tools/openftlist/) (fetch with `scripts/fetch_real_fixtures.py`). Each GIF alternates the `to_html` output at the pre-fix commit vs `31e97a42` — red is before (tables lost, rows flattened into prose), green is after.

**PMC8103274 — Table 4 restored, plus a second table `main` missed:**

PMC8103274 before/after

**PMC8025823 — complications table restored (identical to `main`):**

PMC8025823 before/after


yfedoseev:
@tobocop2 I will review them tomorrow and run the changes of our regression set, thanks for your great work. Triggerred CI CD checks now

tobocop2:
thank you!

yfedoseev (changes requested):
Reviewed this against the PDF spec and ran it through our internal corpus regression sweep (extract_text/to_markdown_all/to_html_all across a broad real-world PDF set, human-read every diff, not just an automated score). All 3 of this PR's own new integration tests pass, and the spec grounding is solid — I verified the rotated-frame coordinate mapping algebraically for all three quadrants (90/180/270) and it round-trips exactly, and the CTM stroke-width scaling (`sqrt(|det(CTM)|)`) is the standard renderer approximation and correctly cited to §8.4.3.2.

That said, the regression sweep surfaced two issues that don't show up in this PR's own synthetic tests but are real on real-world documents. I'd want both addressed (or at least understood) before merge.

### 1. The #811 fix makes the exact bug #816/#817 describes *worse*, not better

Condition: two adjacent spans of 1-2 ASCII digits each, on the same line/MCID, where the true geometric gap between them sits in roughly the 0.8x-1.7x font-size range — i.e. exactly the band that's ambiguous between "two halves of a split-box decimal amount" and "two items in a comma/space-separated numeric list or subscript index" (e.g. typeset trace/state names like `T_16`, or index lists like `i_2, i_4, i_5, i_7`).

Before this PR, on several real documents, that ambiguous case happened to fall outside the (buggy) computed gap and stayed un-merged by luck — the old `char_widths` bug this PR fixes was accidentally producing a slightly-wrong gap that kept some of these pairs apart. After the `char_widths` positional-lockstep fix, the *now-correct* gap computation newly qualifies for the existing `decimal_merge` heuristic in more cases, so text that used to render correctly (if accidentally) now gets a fabricated decimal point where none exists in the source: a token like "45" immediately followed by "16" as a new sentence can render as "45.16" (glued into one fake decimal, worse than before), and short numeric list/index items like "5" and "7" get glued into "5.7".

I checked this against #816/#817's fix (tightened decimal-merge ceiling from 2.0x to 1.3x font size) merged together with this branch — it's a clean, non-conflicting textual merge — and confirmed **the combination only partially resolves this**. Wider-gap cases (roughly 1.3x-2.0x, e.g. the sentence-boundary case above) are fixed by #817's tighter ceiling. But narrower-gap cases (comma/space-separated list or subscript-index digit pairs sitting at a gap comparable to a genuine split-box amount, under ~1.3x) are geometrically indistinguishable from a real decimal amount and are **still wrongly merged even with both PRs applied together**.

This isn't fixable by adjusting the gap-ceiling threshold alone, since the ambiguous band is real — a genuine split-box amount and a subscript/list-index pair can sit at literally the same gap-to-font-size ratio. It probably needs a different signal than gap-vs-font-size alone (e.g., checking whether the two spans are separated by a `,`/space glyph already present in a nearby span, an MCID/structure boundary, or a preceding subscript/`Ts`-flag context) before this PR's more-accurate width computation ships.

### 2. Real academic "three-line" (booktabs-style) tables lose their structure entirely

Condition: a table whose only ruling is a small number (commonly 2-3) of full-width **horizontal** separator rules — a heavy top rule, a thin sub-header rule, a heavy bottom rule — each drawn as a geometrically **zero-height** straight-line stroke (`bbox.height == 0.0` exactly), with light stroke widths (I measured ~0.3pt and ~0.8pt on two different real documents showing this failure), and critically **no vertical column-divider rules at all**. This is the standard "booktabs"/three-line table style common in academic PDF templates (LaTeX `booktabs`, many journal styles).

On two independent real documents with this exact table style, the table was correctly detected and rendered as an HTML `<table>` before this PR, and completely lost after — the cell text scatters into disconnected paragraphs with row labels separated from their values (e.g. a 6-row logistic-regression table's Beta/SE/OR/p-Value columns end up as orphaned paragraph blobs elsewhere in the document, no longer aligned to their row labels).

I checked the individual rule geometry against this PR's new classifiers directly (`rendered_bbox`, `is_horizontal_line`, `is_table_primitive`) and each rule individually still classifies correctly as a horizontal table-primitive line under the new logic — so the regression isn't in the per-line classifier itself. It's more likely in the row/column grid-formation or line-clustering step (`group_lines_into_clusters` / `detect_tables_in_cluster`) not handling a table that has zero vertical rules and only 2-3 horizontal ones well once it's reasoning over `rendered_bbox` instead of raw `bbox` for clustering/banding — possibly other nearby unrelated decorative strokes on the same page get pulled into the same cluster under the rendered-extent-based grouping, or the "vertical-rule-derived column count" path silently fails when v_xs is empty and this table used to be caught by some other fallback that's now perturbed. I haven't fully root-caused it at the Rust level, but the condition to reproduce is precise: build (or find) a table with only 2-3 zero-geometric-height horizontal rules and no vertical rules, and check it still resolves into a `<table>` after this branch.

Given both of these are real, reproducible regressions on real-world documents that the PR's own synthetic tests don't currently cover, I'd suggest adding regression tests for both conditions above (synthetic in-code PDFs, per this repo's convention) before merge. Happy to re-run the corpus sweep once there's an update.

tobocop2:
@yfedoseev Both addressed in 688ddb7f — regression tests included for each, per the repo's synthetic in-code PDF convention.

**1. Fabricated decimals (`P₁,₀` → `P1.0`)** — took your intervening-glyph suggestion: the decimal merge now checks the gap for ink across all glyph runs, so the subscript comma (drawn out of content-stream order) keeps the digits separate, while a genuine split-box amount — empty gap — still merges. On the in-repo topology paper (`1008.3918v2.pdf`) this alone takes the decimal merges from 28 to 0, and it composes with #817: the ink signal covers the narrow band, the 1.3× ceiling the wide band.

**2. Booktabs three-line tables** — root cause: a short decorative segment with a heavy stroke (tick/list dash), which `main` ignored entirely, now classified as a vertical primitive under rendered extents, landed in `v_edges`, and the `v_edges.is_empty()` gate disabled the horizontal-rule fallback for the whole page. Fix: a stroke-promoted vertical (geometric speck) only counts as ruling evidence when its painted bar crosses one of the horizontal rules — which a genuine stroke-encoded column rule always does. Geometric verticals and rectangles keep their pre-existing veto unchanged; reproduced both directions (main detects, branch lost it) before fixing.

Full suite is green locally (`make check`, 333 suites). A corpus re-run when you have a chance would be appreciated.

yfedoseev (changes requested):
Re-ran the corpus regression sweep against `688ddb7f`. Good news first: both digit-pair-merge fixes verify cleanly (the "intervening ink" check correctly separates subscript index pairs while still merging genuine split-box amounts), and your two new unit tests pass. But the sweep surfaced two more issues from the table-detection change — one new, one that your fix doesn't actually reach. I don't have a way to hand you the source PDFs (all third-party, can't redistribute), so below is the exact geometry needed to reproduce both without our corpus.

## 1. New regression: scattered fraction bars misdetected as a table (severe)

On a dense-equation math paper, the page draws 27 stroked paths — every one a **zero-height horizontal stroke** (a fraction vinculum), at 12 different, unrelated y-positions spread across the whole page, widths ranging ~11-279pt, stroke width a consistent 0.478pt, **no vertical rules anywhere**. That's the identical geometric signature as a booktabs table (several zero-height horizontal strokes, no verticals) — except these bars belong to unrelated equations at different heights on the page, not rows of one table.

Before this commit, this exact document was byte-for-byte nearly identical to `main` (word_jaccard 0.9766, -2 words — the horizontal-rule fallback wasn't firing on it at all). After this commit it drops to **word_jaccard 0.5652, -32 words**, with large chunks of equation text duplicated and split into disconnected floating fragments — the horizontal-rule-bounded fallback is now grouping unrelated fraction bars from different equations into a fake table and reflowing the text into fake rows, badly scrambling top-to-bottom reading order.

**Suggested repro**, in the style of your `booktabs_fixture_pdf`: build a page with ~6 short strokes (10-60pt wide, 0.478pt stroke) at clearly different x AND y positions spanning most of the page height (e.g. y = 700, 650, 600, 550, 500, 450), each with ordinary body text immediately above/below it (not table-cell text), and assert the extracted text preserves natural top-to-bottom document order with no `<table>` inserted — this is the inverse case of your existing fixture: rules that must **not** be treated as a table because they aren't spatially/dimensionally coherent as one.

The likely fix shape: the horizontal-rule-bounded fallback needs a coherence check before forming a table from a set of horizontal rules — e.g. requiring the rules span a comparable, overlapping x-range (a real table's rules line up), and/or aren't spread across most of the page's y-extent. A scattered, unrelated set failing that check should fall through to ordinary paragraph flow instead of forming a table.

## 2. The two original regressions (PMC8103274, PMC8025823) are still broken — this commit doesn't reach them

This is the more concerning one: I diffed both documents again against `688ddb7f` and got **byte-for-byte the same diff as before this fix** — same `dword`, same missing `<table>`, same scattered cells. Your fix didn't change their output at all, which means whatever's actually breaking these two real documents isn't the decorative-speck-in-`v_edges` mechanism this commit addresses.

Here's the real geometry from both (extracted via `extract_paths()`), so you can reconstruct the page composition exactly:

**PMC8103274, the broken table's own rules** (a clean 6-row logistic-regression table, "Table 4."):
```
bbox=(167.26, 663.29, 392.01, 0.00)  stroke_width=0.797   <- heavy top rule
bbox=(167.26, 647.28, 392.01, 0.00)  stroke_width=0.299   <- thin sub-header rule
bbox=(167.26, 576.48, 392.01, 0.00)  stroke_width=0.797   <- heavy bottom rule
```
By itself, this is exactly your `booktabs_fixture_pdf` shape and (per your own passing test) should detect fine. But the same page also carries several **unrelated** decorative elements elsewhere — an author-affiliation box with dashed borders, built from alternating short dash segments and tiny near-square "joint" strokes (from horizontal/vertical border overlap), e.g.:
```
bbox=(-24.52, 795.92, 198.00, 0.96)   stroke_width=1.0    <- dash segment
bbox=(173.48, 795.92, 0.96, 0.96)     stroke_width=1.0    <- near-square joint speck
bbox=(174.44, 795.92, 76.92, 0.96)    stroke_width=1.0    <- dash segment
bbox=(251.36, 795.92, 0.96, 0.96)     stroke_width=1.0    <- joint speck
... (repeats at y=781.94, 699.98, 253.70, 239.72, 225.68, 211.70, 170.24, each with the
     same dash+speck alternation pattern, all well outside Table 4's y=576-663 band)
```
**PMC8025823** is the same story, smaller table ("Table 3.", 4 rows):
```
bbox=(167.26, 346.24, 392.01, 0.00)  stroke_width=0.797
bbox=(167.26, 330.23, 392.01, 0.00)  stroke_width=0.299
bbox=(167.26, 270.38, 392.01, 0.00)  stroke_width=0.797
```
— plus this page also has a **second** set of 392.01pt-wide zero-height rules further down (`y=186.93`, `170.92`, `122.04`), i.e. two separate booktabs-style tables stacked on one page, plus the same kind of dash-bordered box (`y=336.71/322.73/254.19`, widths ~284pt and ~109pt) elsewhere.

My guess (not verified against the actual detector code) is that `group_lines_into_clusters`'s spatial clustering (`EXPANSION=3.0`) is pulling these unrelated dash/speck elements into the same cluster as the real table rules, and something downstream of that — not the `v_edges`/decorative-speck path this commit fixes — is what actually breaks. Whatever it is, it's a distinct bug from what this commit addresses, since the output is provably unchanged.

**Suggested repro**: extend your `booktabs_fixture_pdf_with_decorations` — instead of one 8pt-stroke speck far from the table, add a cluster of *several* dash-segment + tiny-square-joint pairs (mirroring the coordinates above, 1.0pt stroke, ~0.96pt squares) positioned in a band *outside* the table's own y-range but on the same page, and a *second* booktabs table lower on the page — then assert the *first* table still resolves to `<table>`.

## Process note

We're now two rounds in where a targeted fix for a synthetic repro didn't fully cover the real-world case it was aimed at, plus introduced a new regression in the same area. Since I can't hand you the actual source PDFs (all third-party), I've tried to give you exact reproducible geometry above instead — hopefully that closes the loop without needing repeated round-trips through our corpus. Happy to re-run the sweep again once there's an update; just flagging that a synthetic-fixture-only test strategy is proving to have real gaps for this particular code path (line clustering interacting with unrelated nearby geometry), so a couple of "messier page" fixtures like the ones above alongside the clean single-table cases would probably catch the next one before it ships.

tobocop2:
@yfedoseev Both fixed in 31e97a42, validated on the real documents this time.

**Key find:** both failing articles are in the [PubMed Central Open Access subset](https://www.ncbi.nlm.nih.gov/pmc/tools/openftlist/) — the portion of PMC published under Creative Commons licenses that explicitly allows redistribution and reuse (unlike regular PMC content). Both are [CC BY](https://creativecommons.org/licenses/by/4.0/): [PMC8103274](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8103274/) (Tomography 2021) and [PMC8025823](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8025823/). So the exact documents your sweep flagged are freely fetchable and the tests no longer depend on your corpus. New files:

- [`scripts/fetch_real_fixtures.py`](https://github.com/tobocop2/pdf_oxide/blob/31e97a42/scripts/fetch_real_fixtures.py) — downloads both articles into the gitignored `tests/fixtures/real/` (no binaries in the repo)
- [`tests/test_pmc_real_document_tables.rs`](https://github.com/tobocop2/pdf_oxide/blob/31e97a42/tests/test_pmc_real_document_tables.rs) — opt-in tests on the real files (v054 style: skip with a log line when absent), pinning the exact tables at `to_html` level; path geometry matches your dump exactly
- [`tests/test_booktabs_table_detection.rs`](https://github.com/tobocop2/pdf_oxide/blob/31e97a42/tests/test_booktabs_table_detection.rs) — the synthetic messy-page fixtures (fraction bars, dash boxes, stacked tables), each verified failing before the fix

**Root causes found by tracing the real files — three, not one** (all in [`spatial_table_detector.rs`](https://github.com/tobocop2/pdf_oxide/blob/31e97a42/src/structure/spatial_table_detector.rs)):

| # | Mechanism | Fix |
|---|---|---|
| 1 | Fraction bars at unrelated x/y pair up in the fallback's adjacent-y walk → fake tables | Wide h-edges grouped into x-range-coherent families (overlap/union ≥ 0.85); pairing only within a family |
| 2 | Dash boxes' short **vertical** dash runs veto the fallback page-wide | Vertical ruling evidence now = crossing ≥ 2 *distinct* rule levels (border dashes cross ≤ 1; joint specks no longer double-count one dashed line's collinear edges) |
| 3 | Dash lines interleave the table's rule ys → its own rules never pair; and a box band over a table band detects the same rows twice | Families fix the pairing; overlapping detections deduped, tighter one wins |

Two notes on your suggestions:
- **Repro**: the fixture uses your *measured* geometry (14 bars, 11–279pt wide, 0.478pt stroke, several bars per y) rather than the sketched 10–60pt — bars under 100pt can't trip the fallback at all (it only considers wide edges), so the short-bars-only version passes everywhere and proves nothing. The faithful version reproduced 3 fake tables + scrambled order before the fix.
- **Your second signal ("and/or y-extent") — concern confirmed, implemented as containment instead** (396fe150): I built the shape x-coherence misses — an aligned multi-step derivation, where consecutive steps' vinculums line up at the relation sign with near-identical widths — and it *did* still form a fake 10-row table. So the extra signal is needed. But a page-y-extent cap can't be it: a legitimate full-page table's rules also span most of the page, so extent can't separate the two. What does separate them is **containment**: a table's rules span the rows they bound, while a fraction bar floats inside math that continues past it on both sides (lhs symbols, equation numbers). A rule pair now only forms a region when >70% of the band's text width sits inside the rules' shared x-range. The [aligned-derivation fixture](https://github.com/tobocop2/pdf_oxide/blob/396fe150/tests/test_booktabs_table_detection.rs) fails without this check and passes with it; every prior fixture, the real-document tests, and the sweep stay green.

Note: the fraction-bar coherence gap turns out to be a pre-existing latent bug — my fixture fails on `main` too; your math paper was shielded on `main` only by the accidental vertical-evidence veto. **This PR fixes it for `main`'s latent case as well** (the family-grouping fix, row 1 in the table above), and the fixture guards it from here on.

**Before / after on the real documents** (red = pre-fix `to_html` output, green = `31e97a42`; also embedded in the PR description):

PMC8103274 before/after

PMC8025823 before/after

**Results:**
- PMC8025823: identical to `main` (complications table back)
- PMC8103274: `main`'s Table 4 back, plus the page's second table (which `main` missed) — once, no duplicates
- Sweep over all 27 in-repo PDFs vs pre-fix: 26 byte-identical, 1 improved (CJK note: 0 tokens lost, 8 recovered)
- 333-page manual: byte-identical
- Full suite green

Happy for another corpus sweep. I also updated my PR description with visual evidence:

yfedoseev (commented):
Re-ran everything against `31e97a42`, plus extended the test set well beyond the original 419-doc corpus to cover the other surfaces this branch touches — 145 additional documents total, built specifically to probe each mechanism separately, all diffed against a fresh `main` build (not cached results).

## Confirmed fixed (independently re-verified)

- **PMC8025823** — byte-identical to `main` again. Fully resolved.
- **PMC8103274** — original table restored, and the claimed bonus is real: a second table that was broken even on `main` (pre-existing, not something this branch caused) is now correctly recovered too.

## Still broken, unchanged from the last report

- **`2503.09472_pg15.pdf`** (the dense-equation fraction-bar page) — byte-for-byte identical to before this commit. Same severity (word_jaccard ~0.44-0.66, -32 words, duplicated/disconnected fragments). Re-verified via a completely independent manual extraction, not a stale artifact.

## Extended test set — how widespread is the remaining issue, and are there any new ones?

Two batches, both diffed `main` vs `31e97a42`:

**Batch 1 — 50 more pages sharing the fraction-bar geometric signature** (≥6 scattered zero-height horizontal strokes, no verticals, spread across >150pt of the page), pulled from the other ~520 arxiv-math pages in the corpus, not just the one already reported: **23 changed, none anywhere near the severity of the known case** (worst was word_jaccard 0.81, -28 words — a moderate reflow, not scrambling). So the severe failure mode looks isolated to that one document's specific layout, not a general property of "page has scattered fraction bars" — there's some more specific trigger condition I haven't pinned down.

**Batch 2 — 95 pages covering the *other* surfaces this branch touches**, sampled broadly across ~1,180 documents (pdf_benches, pdfjs test suite, pdfbox test suite): 44 pages with real two-axis table grids (horizontal + vertical rules), 23 more booktabs-style tables, 3 more dash-bordered decoration boxes (the PMC affiliation-box pattern), and 25 pages with rotated text (probing #813, since this branch touches shared code there too). Result: **only 8 of 90 valid extractions changed at all**, and of those:
  - 1 genuine bonus improvement (`PMC8154847` — a validation-studies table that was fragmented into ~300 disconnected one-word lines on `main` now reflows into coherent rows).
  - The already-reported `PMC8129076` minor false-positive (mild, no content loss, unchanged in severity).
  - The remaining 6 are all word_jaccard ≥0.95, dword within ±6 — noise-level.
  - **Zero new regressions** among the 25 rotation candidates, the 44 real-grid tables, or the 23 additional booktabs candidates — all either byte-identical to `main` or changed only within noise.

So: the fix is solid and broad — real tables, other booktabs pages, dash-boxes, and rotated-text pages all check out clean. The one open item is still just `2503.09472_pg15.pdf`, and it looks like a narrower trigger than "scattered fraction bars" in general — happy to dig into what's specific about that page's geometry if it'd help, or you're welcome to since I don't have a faster way to bisect it from the outside.

tobocop2:
@yfedoseev Got the open item — found the actual document, traced it end to end, and the trigger you couldn't pin down is identified. It isn't tables.

Heads-up first: your sweep ran against 31e97a42; the branch has moved since — latest is [`ae46f520`](https://github.com/tobocop2/pdf_oxide/commit/ae46f520) ([`396fe150`](https://github.com/tobocop2/pdf_oxide/commit/396fe150) added a containment guard for x-coherent bar sets, `ae46f520` hardened it). Worth re-running against that.

**The file:** arXiv:2503.09472v1, page 15 (1-based) — 27 zero-height bars at 13 y-levels, matching your dump.
```
curl -sL -o 2503.09472.pdf https://arxiv.org/pdf/2503.09472v1   # page index 14
```

**Not tables:** at HEAD, zero tables form on that page in any pipeline (gate open, no fallback regions, nothing from intersections/clusters). The fake-table mechanism is fully closed by 396fe150.

**Root cause — file-level bisect** (main src + one file swapped, your metric):

| src state | word_jaccard vs main |
|---|---|
| `text.rs` @ 65b71390 (pre decimal-ink fix) | **0.9766** — your "before" |
| `text.rs` @ 688ddb7f (decimal-ink fix) | **0.5652** — your "after" |

The trigger is the #816/#817 decimal-merge ink check, not the table work. This page is dense with subscript digit pairs sitting in the merge band — it's the page where #816's bug fired hardest on `main`.

**Is the new output actually worse?** Independent reference (poppler `pdftotext`, same page):

| | vs poppler | fabricated decimals in output |
|---|---|---|
| `main` | 0.447 | **10** — `1.2`, `11.11`, `12.10`, `20.20`, `11.20`, … |
| this branch | 0.401 | **0** |

Statistically the same distance from ground truth (every extractor struggles here — poppler's own output is scrambled fragments), but the branch invents zero tokens while `main` invents ten. Most of the −32 dwords are those fabrications un-merging (`b`+`11` clumps re-forming as the real `b11`). word_jaccard-vs-main penalizes exactly the corrections — which is why this one page scores far below your other 50 fraction-bar pages.

**Residue — traced and filed separately:** the `=dt` glue is **pre-existing on `main`** — main's own output for this page already contains `dx 3 =dt` / `dx 4 =dt` (2 occurrences; the branch surfaces 2 more because the changed span topology routes more fractions down the same path). The mechanism lives in display-math line composition (`=` at fraction mid-height y-overlaps the denominator; negative x-gap fuses them), code this branch doesn't touch. Filed with exact span geometry and repro as #830 — now fixed in #831 (with before/after on this exact page).

Also in ae46f520: containment guard made division-free (no NaN branch on degenerate bands).

yfedoseev (changes requested):
Did a full exhaustive read of every remaining diff in the 419-doc sweep against `ae46f520` (not just the previously-flagged cases) — confirms your fraction-bar/PMC fixes are solid, but surfaces two new issues, one of them significant.

## Confirmed still good
- `2503.09472_pg15.pdf`'s diff is unchanged (as expected — 396fe150/ae46f520 only touch table code, not decimal-merge), and I independently re-verified your bisect: the branch's output has zero decimal-looking tokens on that page vs 18 in `main`'s. Your analysis holds.
- PMC8103274/PMC8025823 unaffected by this round, still fixed.

## New issue 1 (significant) — monospace code/hex-dump listings misdetected as tables

Reproducible across 4 documents (technical zines with source-code and hexdump listings, letter-spaced monospace text — think PoC||GTFO-style writing, but this'll hit any PDF with code blocks). Example, a C signal handler that `main` renders correctly as one flowing paragraph:

```
main:  <p>1 void s e g f a u l t _ s i g a c t i o n ( int s i g n a l , ... }</p>

cand:  <p>1</p>
       <table><tbody>
       <tr><td>void s e g</td><td>f a u l t _ s i g</td><td>a c t i o n ( int</td>...</tr>
       <tr><td>( (</td><td>ucontext_t ∗)</td>...</tr>
       </tbody></table>
       <p>3 }</p>
```

Words are split mid-token across `<td>` boundaries (`f a u l t _ s i g` / `a c t i o n ( int`) — this isn't a cosmetic reflow, it's breaking the actual tokens. Same pattern hits a hexdump table elsewhere in the same document (existing table gets re-split into more, worse-aligned columns) and 3 more zines with similar code/asm/bit-diagram content. My guess: the same booktabs/grid detection logic that now (correctly) catches real data tables is also firing on letter-spaced monospace text, where the per-character x-gaps between spaced-out glyphs look like column boundaries.

I'd treat this as blocking or needing a fast follow-up — it's not an edge case, it's "any PDF containing a code listing," which is common enough that I don't want to sign off with this live.

## New issue 2 (minor) — decimal-merge glues unrelated table values

```
main: C (dyn−1 · cm5) 4 · 10−4 3 · 10−4 3 · 10−4 4 · 10−4
cand: C (dyn−1 · cm5) 4 · 10−4.3 · 10−4.3 · 10−4.4 · 10−4
```
Four independent scientific-notation values in one table row get fused into a nonsensical string via fabricated decimal points — the opposite of what the intervening-ink check is supposed to prevent. Low severity (jaccard ~0.99) but worth fixing alongside #1 since it's the same mechanism class.

Both independently re-verified against the actual diff output, not just taking a report at face value. Otherwise the extended sweep (145 additional documents beyond the original 419, covering fraction-bar geometry broadly, real table grids, more booktabs/dashbox pages, and rotated text) came back clean — these two are the only new findings.

yfedoseev:
One more thing, separate from the technical findings above: we're now three rounds in where a fix that passed its own synthetic fixture (and, this round, a real 27-document sweep) still missed something on broader real-world content — first the two PMC tables, then the fraction-bar page, now these code-listing false-positives. Nothing against the fixes themselves, they've each been well-diagnosed and correctly scoped to what you could see — but I think we're past the point where synthetic fixtures + a curated document set are catching everything for this particular code path (line/rule clustering interacting with arbitrary page content).

Would you be able to test the next iteration against a broad, diverse PDF sample before pushing — ideally hundreds of real-world documents spanning different producers/content types (code listings, forms, scanned docs, multi-table pages, etc.), not just the specific repros for each bug found so far? Doesn't need to be our exact corpus (I know you don't have access to it) — any large public PDF set (arXiv dumps, a GitHub Actions artifact corpus, Common Crawl PDF sample, whatever's convenient for you) run through a quick jaccard-vs-previous-commit diff would likely surface this class of issue before I do. Happy to talk through what a lightweight version of that harness could look like if useful.

tobocop2:
> One more thing, separate from the technical findings above: we're now three rounds in where a fix that passed its own synthetic fixture (and, this round, a real 27-document sweep) still missed something on broader real-world content — first the two PMC tables, then the fraction-bar page, now these code-listing false-positives. Nothing against the fixes themselves, they've each been well-diagnosed and correctly scoped to what you could see — but I think we're past the point where synthetic fixtures + a curated document set are catching everything for this particular code path (line/rule clustering interacting with arbitrary page content).
> 
> Would you be able to test the next iteration against a broad, diverse PDF sample before pushing — ideally hundreds of real-world documents spanning different producers/content types (code listings, forms, scanned docs, multi-table pages, etc.), not just the specific repros for each bug found so far? Doesn't need to be our exact corpus (I know you don't have access to it) — any large public PDF set (arXiv dumps, a GitHub Actions artifact corpus, Common Crawl PDF sample, whatever's convenient for you) run through a quick jaccard-vs-previous-commit diff would likely surface this class of issue before I do. Happy to talk through what a lightweight version of that harness could look like if useful.

I'll take any feedback, otherwise I'll build something to make this more exhaustive given your above feedback. Thanks again for all the reviews and I apologize for the noise. I'll mark this as a draft for now.

tobocop2:
@yfedoseev Both findings fixed in 7487dda0. This round ran your suggested broad-corpus validation before pushing, and it caught a third, worse bug in the process.

**The harness** (your "lightweight version", I took the offer): [corpus_sweep.py + pinned manifest](https://gist.github.com/tobocop2/9cdecc49b4ef47899eb1562191166992). 569 public documents, every URL explicit: 193 arXiv (8 categories), 100 PMC OA (CC BY), ~200 govdocs1 (crawled .gov: scans, office exports, forms), 5 PoC||GTFO zines (code listings, polyglots), 60 pdf.js test-corpus PDFs, 10 IRS forms. `fetch` → `run` per build → `diff` emits per-doc jaccard / Δwords / Δ`<table>`, ranked.

**Fixes:**

| Finding | Root cause | Fix |
|---|---|---|
| Code/hexdump listings become tables | Zines frame listings with rules; letter-spaced monospace puts each glyph on a grid, so the rule-bounded fallback saw aligned single chars as columns | A band that is ⅓ single letters or ½ single characters is spread-out text and falls through to paragraph flow. Verified on a real zine: asm listing + `lspci` session no longer table-ify; its two genuine tables (hex diagram, hexdump) survive |
| Sci-notation values fuse (`4 · 10−4.3 · 10−4`) | Exponent digit + next value's mantissa digit are both pure-digit runs in the merge band | Decimal merge now requires both runs' font sizes to match within 15%: an exponent is markedly smaller; a real split-box amount prints both halves at one size |
| **New (sweep-caught): OOM kill, silent truncation** | `topological_block_order`'s iterative DFS re-pushes all unvisited predecessors per expansion (no on-stack marking), which is exponential on heavy fan-in; a dense-equation arXiv page (same class as 2503.09472; repro: [arXiv 2607.03406](https://arxiv.org/abs/2607.03406) p.6) grew ~15 GB and died, losing 13k words | Replaced with Kahn's algorithm: each block visited once, reading-order drain, unconditional termination. Page extracts in milliseconds; existing `topo_*` ordering tests unchanged |

**Sweep verdict** (`main` vs `7487dda0`, 569 docs): 334 byte-identical; all 235 changed docs audited, and every one lands in an intended class: fabricated decimals removed, `0 .0022`-style splits normalized to `0.0022`, side-by-side benchmark tables emitted sequentially with datum-complete content (spot-verified value-by-value), fraction-bar fake tables gone, booktabs tables recovered. Worst residuals are gauge-dial/figure-art token noise.

The zine, PMC, and arXiv regression documents are all opt-in tests in the branch now (fetch-and-skip pattern). Happy for your corpus to have the final word.

yfedoseev (approved):
Re-ran everything against `66aca902`. This round closes the review cycle out well — both regressions from the last round are fixed and generalize, and the OOM fix you found yourself is a solid catch.

## Both previously-reported regressions: confirmed fixed

- **Sci-notation decimal glue** (`2503.04092_pg10.pdf`): zero diff vs `main` now — byte-identical.
- **Code-listing-as-fake-table** (the `segfault_sigaction` case in `pocorgtfo08`): grepped the current diff, zero matches — confirmed gone. And it generalizes: `multi_column_...page_1_pg1.pdf` and `headers_footers_f00ddca2.../page_5.pdf` both show the same class of fake-table-to-clean-paragraph fix on documents outside your own repro set.

Bonus: `PMC8025752`/`PMC8103274` show real clinical tables getting reassembled from scattered orphan paragraphs — likely a side benefit of the block-order fix. `pocorgtfo08` also had a *severe* pre-existing duplication bug (main duplicated an entire hexdump into a garbled word-soup paragraph) that's now substantially reduced, though not fully clean.

12/12 unit tests pass. Full 419-corpus sweep: 149 total diffs, only 3 in the "large" bucket — all pre-existing/already-understood (the `2503.09472` decimal-ink case from earlier rounds, and one lateral case already classified as improvement).

(One red herring worth mentioning so it doesn't confuse anyone later: `war_peace.pdf` shows up as a "crash fixed" in the diff — it's not real. I hand-timed it against a clean `main` build and got ~62s either way; it's just a slow document sitting at our harness's 60s timeout, a pure flake. Same false-positive showed up identically on #829 and #831's sweeps too, which is what tipped me off — neither of those PRs touches this code path at all.)

## Two minor residuals — same bug class, much smaller scale, not blocking

1. `pocorgtfo03`: two hex-byte pairs in a genuine table glue into `AA 02.50`-style tokens. 2-token impact, one document.
2. `pocorgtfo04`: a chemical formula (H₂SO₄) has its two subscript digits fuse into "H SO42" (5 occurrences, minor readability hit, no data loss). Root cause is understandable — your font-size-match-within-15% heuristic doesn't catch this because both subscript digits are similarly sized *to each other*, even though they're not sequential digits of the same number.

Both are the same bug class as what you already fixed, just at 2-11 word impact instead of the -97/-243 seen last round — I'd suggest a fast-follow rather than holding this up further.

**Recommendation: good to merge.** Thanks for running the broad sweep yourself and catching the OOM issue independently — that's exactly the kind of testing I'd asked for, and it clearly paid off.

tobocop2:
thank you!!!

tobocop2:
so i'm happy to address the those findings or I'll leave it to you to merge, either way, I greatly appreciate you and thanks again.

yfedoseev:
Wanted to circle back with a proper thank-you now that this has landed cleanly: five review rounds is a lot to go through, and every single one of them got a real, substantive response instead of a defensive one — the round-4 findings (code-listing-as-table, sci-notation glue) were fixed cleanly *and* you went and ran your own 569-doc sweep on top of it, which is how the OOM bug got caught before it ever reached anyone else. That's the kind of testing discipline that makes a maintainer's job easy. Really solid work — thanks for sticking with it through all the back-and-forth.

yfedoseev:
Heads up — CI is currently failing on `66aca902`, and it's a real issue, not a flake. Clippy and Test (ubuntu-latest, stable) — both required checks — fail to compile `tests/test_booktabs_table_detection.rs` (the repo builds test code with `-D warnings`, so lint violations are hard errors). That single compile failure is also why Code Coverage, FIPS CryptoProvider, Rendering Tests, and Test on macos/windows all show red — they all depend on the same test binary building.

9 violations, all in the tests you added this round:

- `unused_mut` — `let mut doc = ...` where `doc` is never mutated:
  - line 130: `let mut doc = PdfDocument::from_bytes(booktabs_fixture_pdf())...`
  - line 619: `let mut doc = PdfDocument::from_bytes(std::fs::read(p)...)...`
- `clippy::field_reassign_with_default` — fields set individually after `T::default()` instead of struct-update syntax, at lines 98, 120, 225, 374, 399, 513, 590 (all `config.<field> = ...;` right after `let mut config = ...::default();`)

`cargo clippy --fix --tests --allow-dirty` should clear all of these in one pass, or it's a few minutes by hand (e.g. `let config = TableDetectionConfig { text_fallback: false, ..Default::default() };`). Once that's green I'll re-run the full sweep against the new commit.

tobocop2:
Sorry about that. 

@yfedoseev fixed in `30575df0`.

`cargo clippy --features python --all-targets -- -D warnings` is clean locally, and both test files pass (booktabs 12, pmc 3). Ready for the re-run whenever you are.

PR #828, opened 2026-07-07
Fix Hebrew visual-order extraction for per-glyph OCR text layers (fixes #826)

### LINKED ISSUE #826: Hebrew visual-order text layer extracted reversed (scanned OCR PDF, v0.3.73) - #657 fix does not fire
## Summary

On a scanned Hebrew PDF (Vilna-print Talmud, OCR text layer stored in **visual order**), `extract_text` in **v0.3.73** returns essentially all Hebrew text character-reversed (visual order preserved instead of being converted to logical order). PDFium extracts the same file in correct logical order, so the information needed to recover reading order is present in the file.

This looks like a case the #657 fix (v0.3.58–0.3.62) does not catch: the visual-order-run detection (`reverse_rtl_visual_order_runs`) apparently never fires for this document, so the visual-order text passes through unreversed.

## Reproduction

```rust
let mut doc = pdf_oxide::PdfDocument::open("beitza.pdf")?;
let text = doc.extract_text(5)?; // any page
```

Actual (page 5, first line):

```
תרוסמהם״ש הציב ןושאר קרפ הציב
```

Expected (what PDFium returns for the same page):

```
ביצה פרק ראשון ביצה מסורת הש״ם
```

(The actual output is the expected line with every Hebrew run character-reversed; note also the missing space in `תרוסמהם״ש`.)

## Quantification (whole document, 78 pages)

Word-frequency check over the full extracted text, pdf_oxide 0.3.73 vs PDFium (same file, same machine):

| word (logical) | reversed form | PDFium: logical / reversed | pdf_oxide: logical / reversed |
|---|---|---|---|
| ביצה (tractate name) | הציב | 281 / 0 | **0 / 246** |
| אמר | רמא | 2,519 / 15 | **24 / 2,313** |
| גמרא | ארמג | 57 / 0 | **0 / 52** |
| משנה | הנשמ | 62 / 3 | **3 / 54** |

Per-page unique-token containment vs the PDFium baseline: only **4.8%** of PDFium's Hebrew tokens appear in pdf_oxide's output for the same page (a correct extraction would be near 100%). Total Hebrew letter counts are nearly identical (553,293 vs 553,311), i.e. all content is extracted — just in reversed order.

Extraction was clean otherwise: 78/78 pages, no errors, no panics.

Timing note: single-threaded extraction of this 6 MB / 78-page file took ~9.6–13 s vs ~2.2–2.6 s for PDFium on the same machine (Windows 11, release build).

## Environment

- pdf_oxide 0.3.73 (crates.io), default features
- Rust stable, Windows 11 x64, release profile
- File: scanned Hebrew Talmud (Tractate Beitza), 6.2 MB, 78 pages, OCR'd Hebrew text layer, simple fonts (not Type0), text stored in visual order — typical of a very large corpus of scanned Hebrew religious texts

I can provide the PDF file (it is freely redistributable) — happy to attach it here or send a minimal single-page repro if that helps.

## Context

We are evaluating pdf_oxide as a pure-Rust replacement for PDFium in the indexing pipeline of [Otzaria](https://github.com/Sivan22/otzaria), an open-source Hebrew library app with thousands of similar scanned PDFs. Correct visual→logical conversion on this class of documents is the deciding factor for adoption.

palmoni5:
Fixes #826.

## Problem

Scanned-Hebrew PDFs whose OCR text layer emits **one show op per glyph** (tesseract-style layers — an enormous corpus of scanned Hebrew religious texts) come out character-reversed from `extract_text`. Root cause chain:

1. Per-glyph spans carry no storage-order signal of their own, so the flush-path detectors (`detect_visual_order_run`, the `accumulated_width` heuristic) never fire — each span is a single character.
2. `merge_adjacent_spans` walks the reading-order-sorted (x-ascending) spans and concatenates them — producing the **visual** left-to-right sequence for RTL text.
3. Nothing downstream reverses it: Pass 0 of `reverse_rtl_visual_order_runs` is gated on Arabic Presentation Forms (Hebrew has none), and the short-span runs that Pass 1 targets no longer exist — the merge already collapsed them.
4. For letter-spaced runs that *escaped* the merge (running heads), Pass 0.5 reversed the span order and Pass 1 then merged in reverse span order — **two reversals that cancel out**, leaking visual-order text.

## Fix

1. **`extractors/text.rs`** — `merge_adjacent_spans` tracks whether the accumulated span was built entirely from ≥2 single-glyph RTL pieces. Such a concatenation is the visual sequence *by construction* (regardless of how the producer stored the text), so `finalize_visual_rtl_glyph_span` reverses it to logical order — keeping digit runs forward (UAX #9 rule L2) and combining marks attached to their base. Build-time-corrected multi-char show strings and Latin/CJK spans are untouched.

2. **`document.rs`** — the short-span run merge moved **before** Pass 0.5 (as "Pass 0.75") and emits logical characters directly, using atomic groups (base+marks, 2-char build-time pieces kept in order, folded digit runs). This removes the Pass 0.5 ↔ Pass 1 cancellation; the old Pass 1 is deleted as subsumed.

## Verification

On a 78-page scanned Talmud volume (Tractate Beitza, one-glyph-per-Tj OCR layer, ~9,000 spans/page):

| metric | before | after |
|---|---|---|
| per-page unique-token containment vs PDFium | 4.8% | **91%** |
| reversed-word occurrences (e.g. logical אמר vs reversed רמא) | 24 / 2,313 | **2,420 / 24** |
| running head | `הציב ןושאר קרפ` | `ביצה פרק ראשון` ✓ |

The full test suite (5,714 tests) passes unchanged. The behavior was consistent across a 17-volume sample (containment 90–92% on every volume).

Happy to add a regression fixture if you can point me at the preferred format for golden corpus additions — the repro PDF from #826 is freely redistributable.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -6822,10 +6822,41 @@ impl PdfDocument {
         // the same run, cancelling out and leaking visual-order text —
         // letter-spaced Hebrew running heads were the visible symptom.
         if spans.len() >= 4 {
+            use crate::text::rtl_detector::is_rtl_diacritic;
+            // A span counts as a genuine single-glyph OCR piece only when it
+            // holds at most one BASE RTL letter (mirrors `extractors/text.rs`'s
+            // `is_rtl_glyph_piece`, which this run-membership test should have
+            // matched from the start) — not merely "≤2 characters total".
+            // A run of ordinary two-letter spans (a routine font kerning-pair
+            // or subsetting-run boundary in any digitally-authored Hebrew/
+            // Arabic PDF) also satisfies "≤2 chars", so the old gate merged
+            // and reversed already-correct text on non-OCR documents. A
+            // base+combining-mark pair (one glyph, two Unicode scalars) still
+            // passes here, since the mark doesn't count as a second base.
+            let is_single_base_rtl_or_space = |s: &TextSpan| -> bool {
+                let mut bases = 0usize;
+                for c in s.text.chars() {
+                    if c.is_whitespace() {
+                        continue;
+                    }
+                    let cp = c as u32;
+                    if is_rtl_diacritic(cp) {
+                        continue;
+                    }
+                    if c.is_alphabetic() && !is_rtl_text(cp) {
+                        return false;
+                    }
+                    bases += 1;
+                    if bases > 1 {
+                        return false;
+                    }
+                }
+                true
+            };
             let mut i = 0;
             while i < spans.len() {
-                let is_short_rtl = spans[i].text.chars().count() <= 2
-                    && spans[i].text.chars().any(|c| is_rtl_text(c as u32));
+                let is_short_rtl = spans[i].text.chars().any(|c| is_rtl_text(c as u32))
+                    && is_single_base_rtl_or_space(&spans[i]);
                 if !is_short_rtl {
                     i += 1;
                     continue;
@@ -6835,12 +6866,7 @@ impl PdfDocument {
                 let mut j = i + 1;
                 while j < spans.len() {
                     let y_same = (spans[j].bbox.y - y).abs() < 2.0;
-                    let is_short = spans[j].text.chars().count() <= 2;
-                    let has_rtl_or_space = spans[j]
-                        .text
-                        .chars()
-                        .all(|c| is_rtl_text(c as u32) || c == ' ');
-                    if y_same && is_short && has_rtl_or_space {
+                    if y_same && is_single_base_rtl_or_space(&spans[j]) {
                         j += 1;
                     } else {
                         break;
--- a/src/document.rs
+++ b/src/document.rs
@@ -6763,11 +6763,11 @@ impl PdfDocument {
         // RTL ratio so that punctuation or stray markers adjacent to
         // Arabic do not trigger a reversal.
         //
-        // Pass 1 below handles the other common shape where each Arabic
+        // Pass 0.75 below handles the other common shape where each RTL
         // character is emitted as its own short span and the reversal is
         // a span-granularity concern. The two passes are independent:
         // a span either fires Pass 0 (pre-shaped, reverse in place) or
-        // Pass 1 (per-glyph spans, reverse span order), never both.
+        // Pass 0.75 (per-glyph spans, merge + reverse), never both.
         //
         // This is separate from `normalize_arabic_presentation_forms`,
         // which runs later on the assembled output string and unshapes
@@ -6806,6 +6806,108 @@ impl PdfDocument {
             {
                 let reversed: String = span.text.chars().rev().collect();
                 span.text = reversed;
+            }
+        }
+
+        // Pass 0.75 (#826): merge per-glyph visual-order runs BEFORE the
+        // word-order pass. Some producers (scanned-Hebrew OCR text layers
+        // are the common case) emit one show op per glyph; the reading-order
+        // sort leaves those single-glyph spans in x-ascending — i.e. VISUAL —
+        // order. Merge each maximal run into one span whose characters are
+        // reversed to logical order (keeping digit runs forward, combining
+        // marks attached to their base, and 2-letter build-time pieces
+        // atomic). Running before Pass 0.5 is essential: the previous
+        // arrangement (span-order reversal in Pass 0.5 followed by a
+        // reverse-order merge in the old Pass 1) applied two reversals to
+        // the same run, cancelling out and leaking visual-order text —
+        // letter-spaced Hebrew running heads were the visible symptom.
+        if spans.len() >= 4 {
+            let mut i = 0;
+            while i < spans.len() {
+                let is_short_rtl = spans[i].text.chars().count() <= 2
+                    && spans[i].text.chars().any(|c| is_rtl_text(c as u32));
+                if !is_short_rtl {
+                    i += 1;
+                    continue;
+                }
+                let run_start = i;
+                let y = spans[i].bbox.y;
+                let mut j = i + 1;
+                while j < spans.len() {
+                    let y_same = (spans[j].bbox.y - y).abs() < 2.0;
+                    let is_short = spans[j].text.chars().count() <= 2;
+                    let has_rtl_or_space = spans[j]
+                        .text
+                        .chars()
+                        .all(|c| is_rtl_text(c as u32) || c == ' ');
+                    if y_same && is_short && has_rtl_or_space {
+                        j += 1;
+                    } else {
+                        break;
+                    }
+                }
+                let run_end = j;
+                let run_len = run_end - run_start;
+
+                // Only process runs of 4+ spans (avoid false positives)
+                if run_len >= 4 {
+                    use crate::text::rtl_detector::is_rtl_diacritic;
+                    // Build atomic groups in forward (visual) order, then
+                    // emit the groups in reverse to obtain logical order.
+                    let mut groups: Vec<String> = Vec::new();
+                    for span in spans[run_start..run_end].iter() {
+                        let cs: Vec<char> = span.text.chars().collect();
+                        let atomic_pair = cs.len() == 2
+                            && !cs[0].is_whitespace()
+                            && !cs[1].is_whitespace();
+                        if atomic_pair {
+                            // Ligature/base+mark piece decoded from a single
+                            // show op — its chars are already logical.
+                            groups.push(cs.iter().collect());
+                        } else {
+                            for &c in &cs {
+                                if is_rtl_diacritic(c as u32) && !groups.is_empty() {
+                                    // Combining mark rides on the preceding base.
+                                    groups.last_mut().unwrap().push(c);
+                                } else {
+                                    groups.push(c.to_string());
+                                }
+                            }
+                        }
+                    }
+                    // Fold consecutive single-digit groups into one atomic
+                    // group so numbers stay forward (UAX #9 rule L2).
+                    let mut folded: Vec<String> = Vec::new();
+                    for g in groups {
+                        let is_digit_group =
+                            g.chars().count() == 1 && g.chars().all(|c| c.is_ascii_digit());
+                        if is_digit_group {
+                            if let Some(last) = folded.last_mut() {
+                                if last.chars().all(|c| c.is_ascii_digit()) {
+                                    last.push_str(&g);
+                                    continue;
+                                }
+                            }
+                        }
+                        folded.push(g);
+                    }
+                    let reversed_text: String =
+                        folded.iter().rev().flat_map(|g| g.chars()).collect();
+
+                    // Merge into first span, expand bbox to cover entire run
+                    let last_span = &spans[run_end - 1];
+                    let new_width =
+                        (last_span.bbox.x + last_span.bbox.width) - spans[run_start].bbox.x;
+                    spans[run_start].text = reversed_text;
+                    spans[run_start].bbox.width = new_width;
+
+                    // Remove the rest of the run
+                    spans.drain(run_start + 1..run_end);
+
+                    i = run_start + 1;
+                } else {
+                    i = run_end;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


yfedoseev:
## Thank you — and a regression found: real Hebrew/Arabic PDFs now come out reversed via `extract_text()`

First off, thank you for reporting #826 and for this fix — the per-glyph diagnosis
and the Beitza-volume validation (4.8%→91% containment) are excellent, careful work
and I can reproduce the improvement locally.

Before merging, we ran this branch through our standard 419-PDF corpus regression
sweep (byte-diff vs `main`, every format) — our usual gate for any change touching
extraction. It flags a regression on two documents in the corpus:
`wiki-cat-he/source.pdf` and `wiki-cat-ar/source.pdf` (Hebrew and Arabic Wikipedia
article exports — ordinary, digitally-authored PDFs, not scanned/OCR).

### Repro

```python
import pdf_oxide
d = pdf_oxide.PdfDocument("wiki-cat-he/source.pdf")
print(d.extract_text(0)[:80])
```

- `main`: `חתול הבית הוא מין מבוית של יונק מסדרת הטורפים...` (correct — "The domestic cat is a domesticated species of...")
- this branch: `לותח תיבה אוה ןימ תיובמ לש קנוי תרדסמ םיפרוטה...` (every word's letters individually mirrored — garbage)

Same pattern on `wiki-cat-ar` (Arabic), though the corruption there is a partial
jumble rather than a clean mirror — likely because the atomic-grouping heuristic in
`finalize_visual_rtl_glyph_span` doesn't cleanly reconstruct Arabic's more complex
diacritic/ligature clustering.

### Root cause

`merge_adjacent_spans`'s new `is_rtl_glyph_piece`/`finalize_visual_rtl_glyph_span`
path fires whenever a maximal run of ≥2 short (≤2-char) RTL-only spans shows up
adjacent in x-order — it doesn't check *why* the spans are short. Scanned-OCR
text layers are one reason (your target case), but ordinary Hebrew/Arabic PDF
producers routinely emit short spans too, for unrelated reasons (kerning pair
breaks, font-run boundaries, subsetting) — and *their* short spans already carry
correctly-ordered logical text. The merge-and-reverse fires on both shapes
identically, so it silently corrupts the second, much more common, category.

This isn't a narrow edge case — of the ~120 broad-coverage PDFs in our sweep,
2/2 real-world Hebrew/Arabic documents in the corpus hit it.

### Suggested fix direction

Gate the merge-and-reverse on some signal that's actually specific to OCR-sandwich
text, not just "short RTL spans in a row." The most direct one available at this
point in the pipeline is text render mode (`Tr` 3/7 — invisible, neither fill nor
stroke — is what OCR-sandwich layers use almost universally, since the recognized
text has no rendering-correctness pressure). If `TextSpan` doesn't currently carry
render mode through to `merge_adjacent_spans`, that'd need threading through, similar
to how `TjBuffer::render_mode` is captured for the buffer-flush sites elsewhere in
this file.

Happy to pair on this / test a revised version against the same corpus + the Beitza
volume once you have a candidate — want to make sure the per-glyph fix (which is
clearly needed and well-validated) doesn't ship with this regression attached.

In the meantime, we opened #832, which fixes the narrower **per-word** OCR-sandwich
shape of #826 (one `TJ` array per recognized word, rather than per glyph) in
`extract_text()` — validated against the same 419-PDF corpus with zero regressions.
It's a different code path from this PR (`extractors/text.rs`'s buffer-flush sites,
not `merge_adjacent_spans`) and doesn't cover the per-glyph shape this PR targets, so
they're complementary rather than overlapping — #826 will need both, once this one's
regression is resolved.

Thanks again for digging into this — real scanned-Hebrew corpora are exactly the kind
of hard case that's easy to get subtly wrong, and the Beitza-volume before/after
numbers made this much easier to review.

palmoni5:
Thank you — this is a fair catch, and the 419-doc sweep gating is exactly the safety net this kind of change needs. I dug into the fix direction you suggested and hit two empirical dead-ends I want to share, because they narrow the problem in a useful way and I think I now need one thing from your corpus to finish it.

## Render mode (`Tr 3/7`) doesn't separate the cases

I threaded render mode through to `TextSpan` and checked the target corpus first. **Every span in the Beitza volume is `Tr 0` (fill / visible), not 3/7.** These scanned-Hebrew seforim aren't the classic invisible-text-over-image OCR sandwich — the recognized text is drawn in fill mode as the visible content (image-free "text PDFs" from the OCR/typesetting tool). So gating the reversal on invisible render mode would disable the fix for exactly the corpus #826 is about. Render mode is out.

## Draw direction doesn't separate them either

My next hypothesis was the storage-order signal the codebase already computes — `detect_rtl_draw_direction` / `rtl_draw_logical` — reverse only visual-order (LTR-drawn) runs, leave logical-order (RTL-drawn) runs alone. But measuring it:

- The Beitza per-glyph runs are flagged `rtl_draw_logical = true` — i.e. they *are* drawn right-to-left (decreasing x), logical storage.
- Gating the reversal on `!rtl_draw_logical` dropped Beitza from 91% containment back to ~10% (the reversed forms return).

So the target and your regression case are **both logical-order-drawn**. Draw direction can't tell them apart.

## What I think is actually going on

The reading-order sort sits between `detect_rtl_draw_direction` (pre-sort) and `merge_adjacent_spans` (post-sort). For the per-glyph Beitza case, the sort reorders the individually-drawn glyphs into x-ascending (visual) order, and my merge-then-reverse puts them back to logical — which is why it works there. Your `wiki-cat-he`/`-ar` case is also logical-drawn but, I suspect, **not one-glyph-per-op** — the short spans come from kerning/subsetting and are multi-glyph, each already internally logical. If so, the corruption is most likely coming from the `document.rs` "Pass 0.75" run-merge (which triggers on any ≤2-char RTL run) over-firing on those, rather than from the `merge_adjacent_spans` glyph-piece path (which requires genuinely single-glyph pieces). That would make the real discriminator *span granularity* (true one-glyph-per-op) plus a per-run visual-order check done **pre-sort** where the x-monotonicity is still intact — mirroring your existing `detect_visual_order_run`, but across a per-glyph run instead of within one multi-glyph cluster.

I can't confirm that without reproducing the regression, and it's not in anything I have locally.

## The ask

Could you share `wiki-cat-he/source.pdf` and `wiki-cat-ar/source.pdf` (or, if they're not redistributable, a `extract_spans(0)` dump for page 0 — per-span `text`, `bbox.x/y`, `char count`, and `rtl_draw_logical`)? With the actual span shapes next to Beitza's I can pin down the discriminator and re-gate precisely, then you can re-run both against the full corpus.

In the meantime #832 landing the per-word shape safely is great — that's the more common case, and it means #826 isn't blocked on getting the per-glyph shape perfect. No objection to this PR sitting until I have a version that passes your sweep; I'd rather it wait than ship the regression.

yfedoseev:
Thanks for the careful root-causing and the honest write-up above — the render-mode and draw-direction dead-ends you shared saved me from repeating them. Pushed a commit onto this branch: one real fix, plus a fixture test file documenting a second, deeper problem I found but haven't solved.

## Fixed: `document.rs`'s Pass 0.75 gate was too loose

Pass 0.75 (the merge-and-reverse for per-glyph scanned-OCR text) fires on any run of 4+ short (≤2-char) same-line RTL spans — it doesn't check *why* a span is short. A genuine one-glyph-per-show-op OCR run and four ordinary already-correct 2-letter spans (a routine font kerning-pair or subsetting-run boundary — common in any digitally-authored Hebrew/Arabic PDF) both satisfy "≤2 chars," so the second case got merged and reversed too.

Tightened the gate to require exactly one *base* RTL letter per span — mirrors `extractors/text.rs`'s `is_rtl_glyph_piece`, which already gets this right (`bases > 1 → false`), so Pass 0.75 is just catching up to a discriminator this codebase already has correct elsewhere. `ordinary_two_char_spans_are_not_reversed` in the new test file pins this.

Verified: full lib suite (5716 tests) + every other RTL integration test file in the repo (`rtl_tj_array_word_buffer`, `rtl_ocr_multiscript_regression`, `rtl_struct_order`, `rtl_markdown_heading`, `rtl_paragraph_reflow_md`, `test_rtl_script_support`) all pass unchanged.

## Still broken: this fix isn't sufficient on its own

Rebased onto current `main` (which now has the per-word OCR-sandwich fix) and re-tested against the real documents from my first review — `wiki-cat-he/source.pdf` and `wiki-cat-ar/source.pdf` still come out with every word individually letter-mirrored, even with the Pass 0.75 fix applied.

Traced it further: those four short spans are tightly kerned (touching, zero gap), so they're also getting swept into an *ordinary* adjacent-span merge — unrelated to any OCR/glyph-piece detection — and the resulting combined span is then reversed by whatever this codebase's general geometric visual/logical detector is doing at that point. I confirmed this precisely (not by eyeballing Hebrew text in a terminal, which is unreliable — bidi rendering will fool you): the failing test's output is an exact character-for-character reversal of the full 8-char run, span boundaries preserved as line breaks. That's a different code path than Pass 0.75 and I haven't found or fixed it — `ordinary_two_char_spans_are_not_reversed`'s doc comment has the full trace.

This is a genuinely deeper problem than a gate tweak: it means "one base RTL char per span" *and* "tightly-kerned short spans" are both unreliable OCR signals, since ordinary digitally-authored PDFs (browser/print-to-PDF renderers doing precise per-glyph kerning, common for justified RTL text) produce the same shapes. Worth trying: checking for an underlying scanned-image XObject on the page — real OCR text sits over a scan, ordinary digital PDFs don't have one. Not available at `merge_adjacent_spans`'s point in the pipeline today, so it'd need threading through, but it's a structural signal rather than another span-shape heuristic, which is what's kept failing here.

Also worth flagging: there's an existing dependency-freshness tracking issue (#793) if any of this touches crate versions — probably not relevant here, just noting it exists.

No PDFs shared (per this repo's fixture policy) — all reproductions are synthetic, built in-code, in the new test file.

yfedoseev (commented):
## 🔴 This can't merge yet — the branch leaves a failing test

@palmoni5 thanks again for the deep root-causing. The `document.rs` Pass-0.75 gate tightening is a strict improvement. But the branch currently has a **committed, non-`#[ignore]`d failing test** — reproducible with:
```sh
cargo test --test rtl_short_span_glyph_granularity
# ordinary_two_char_spans_are_not_reversed FAILS:
#   got  "חז\nוה\nדג\nבא"   expected  "אבגדהוזח"
```
So `cargo test` is red on this branch (the reversal comes from a pre-existing `flush_tj_buffer` + Pass-0.5 path, not Pass 0.75).

Separately, the net-new `finalize_visual_rtl_glyph_span` (`src/extractors/text.rs`) reverses any span merged from ≥2 single-glyph RTL pieces **without a text-render-mode (`Tr`) gate** — unlike `main`'s sibling per-word fix — so it can reverse already-correct non-OCR Hebrew/Arabic (the 419-doc sweep note in `rtl_tj_array_word_buffer.rs` documents this).

Suggested path to green: split the sound `document.rs` gate from the risky `text.rs` finalize; `#[ignore]` the known-failing test with a tracking issue, or `Tr`-gate `finalize_visual_rtl_glyph_span`.

yfedoseev:
Status check on this one, re-verified today against current `main`.

- `ordinary_two_char_spans_are_not_reversed` still fails on the branch as it stands (`1 passed; 1 failed`). To be clear about whose ball this is: that test came in with @yfedoseev's own commit `9595e92c`, added deliberately to document the deeper problem found during review, so the red is a known marker rather than anything you left behind.
- The branch is now 74 commits behind `main` and no longer merges cleanly — `src/extractors/text.rs` conflicts. That file has moved a lot in the meantime, including RTL-adjacent work (word-fusion fixes in #847, `/PlacedPDF` handling in #896, per-glyph advance accuracy), so it's worth rebasing before any further work: some of what you were fighting may have shifted underneath.

Nothing needed from you beyond the rebase right now — the outstanding design question is on our side, and I've asked @yfedoseev whether that documented test should be resolved or marked `#[ignore]` with a pointer so the branch can go green. The Pass-0.75 gate tightening itself still looks like a strict improvement, and the root-causing you did here remains some of the most careful work on this repo.

yfedoseev:
Closing this — for inactivity plus an unresolved regression, not for the quality of the work. I want to be precise about which is which, because the diagnosis here was good and I do not want the close to read as a judgement on it.

**The regression is still there, and I re-verified it today.** Re-measured from scratch against current `main`, with the merge conflict resolved fresh, it reproduces to four decimal places: the same documents, the same magnitudes. Three files that `main` extracts perfectly come out with every word reversed. That is not a threshold to tune — the gate fires on "≥2 merged pieces, ≥2 RTL chars, no strong LTR letters", and that description fits correctly-stored Hebrew exactly as well as it fits the visually-stored OCR text you are targeting. There is no signal in it separating the two.

**Independent research reached the same conclusion.** I put the underlying question — how can a reader tell visually-stored RTL from logically-stored RTL — to a fresh investigation with no knowledge this PR existed. It concluded that for an invisible OCR layer there is **no reliable in-PDF discriminator**, and that the only normative signal is `/ReversedChars` (ISO 32000-1 §14.8), which declares that show strings are stored in reverse of page content order. This library parses that tag today and never acts on it for ordering. That is the gap worth attacking.

**What is not on you.** The failing `ordinary_two_char_spans_are_not_reversed` test came in on my own commit `9595e92c`, added deliberately to document the deeper problem I found while reviewing. The red marker is mine, not something you left behind. Your render-mode and draw-direction dead-ends were genuinely useful — they narrowed the problem and saved me repeating them, and they are why I am confident the heuristic route is closed rather than merely unfinished.

**Why now.** The branch has been inactive since 29 July, is far behind `main`, and no longer merges cleanly — `src/extractors/text.rs` has moved substantially, including RTL-adjacent work. Under [CONTRIBUTING](https://github.com/yfedoseev/pdf_oxide/blob/main/CONTRIBUTING.md), a change carrying a known regression is not in a mergeable state, and carrying it open indefinitely helps nobody: it reads as pending when it is blocked on a design question neither of us has solved.

**#826 stays open.** The bug is real and worth fixing. If you want to come back to it, the version I would review starts from `/ReversedChars` as the gate rather than a shape heuristic, with a fixture proving correctly-stored RTL survives untouched — that negative case is what a future attempt most needs, and its absence is what let this one through. A fresh PR against current `main` would be a smaller job than reviving this branch.

Thank you for the report and the work on it. Sorry it took a month to reach a decision.

PR #829, opened 2026-07-07
perf: opt-out for the char_x_offsets stamp's full second extraction per page

palmoni5:
## Problem

While profiling `extract_text` on the scanned-Hebrew corpus from #826 (dense OCR pages, ~9,000 spans/page), call-counting showed **every page being fully extracted twice**:

```
extract_text_spans: 702 calls over 78 pages   (9× per page)
extract_chars:       78 calls                 (1× per page)
```

One of the two full-pipeline runs per page comes from `postprocess_spans → stamp_char_x_offsets → extract_chars`: a complete second parse + snap + sort + dedup + merge whose only product is `TextSpan::char_x_offsets` — spec-aligned glyph x-origins consumed by `to_chars()` (and through it `extract_words` / `extract_text_lines`). Assembled-string consumers (`extract_text`, `to_plain_text`, markdown/HTML converters' text surface) never read that field, so for them the entire second extraction is pure overhead.

(The remaining ~7 extra calls/page in that count come from annotation AP streams and are proportional to annotation count — a separate, smaller matter.)

## Change

`PdfDocument::set_skip_char_x_offsets(true)` — an opt-out flag checked at the single `stamp_char_x_offsets` call site in `postprocess_spans`.

- **Default `false`: behaviour is byte-identical to today.**
- With `true`: the assembled string is unchanged (the stamp never touches `span.text`); spans cached while skipping have empty `char_x_offsets`, so `to_chars()` falls back to the existing legacy prefix-sum path. The doc comment spells out the contract (text-only documents; don't mix modes on one document because of the per-page span cache).

## Measurement

78-page scanned Talmud volume, 8-thread batch extraction: whole-book `extract_text` wall time **3.9s → 3.0s (~25% faster)**; on single-threaded runs the saving is proportionally larger since the stamp's second extraction dominates more of the per-page cost.

Full test suite passes: `5714 passed; 0 failed`.

An alternative design would be lazy stamping (populate `char_x_offsets` on first `to_chars()` access) which would need no API at all — happy to rework in that direction if you prefer it.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -376,6 +376,14 @@ pub struct PdfDocument {
     /// Uses Arc to avoid expensive deep clones on every page extraction.
     /// Mutex provides interior mutability for `&self` read-path methods (#398).
     structure_tree_cache: Mutex<Option<Option<Arc<crate::structure::StructTreeRoot>>>>,
+    /// When true, `postprocess_spans` skips `stamp_char_x_offsets` — the
+    /// pass that runs a full second extraction (`extract_chars`) per page
+    /// purely to fill `TextSpan::char_x_offsets` with spec-aligned glyph
+    /// positions for `to_chars()`. Text-only consumers that never decompose
+    /// spans (assembled-string surfaces like `extract_text`) can opt out via
+    /// [`Self::set_skip_char_x_offsets`] and halve per-page extraction work.
+    /// Default `false` keeps existing behaviour byte-identical.
+    skip_char_x_offsets: AtomicBool,
     /// Cached per-page structure tree traversal results.
     /// Built once from the structure tree, then O(1) lookup per page.
     /// Mutex provides interior mutability for `&self` read-path methods (#398).
@@ -1054,6 +1062,7 @@ impl PdfDocument {
             font_identity_cache: Mutex::new(BoundedEntryCache::new(512)),
             font_id_hash_cache: Mutex::new(HashMap::new()),
             structure_tree_cache: Mutex::new(None),
+            skip_char_x_offsets: AtomicBool::new(false),
             structure_content_cache: Mutex::new(None),
             actualtext_index_cache: Mutex::new(None),
             mc_actualtext_mcids: Mutex::new(HashMap::new()),
@@ -11529,7 +11538,17 @@ impl PdfDocument {
         // spec-aligned positions instead of drifting prefix-sums. Runs last, on
         // the fully post-processed spans, so alignment sees the same text the
         // consumers do.
-        self.stamp_char_x_offsets(page_index, &mut spans);
+        //
+        // Text-only consumers opt out via `set_skip_char_x_offsets(true)`:
+        // the stamp only fills `char_x_offsets` (never `span.text`), so the
+        // assembled string is identical, but it costs a full second
+        // `extract_chars` extraction per page.
+        if !self
+            .skip_char_x_offsets
+            .load(std::sync::atomic::Ordering::Relaxed)
+        {
+            self.stamp_char_x_offsets(page_index, &mut spans);
+        }
 
         Ok(spans)
     }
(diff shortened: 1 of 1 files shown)


yfedoseev (changes requested):
Reviewed against SOLID/KISS/DRY (clean — minimal, correctly scoped, single responsibility) and spec (not spec-relevant, pure internal perf path). Then ran the full 419-doc corpus sweep for the default path and wrote a standalone test to exercise the opt-in path directly, since it isn't exposed to any binding yet (more on that below). Default path is solid; the opt-in feature itself has two real bugs.

## Default path (`skip_char_x_offsets` never set) — confirmed safe

Full 419-PDF sweep vs `main`: 0 byte-diffs across all 1257 extractions. Byte-identical, as claimed.

(One near-miss worth mentioning since it briefly looked like an unrelated bonus fix: `books/war_peace.pdf` shows a timeout in our baseline but completed in this branch's sweep. Turned out to be a red herring — I hand-timed it against a fresh unmodified `main` build and got ~62s, same as this branch. It's just a slow document sitting right at our harness's 60s timeout; the baseline run was a timing flake, nothing to do with this PR.)

## Opt-in path (`set_skip_char_x_offsets(true)`) — two confirmed bugs

Since this API isn't exposed to Python (or any binding — checked the whole worktree, `src/document.rs` is the only reference), I wrote a small local Rust test to call it directly against 5 sample documents and compare against the default path.

**1. `extract_words()` returns completely empty** under the flag, at least on `war_peace.pdf` (all pages I checked). The doc comment says `to_chars()` "falls back to the existing legacy prefix-sum path" when `char_x_offsets` is empty — that's not what happens; it silently returns zero words. No error, no panic, just empty results for a caller who calls `extract_words()` after opting in (even though the doc comment says not to mix modes, nothing enforces that, and the failure mode is silent data loss rather than an error).

**2. `extract_text()` itself diverges between modes** on a real document (`arxiv_math/2503.04048_pg46.pdf`) — not whitespace/reordering, genuine content differences. E.g. default output has `cos γ − sin ψ\n qdγ\nψ1` while opt-in output has `cosγ − sinψ\n qdγ\nψ1 cosψ)` — an extra fragment appears that isn't in the default output. This directly contradicts the PR's core claim ("the assembled string is unchanged, the stamp never touches `span.text`") — on this document, at least, that's not true in practice.

Since the flag is off by default, neither bug threatens existing callers. But they mean the feature doesn't do what it says for anyone who opts in — and per your own motivation (Hebrew OCR corpus, dense/complex pages), that's exactly the document class most likely to exercise whatever's actually going wrong here.

## Smaller notes
- `Ordering::Relaxed` on the `AtomicBool` — low severity given the safe-by-default failure mode (a stale read just misses the fast path), but worth a comment or tightening given the doc comment already reasons carefully about cross-call ordering ("don't mix modes on the same document").
- No dedicated test was added for the new API surface — the PR relies on "existing suite passes unchanged," which only proves the default path, not the feature itself. A test exercising the opt-in path the way I did above would very likely have caught both bugs before this review.
- No binding exposure yet (Python or otherwise) — not necessarily blocking for a Rust-only perf knob, but worth confirming whether that's intentional, since the motivating workload (batch OCR extraction) sounds like it'd often be driven from Python in practice.

**Recommendation: land the default-path-safe part, but the opt-in feature needs the two bugs above fixed (and ideally root-caused — if `extract_text` itself can diverge, I'd want to understand why before trusting the perf numbers) before it's something callers should actually use.**

palmoni5:
Thank you for the exceptionally thorough review — the 419-doc sweep and, especially, writing a standalone test to exercise the opt-in path directly is exactly what I should have done before opening this.

You're right, and I traced the root cause: my central claim was wrong. `stamp_char_x_offsets` doesn't touch `span.text`, but the text-assembly path consumes `to_chars()` transitively — the RTL line merge and the word/gap clustering (`document.rs` ~10110 and ~16087) — and `to_chars()` derives glyph positions from `char_x_offsets`. When the stamp is skipped, `char_x_offsets` is empty, `to_chars()` falls back to prefix-sum positions, and those different positions flip boundary decisions in assembly. So on documents where assembly reaches those paths, `extract_text` genuinely diverges (your `arxiv_math/2503.04048_pg46.pdf` case), and `extract_words()` — built entirely on `to_chars()` positions — degenerates to empty. Both bugs are the same underlying fact.

The deeper conclusion is that the second extraction isn't dead weight: it feeds positional data that legitimately influences text assembly on some documents. My token-containment metric (bag-of-words per page) was blind to exactly the spacing/fragment differences you caught, which is why I mistook "same tokens" for "byte-identical."

Given that, there's no default-path change to land separately here — the whole PR is the opt-in flag, and the flag isn't safe as designed. Lazy-stamping wouldn't rescue the perf win either, since the assembly that made me want to skip the stamp is the same code that needs the positions. So I'm closing this rather than iterating on a premise that doesn't hold.

If the double extraction is worth attacking, the right place is likely *inside* the stamp/`extract_chars` path (it re-runs the full parse+sort+merge pipeline a second time to recover accurate glyph x-origins that the first pass already computed and discarded), not skipping it wholesale — but that's a real change to reason about, not a flag. I'll leave it for now.

Closing. Thanks again — the review taught me something about this codebase.

PR #884, opened 2026-07-17
perf(text): stop re-deriving per-glyph facts that cannot change (#882)

### LINKED ISSUE #882: [Bug]: `extract_words` / `extract_text` are far slower than `extract_spans` on large PDFs
### Bug Description

On large multi-page PDFs, `extract_words`, `extract_text` take many times longer than `extract_spans` over the same pages. extract_chars/extract_spans stay fast; the word/text methods are 10–22× slower for a full-document parse.

### Steps to Reproduce

1. Open a large multi-page PDF.
2. Time a full pass of extract_spans, then `extract_words`, then `extract_text` over every page.
3. Compare totals.

### Expected Behavior

`extract_words`/`extract_text` over a whole document should be within a small factor of `extract_spans`.

### Actual Behavior

| PDF | pages | extract_spans | extract_words | extract_text |
| :--- | :--- | :--- | :--- | :--- |
| BILLS-116hr133enr.pdf | 2124 | 0.50 s | 5.93 s (12×) | 4.55 s (9×) |
| us-public-health-and-welfare-code.pdf | 13234 | 2.67 s | 59.2 s (22×) | 41.3 s (15×) |

- [BILLS-116hr133enr.pdf](https://github.com/user-attachments/files/30097562/BILLS-116hr133enr.pdf) — 2124 pages, public domain (US legislation, govinfo.gov). Shareable: Yes.
- [us-public-health-and-welfare-code.pdf](https://showcase.apryse.com/gallery/us-public-health-and-welfare-code.pdf) — 13,234 pages, public. Shareable: Yes.

### Environment

- OS: Linux
- Rust version: n/a (Python bindings)
- pdf_oxide version: 0.3.74
- Python version: 3.12

### Code Sample

```rust
import time, pdf_oxide as ox
d = ox.PdfDocument("us-public-health-and-welfare-code.pdf")
n = d.page_count()
for m in ("extract_spans", "extract_words", "extract_text"):
    fn = getattr(d, m); t = time.perf_counter()
    for i in range(n): fn(i)
    print(m, round(time.perf_counter() - t, 2), "s")
```

### Error Messages / Logs

```shell
NA
```

### Additional Context

NA

### Checklist

yfedoseev:
Closes #882 — `extract_words` / `extract_text` being far slower than `extract_spans` on large PDFs.

This lands the byte-identical performance wins; the remaining headroom is tracked separately in #885 (fold the char-extraction parse into span extraction) and #886 (lazy running-header/footer marking).

## What the profiler found

The reported ratio is real (I reproduced 11.6× and 9.3× against the reporter's own 2,124-page PDF, vs their 12× and 9×), but the ratio itself is not the defect — the three methods do different amounts of work, and the ticket's baseline (`extract_spans` via the Python binding) reaches `extract_spans_raw` and skips `postprocess_spans` entirely. What *was* defective is how much that post-processing cost.

With no `perf`/`valgrind` on the box, a `pprof` sampling profile pointed somewhere I would not have guessed:

**`show_text` asked the font for its weight and slant once per glyph**, inside the `TextCharIter` loop. Both are name-derived — `get_font_weight()` lowercases `base_font` (allocating) and runs up to a dozen `contains()` scans; `is_italic()` lowercases it again for two more. The font cannot change inside a `show_text` call, so that is ~14 substring searches and 2 allocations **per glyph — 48.7M times** on a 13,234-page document, for an answer fixed at font-load. `str::contains` and friends were **~38% of all samples**. `get_standard_font_width` had the identical shape: re-stripping the subset prefix and re-scanning the 15-name Standard-14 table per glyph purely to pick a width table.

Both are now resolved once per font behind `OnceLock`, mirroring the existing `byte_to_width_table` memo.

## The rest

- `stamp_char_x_offsets` rescanned every char on the page per span to find its baseline — O(spans × chars). Now brackets a y-sorted index, keeping the exact `abs()` predicate so the selected set and its ordering are unchanged.
- `extract_chars` re-parsed each page for `postprocess_spans`, and the content stream was deep-copied on every access (twice per page on the words path). Both are now cached and shared by `Arc`.
- `extract` copied its whole buffer on return (`TextChar` owns a String → one re-allocation per glyph). Added `extract_owned`, which moves it; `extract` keeps its buffer so `char_count`/`chars` still work after it.
- The word/line paths materialized every glyph twice — once per span, then again flattened. Now built once and indexed by span range.
- `parse_article_threads` walks the whole page tree and was called **per page with no cache**, so per-page cost grew with the document. Cached like the running-artifact signatures.
- `deduplicate_overlapping_chars` rebuilt the array, deep-cloning every glyph to drop a handful. Now retains in place.
- **A reintroduced quadratic**: the word-merge loop re-derived `looks_rtl(&prev.text)` every iteration while `prev` grows by `push_str`, so a k-word merge chain rescanned O(k²) characters — the exact blow-up the backtrack guard directly above it exists to prevent. `looks_rtl` is an `any()`, so `rtl(a + b) == rtl(a) || rtl(b)`: the flag is now carried and OR'd on merge.

## Results

Release build, each method in a fresh process. The box has ~28% run-to-run variance, so the smaller deltas were confirmed with interleaved A/B runs.

| | v0.3.74 | this PR |
| :--- | ---: | ---: |
| us-health, 13,234p — `extract_words` | 156.9 s | **~121 s** |
| us-health — `extract_spans` | 88.7 s | **~54 s** |
| us-health — `extract_chars` | 32.0 s | **~20 s** |
| BILLS, 2,124p — `extract_text` | 6.03 s | **3.94 s** |
| BILLS — `extract_spans` | 5.51 s | **3.39 s** |
| 419-doc corpus (all four methods) | 31 s | **25 s** |

## Verification — output is byte-identical

Rust harness, all sides built from source, over the 419-PDF regression corpus. Three-way, because the PR targets `main` while the perf baseline is the last release:

| | per-method | text / md / html |
| :--- | :--- | :--- |
| v0.3.74 → main | 419/419 identical | 0 byte-diffs |
| **main → HEAD (the gate)** | **419/419 identical** | **0 byte-diffs** |
| v0.3.74 → HEAD | 419/419 identical | 0 byte-diffs |

Per-method covers `extract_spans` / `extract_chars` / `extract_words` / `extract_text_lines` including geometry and `char_x_offsets`. The comparator was control-tested by injecting a single word into one output — it flags it, so the zeros are real. 5735 lib tests pass; fmt and clippy clean.

## Scope — what this does *not* do

This does not make `extract_words` "within a small factor" of `extract_spans`, and it shouldn't: `extract_words` genuinely does strictly more work (post-processing + clustering), so a constant multiple is expected and remains ~9×. What was a real defect — the multiple growing with document size — is substantially reduced.

The largest remaining win is untouched and deliberately out of scope: `stamp_char_x_offsets` still re-parses every page via `extract_chars` to recover per-glyph x positions. Disabling `stamp` entirely drops `extract_spans` to 12.3 s — but it also shifts word counts by −5.5%, so it is load-bearing and cannot simply go. Removing that parse means reconciling `show_text`'s per-element uniform-advance model with span-mode's per-glyph widths, which deserves its own change and its own sweep.

Separately, the Python binding's `extract_spans` skipping `postprocess_spans` is a genuine correctness inconsistency (and the reason the ticket's baseline looked so fast). Left alone here because fixing it changes that method's output.

Thanks to @ankursri494 for the report and the two public, shareable reproducers — having the exact PDFs is what made the profile conclusive.

https://claude.ai/code/session_011GJQbfyAFyAj5t3HPF4qM6


ankursri494:
@yfedoseev Thanks for the quick fix here.

Two follow-ups I think are worth considering:

1. The running-header/footer scan (ensure_running_artifact_signatures). #884 shares the result via Arc, but the underlying 0..page_count scan still runs on the first extract_words/extract_text call. It scans the whole document even when only a few pages are extracted, so single-/few-page extraction from a large PDF pays the full O(all-pages) cost. And when `include_artifacts=true` (artifacts kept rather than filtered) the tags don't appear to be consumed, in my testing, disabling the scan on that path left output byte-identical. Could it be scoped to the requested pages, and gated so it only runs when its result is actually used?

2. Today the clean Python bindings, and the structured per-word data (font/size/color) are exactly what my pipeline needs, and `extract_spans` in particular is wonderfully fast (2.67s on a 13,234-page doc, faster than anything else I tried). One data point I wanted to share, mostly out of curiosity about the ceiling: on that same file, poppler's `pdftotext -bbox` produces accurate word positions in ~17s vs ~59s for `extract_words`, and even `extract_chars` alone (~25s) comes in under poppler's full -bbox with pd_oxide's per-glyph positions matching poppler's exactly. `extract_words` does more than -bbox (per-word font/color, artifact tagging), so some gap is expected; it just made me wonder whether there's any significant headroom left in the word/text paths for very large documents. Sharing it since it's a useful signal, since fast text/word extraction on big files is a big part of what drew me to a Rust engine in the first place.

yfedoseev:
Follow-up spike filed as #885 — folding the char-extraction parse into span extraction (the `stamp_char_x_offsets` duplicate parse). Analysis concluded it lands ~1.7–1.9× on span/word extraction but **cannot** be verified byte-identical (float accumulation order between per-glyph and per-string matrix advance), so it's deliberately scoped out of this PR to keep #884's clean-guarantee intact.

yfedoseev:
Thanks @ankursri494 — both are good signals, and you're right on the substance.

**1. The header/footer scan.** Correct — #884 shares the *result* via `Arc`, but the first `extract_words`/`extract_text` still triggers the full `0..page_count` signature scan, so few-page extraction from a large file pays an O(all-pages) cost it shouldn't. One caveat on the "byte-identical when disabled" finding: it holds for `extract_words`/`extract_spans` (tags set but not filtered when `include_artifacts=true`), but **not** for markdown/HTML — `to_markdown` skips artifact spans (`markdown.rs:1435`) and the IR converter behind HTML special-cases pagination/watermark/background (`pdf_to_ir.rs:285`), so both consume `artifact_type` unconditionally. So the scan can't be globally gated off, and it can't be scoped to the requested pages either (a header only counts if it *recurs* across pages — inherently cross-page).

The actionable version is to make the marking **lazy**: keep it out of the shared `extract_spans` path and only trigger the scan when a consumer actually reads `artifact_type` (md/html, or the artifact-*filtering* word/line paths). Your default `extract_words` path would then never pay for it. Filed as #886 with the design and a byte-identity acceptance gate.

**2. The poppler ceiling.** Really useful, thank you — and `extract_chars` matching poppler's `-bbox` positions exactly is a reassuring independent correctness check. There is real headroom in the word/text paths: the largest remaining item is that span extraction re-parses each page a second time purely to recover accurate per-glyph x-positions. Removing that is tracked in #885 — analysis puts it around 1.7–1.9× on the word/span paths, though (unlike #884) it likely can't be made bit-identical, so it needs its own tolerance-based verification. Between #885 and the lazy-artifact change in #886, the large-document word/text paths should get meaningfully closer to the `-bbox` ceiling, with the extra cost being the genuinely-more work you noted (per-word font/color, artifact tagging).

Fast text/word extraction on big files is a priority here too — appreciate you surfacing concrete numbers to aim at.

Issue #99, opened 2026-02-18
Systematic encoding offset in font text extraction — 'Hello' renders as ')FMMP'


yfedoseev:
## Description

11 PDFs from the veraPDF compliance corpus show a systematic character code offset in text extraction. "Hello world" is consistently rendered as ")FMMP\x01\x01" or similar shifted patterns. Each character is offset by a fixed amount, indicating the wrong encoding is being applied.

## Benchmark Impact

**11 PDFs** produce garbage output. All are veraPDF `Unicode_character_maps` and `Font_subsets` compliance test files.

## Reproduction

```
Expected: "Hello world"
Actual:   ")FMMP\x01\x01"

Character analysis: H(0x48)→)(0x29), e(0x65)→F(0x46), l(0x6C)→M(0x4D)
Shift: -31 consistently (0x1F offset)
```

## Root Cause Analysis

The fonts in these PDFs have a ToUnicode CMap or `/Encoding` dictionary that is either:
1. An indirect reference that fails to resolve silently
2. Being ignored because the font is incorrectly classified as symbolic
3. Being overridden by `StandardEncoding` fallback when the actual encoding should be custom

The -31 (0x1F) systematic offset suggests the character codes are being interpreted through a wrong encoding table where the mapping starts at a different base offset.

## Affected Files (11)

From veraPDF corpus (all "Hello world" test files):
- `PDF_A-1a/6.3/Fonts/6.3.8/Unicode_character_maps/*-t01-pass-{f,g,h,i}.pdf`
- `PDF_A-1b/6.3/Fonts/6.3.5/Font_subsets/6-3-5-t03-fail-{a,b,c}.pdf`
- `PDF_A-2u/6.2/.../Unicode_character_maps/6.2.11.7.2/*-t01-{fail-f,pass-h,pass-i,pass-j}.pdf`

## Labels

bug, text-extraction, fonts, v0.3.8

PR #892, opened 2026-07-18
feat(fonts): flag provably undecodable text layers with a structured warning

### LINKED ISSUE #876: [Feature]: Tell callers when a page's text cannot be read, instead of returning garbled text that looks correct
### Problem Statement

A common subsetting pattern produces a PDF whose text layer carries **no glyph→Unicode information at all**. The page renders perfectly, but the text is not recoverable by any conformant extractor. pdf_oxide currently falls back to CID-as-Unicode and emits confident mojibake for these pages, and a caller has no way to tell that output apart from a successful extraction.

The condition is exact and cheap to detect. All of the following hold for a font:

1. `Type0` with `/Encoding /Identity-H` (or `/Identity-V`), `CIDSystemInfo` ordering `Identity` — content-stream codes are CIDs;
2. `CIDToGIDMap /Identity` — CID == GID, so the codes are **raw glyph indices**;
3. no `/ToUnicode` CMap;
4. the embedded `CIDFontType2` program has **no `cmap` table and no `post` table**.

The subset legitimately has no `cmap`: a CIDFontType2 is addressed by glyph index directly, so rendering never consults one, and subsetters routinely strip both it and `post`. What remains is glyph *outlines* and glyph *indices*. Every path to Unicode is severed:

| recovery path | status |
|---|---|
| `/ToUnicode` CMap | absent |
| invert the font's `cmap` (Unicode→GID ⇒ GID→Unicode) | no `cmap` table |
| `post` table glyph names | no `post` table |
| predefined CMap (CID→Unicode for a known collection) | ordering is `Identity` |

**Why it looks like a constant offset.** Subsetters commonly assign glyph IDs in codepoint order across a contiguous ASCII range, so `GID = ord(char) − k`. Reading the index as a character then yields `char − k` — a tidy shift that looks like an encoding bug. It isn't: nothing in the file *states* that relationship. It is an artifact of subset ordering, and `k` changes per font. A non-contiguous subset yields an irregular substitution map instead of a clean shift — the same condition wearing a different mask. This is worth stressing because the signature is easy to misdiagnose as a cmap-inversion bug and "fix" with a circular mapping; I went down that road first.

**Reproducer** (self-contained, no third-party document; `pip install fonttools`, uses the DejaVu already in `tests/fixtures/fonts/`). Builds a `Type0`/`Identity-H` font whose embedded subset keeps `glyf`/`loca` but has neither `cmap` nor `post`:

<details>
<summary><code>build_undecodable.py</code></summary>

```python
#!/usr/bin/env python3
"""Type0 / Identity-H, no /ToUnicode, CIDToGIDMap /Identity, embedded
CIDFontType2 subset stripped of BOTH `cmap` and `post`. The content stream
carries raw glyph indices; nothing maps a glyph back to a character."""
import pathlib, sys
from fontTools.ttLib import TTFont, newTable
from fontTools.pens.ttGlyphPen import TTGlyphPen

DEJAVU = "tests/fixtures/fonts/DejaVuSans.ttf"
PHRASE = "UNITED STATES DISTRICT COURT"
OUT = pathlib.Path(sys.argv[1] if len(sys.argv) > 1 else ".")

src = TTFont(DEJAVU); cm = src.getBestCmap(); gs = src.getGlyphSet()
# Contiguous ASCII glyph order => GID = ord(c) - 29, the constant offset seen in the wild
used = [chr(x) for x in range(0x20, 0x7F)]
order = [".notdef", "pad1", "pad2"] + [f"u{ord(c):04X}" for c in used]
gid = {c: order.index(f"u{ord(c):04X}") for c in used}

glyf = newTable("glyf"); glyf.glyphOrder = order; glyf.glyphs = {}
hmtx = newTable("hmtx"); hmtx.metrics = {}
for name in order:
    pen = TTGlyphPen(gs)
    if name.startswith("u"):
        ch = chr(int(name[1:], 16))
        if ord(ch) != 0x20 and ord(ch) in cm:
            gs[cm[ord(ch)]].draw(pen)
    glyf.glyphs[name] = pen.glyph(); hmtx.metrics[name] = (600, 0)

for tag in [t for t in src.keys() if t in {"cmap", "post"}]:
    del src[tag]                      # <- the whole point: no cmap, no post
src.setGlyphOrder(order)
src["glyf"] = glyf; src["hmtx"] = hmtx
src["maxp"].numGlyphs = len(order); src["hhea"].numberOfHMetrics = len(order)
ttf = OUT / "undecodable_font.ttf"; src.save(ttf)
fd = ttf.read_bytes()

cids = b"".join(gid[c].to_bytes(2, "big") for c in PHRASE)   # Identity-H: 2-byte CIDs == GIDs
content = b"BT /F1 24 Tf 72 700 Td <" + cids.hex().encode() + b"> Tj ET"
W = " ".join(f"{gid[c]} [600]" for c in sorted(set(PHRASE), key=ord))
objs = [
 b"<< /Type /Catalog /Pages 2 0 R >>",
 b"<< /Type /Pages /Kids [3 0 R] /Count 1 >>",
 b"<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792] /Resources << /Font << /F1 4 0 R >> >> /Contents 5 0 R >>",
 b"<< /Type /Font /Subtype /Type0 /BaseFont /AAAAAA+Sub /Encoding /Identity-H /DescendantFonts [8 0 R] >>",  # NO /ToUnicode
 b"<< /Length %d >>\nstream\n%s\nendstream" % (len(content), content),
 b"<< /Type /FontDescriptor /FontName /AAAAAA+Sub /Flags 4 /FontBBox [0 0 1000 1000] /ItalicAngle 0 "
 b"/Ascent 800 /Descent -200 /CapHeight 700 /StemV 80 /FontFile2 7 0 R >>",
 b"<< /Length %d /Length1 %d >>\nstream\n" % (len(fd), len(fd)) + fd + b"\nendstream",
 (f"<< /Type /Font /Subtype /CIDFontType2 /BaseFont /AAAAAA+Sub "
  f"/CIDSystemInfo << /Registry (Adobe) /Ordering (Identity) /Supplement 0 >> "
  f"/FontDescriptor 6 0 R /CIDToGIDMap /Identity /DW 600 /W [{W}] >>").encode(),
]
out = bytearray(b"%PDF-1.5\n%\xe2\xe3\xcf\xd3\n"); offs = []
for i, b in enumerate(objs, 1):
    offs.append(len(out)); out += b"%d 0 obj\n" % i + b + b"\nendobj\n"
x = len(out); out += b"xref\n0 %d\n0000000000 65535 f \n" % (len(objs) + 1)
for o in offs: out += b"%010d 00000 n \n" % o
out += b"trailer\n<< /Size %d /Root 1 0 R >>\nstartxref\n%d\n%%%%EOF\n" % (len(objs) + 1, x)
(OUT / "undecodable.pdf").write_bytes(bytes(out))
print("ground truth (what a viewer renders):", repr(PHRASE))
```
</details>

[undecodable.pdf](https://github.com/user-attachments/files/30026370/undecodable.pdf)

Ground truth (what a viewer shows): `UNITED STATES DISTRICT COURT`. Every extractor returns the same thing:

| extractor | output |
|---|---|
| pdf_oxide 0.3.74 | `81,7('67$7(6',675,&7&2857` |
| poppler (`pdftotext`) | `81,7('67$7(6',675,&7&2857` |
| pypdf | `81,7('\x0367$7(6\x03',675,&7\x03&2857` |
| pymupdf | `81,7('\x0367$7(6\x03',675,&7\x03&2857` |

pdf_oxide is not doing anything *wrong* — no conformant reader can do better. The problem is that the output is indistinguishable from a successful extraction.

### Proposed Solution

Detect the condition and surface it, rather than silently emitting a CID-as-Unicode guess. Either of:

1. **Return `U+FFFD`** for the affected runs, matching the precedent already set for `/ToUnicode` misses in `src/fonts/character_mapper.rs` — *"a miss against a present CMap must produce U+FFFD — falling back to AGL or an Identity CID-as-Unicode heuristic produces ciphertext"* (#363); or
2. **Expose a diagnostic** (e.g. an `undecodable_text_layer` flag on the extraction result, per page or per font) so a caller can act on it, with no change to the current text output.

I lean toward (2), since changing the text output is a breaking change for anyone relying on today's behavior — but you know the API's compatibility constraints better than I do.

What a caller does with the signal is their own concern — fall back to OCR, skip the page, flag it for review, or surface it to a user. The library's job here is just to stop asserting that undecodable bytes are text. Today there is no signal at any API level, so every caller silently inherits the mojibake.

The relevant code is the `Identity-H`/`Identity-V` branch of `char_to_unicode` in `src/fonts/font_dict.rs`, which falls through to a CID-as-Unicode fallback when the CID isn't otherwise mapped. The check itself is static and cheap: font subtype, encoding, ordering, `/ToUnicode` presence, and whether the embedded program carries `cmap`/`post`.

### Alternatives Considered

- **Heuristic recovery of the constant offset** (try shifts, score the result as text). Rejected: brittle, `k` varies per font, and it silently invents text — strictly worse than admitting the layer is undecodable.
- **Glyph-outline matching against a reference font** (effectively OCR on the vectors). Out of scope for an extraction library, and OCR on the rendered page is the better-understood tool.
- **Do nothing / document it.** The failure is silent, which is what makes it costly downstream; a caller cannot currently distinguish this from success at any API level.

### Feature Area

Text Extraction

### Priority

Would be helpful

### Use Case

I ran pdf_oxide 0.3.74 as the PDF backend of a document-extraction pipeline over a large corpus (~517k born-digital pages). Roughly 1,300 pages — about 0.25% of native pages — land in exactly this condition and are silently indexed as gibberish.

What I need from the library is simply to *know*. Given the signal, my pipeline would automatically hand those pages to OCR — the rendered glyphs are perfectly legible, so it recovers the text — but that is my choice to make, and another caller might reasonably skip the page or flag it. What none of us can do today is tell the difference between this and a clean extraction, because the API reports both identically.

At 0.25% this is a small slice, but it is a *silent* one, and it is the kind of thing that quietly poisons a search index.

### Example Usage (Optional)

```rust
// (2) diagnostic — no change to existing text output
let doc = PdfDocument::open("scan.pdf")?;
let text = doc.extract_text(page)?;
if doc.page_diagnostics(page)?.undecodable_text_layer {
    // The glyphs render fine but carry no Unicode, so `text` is meaningless.
    // What happens next is the caller's call: OCR the page, skip it, flag it.
}
```

### Additional Context

Related issues:

- **#676** — the constant-offset mojibake signature described there is what this condition looks like from the outside. Worth checking whether that font is also `cmap`/`post`-less, in which case it is undecodable rather than mis-decoded. (Separately: I fetched the current Diyanet sermon PDF from `dinhizmetleri.diyanet.gov.tr` and 0.3.74 extracts it correctly — heading `ZEKÂT VE FITIR SADAKASI` and the Turkish body both clean — so whatever #676 was may already be fixed.)
- **#99, #760** — the byte-as-GID cases that *are* recoverable, because the font retains a usable `cmap`. Already handled, and correctly so.

Happy to send a PR if you like the direction — just say which of the two behaviors you'd want.

### Checklist

tobocop2:
## What this fixes

Implements the diagnostic agreed in #876 (fixes #876). A common subsetting pattern (Type0/Identity-H, Identity ordering, `CIDToGIDMap /Identity`, no ToUnicode, embedded font stripped of `cmap` and `post`) leaves a page with no glyph-to-Unicode mapping at all. Extraction can only echo raw glyph indices, and callers had no way to tell that mojibake from a clean extraction.

## The diagnostic

`FontInfo::has_undecodable_text_layer()` detects exactly that severed shape. When such a font is used on a page, extraction emits a new `undecodable_text_layer` structured warning — page-scoped, once per page/font, with the font name and the §9.10.2 reference.

Text output is unchanged. The raw CID sequence stays available to callers who want it; the warning only adds the provable fact that no Unicode mapping exists in the file.

## Staying silent on recoverable fonts

The predicate is structural and conservative. It proves `cmap` absence from the embedded program's own table directory rather than trusting the decoder's cached view, so a program the lazy accessor never inspected (e.g. a CIDFontType2 delivered as an OpenType FontFile3) is judged by what its tables actually say. Fonts with a live embedded `cmap` (the byte-as-GID cases of #99/#760), usable `post` glyph names, a usable ToUnicode, a non-Identity ordering, an explicit `CIDToGIDMap` stream, or an unparseable program are never flagged — when severedness can't be proven, it says nothing.

## Emission and page attribution

Every page-parameterized extraction API surfaces the warning — `extract_text` and the conversion APIs, `extract_chars`, and `extract_spans_with_config` — each attributing it to the page it was asked about, via a diagnostic page context kept deliberately separate from the MCID scope stack (whose values are public span data and stay untouched). Auxiliary extractors with no named page (annotation appearance streams, font enumeration) stay silent rather than misreport a default page. `WarningSink::push_once` keeps the same page/font fact to a single entry across repeated passes and across APIs, and the per-page check is memoized by font identity, so the predicate runs at most once per font per page and adds no measurable work to the `Tf` hot path.

## Bindings

No binding changes needed: `structured_warnings()` serializes the category via its stable string (`"undecodable_text_layer"`), so Python and the other consumers of that surface pick the new category up automatically. The new Rust-side helpers (`push_structured_warning_once`, `WarningSink::push_once`) match the existing Rust-only push surface.

## Test plan

- [x] 16 new tests: the predicate positive plus negative controls for every recoverable variant (embedded cmap, cmap-without-post reachable only via the table directory, post-only, ToUnicode, explicit CIDToGIDMap stream, no/empty/unparseable program, non-Identity ordering, simple fonts, absent /CIDToGIDMap key, present-but-empty ToUnicode); integration tests asserting text output stays byte-identical while the warning fires with correct fields; dedup across repeated extraction; a two-page attribution regression test proving `extract_chars`, `extract_spans_with_config`, and `extract_text` all report the true page and collapse to one warning. Fixtures are three fully synthetic blank-glyph fonts (604 B–1.2 KB) built with fontTools; the PDFs are built in-test.
- [x] Reproducer parity: the `undecodable.pdf` from #876 extracts byte-identically to released 0.3.74 (`81,7('67$7(6',675,&7&2857`) and raises the warning on page 0.
- [x] No-regression sweep: text + HTML extraction over 577 pinned public PDFs (arXiv, PMC OA, govdocs1, pdf.js corpus, IRS, multilingual/RTL), main vs this branch — 577/577 byte-identical.
- [x] Precision probe over the same corpus: 7 documents flagged, all true positives with provably severed subsets — cmap/post-stripped Wingdings/ZapfDingbats (poppler's `pdffonts` confirms no Unicode map) and one arXiv paper whose embedded ArialMT subsets all lack `cmap` and `post` (verified table-by-table with fontTools; `pdftotext` emits the same mojibake there). Zero false positives, and page attribution is exact.
- [x] `make check` green (fmt, clippy `-D warnings` incl. `--features python`, full test suite); clippy also clean on default features.


yfedoseev:
@tobocop2 first: thank you for this — the diagnosis in #876 and the engineering here are genuinely excellent (the non-destructive design, the byte-identical differential sweep, the careful "only the fully-severed case" predicate). And I owe you an honest note: my earlier comment on #876 said "a PR would be very welcome" and pre-endorsed the diagnostic direction. That was a premature maintainer commitment on my part, made before we'd settled the scope — the bind that creates here is on us, not you.

After sitting with it, we've decided to solve the underlying problem a different way, and I want to lay out the difference plainly rather than just close this.

**The fork: a judgment vs. a fact.** This PR ships a *judgment* — `has_undecodable_text_layer()` and an `undecodable_text_layer` warning that assert "this text can't be read." We landed on exposing the *fact* instead: **which ISO 32000-1 §9.10.2 mapping tier produced each span's text** — `ToUnicode` / `EncodingName` / `PredefinedCMap` / `EmbeddedCmap` / `ActualText`, or **`Fallback`** when no tier mapped it and the value is a fabricated glyph-index echo. That work is in #893 (`MappingProvenance`).

Why the fact rather than the judgment:

1. **Spec.** §9.10.2 leaves the failed-mapping case to reader discretion ("there is no way to determine what the character code represents … a conforming reader may choose a character code of their choosing"). So "undecodable" isn't a conformance gap the library must assert — but *which tier fired* is exactly the spec's own cascade, so reporting it is maximally spec-aligned.

2. **Completeness.** The predicate here is gated on TrueType (`is_truetype_font || cid_font_type == "CIDFontType2"`), so a **CIDFontType0 / CFF** font with the identical severed condition returns "not undecodable" — a false all-clear. Provenance is captured from the mapping cascade, so it's `Fallback` for *every* font type, not just TrueType.

3. **It doesn't overclaim.** A severed-subset scan and an intentional glyph-index payload are structurally identical — the library can't know which. "Undecodable" asserts corruption we can't prove; "no mapping tier produced this" is just true.

4. **No new API surface.** Provenance rides the surfaces spans already have — serde/JSON, the C-ABI element accessor, the Python getter — rather than a new `warnings` subsystem. It's one fact that subsumes undecodable-detection, garbled-flagging, OCR-routing, and confidence, for every caller.

**Your use case is fully served**, and arguably better: instead of a boolean flag, your pipeline routes on `span.provenance == Fallback` (or the `"fallback"` label in the JSON/structured output) — which now also catches the CFF-severed pages #892 would miss. Same OCR-routing outcome, complete coverage, and you keep the raw echo for the payload case.

So I'm going to **decline this PR in favor of #893**, with real credit to you: #893 exists *because* your #876 nailed the condition and the trap (the offset-recovery dead end). I've credited the diagnosis in the CHANGELOG/PR. Thank you for pushing on this — the analysis is what made the right design obvious, and the premature-invite misstep was mine to own.

tobocop2:
sounds good to me. thanks!

PR #893, opened 2026-07-18
feat: mapping provenance — expose which §9.10.2 tier produced span text (all bindings)

### LINKED ISSUE #876: [Feature]: Tell callers when a page's text cannot be read, instead of returning garbled text that looks correct

yfedoseev:
## Summary

Adds **mapping provenance** as a first-class *fact*: which ISO 32000-1 §9.10.2 character-to-Unicode tier produced a span's text — or `Fallback`, meaning the font carried **no** mapping resource and the text is a fabricated glyph-index echo, not read from the file (§9.10.2: "there is no way to determine what the character code represents … a conforming reader may choose a character code of their choosing").

This is the library-correct, spec-aligned answer to "how do I know extracted text is real?": the library exposes the **fact**, callers compose the **policy** (route to OCR, flag a page, keep the raw echo). It replaces the need for per-use-case judgment flags (e.g. an "undecodable text layer" warning), and unlike a font-type-specific structural check it is **complete across all font types**.

## What's included

| Layer | Change |
|---|---|
| **Enum** | `MappingProvenance` (spec-ordered: `ActualText › ToUnicode › EncodingName › PredefinedCMap › EmbeddedCmap › Fallback`) with `rank`/`weaker`/`is_from_file`/`as_str`. Serializes as stable lowercase labels. |
| **Font fact** | `FontInfo::best_mapping_provenance()` — the most authoritative resource a font offers, derived from structure, all font types. `Fallback` = the fully-severed undecodable-subset case. |
| **Span** | `TextSpan.provenance: Option<MappingProvenance>`, populated on the extraction path from each span's resolved font. |
| **serde** | serialized as `"provenance": "<label>"` (omitted when absent) → every JSON/serde binding (**WASM**, JSON surfaces) exposes it automatically. |
| **Python** | `span.provenance` getter. |
| **C-ABI** | `pdf_oxide_element_get_provenance(elements, index, err)` → reaches the typed native bindings (**C#, Java, Go, Swift, Kotlin, C++, Zig, …**). |

## Verification

- Full lib suite **5745 passed, 0 failed** — serializing the field caused **zero** snapshot breakage (absent-provenance spans omit it; extracted spans add it).
- 729 extraction/layout tests green (no regression from the span field).
- Integration: `extraction_api_regression` (52) + `test_structured_extraction` green.
- Compiles under `--features python` and `--features wasm` (wasm lib).
- Font-level + enum: 11 focused unit tests.

## Design note

`FontInfo::best_mapping_provenance()` is intentionally a **fact**, not a judgment — it says "which mapping tier exists", never "this text is corrupt" (a scanned page and an intentional glyph-index payload are indistinguishable, so the library must not assert the latter). This is the primitive that lets a caller build "undecodable page" detection themselves, for all font types.

## Remaining (mechanical, per-language CI)

Idiomatic wrapper accessors over the C-ABI call in the typed native binding sources (a `Provenance` property in C#/Java/Swift/etc.). The FFI + serde surfaces make the fact available in every language today; these are thin last-mile additions best built and tested in each language's own toolchain.

Closes #876 (as the fact-based alternative to the diagnostic-warning approach in #892).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -455,3 +455,10 @@ r/*.tar.gz
 /elixir/deps/
 /elixir/priv/
 /elixir/mix.lock
+
+# Viewer build artifacts (never commit generated output)
+viewer/dotnet/**/bin/
+viewer/dotnet/**/obj/
+viewer/dist-browser/
+viewer/**/node_modules/
+viewer/**/.gradle/
--- a/cpp/include/pdf_oxide/pdf_oxide.hpp
+++ b/cpp/include/pdf_oxide/pdf_oxide.hpp
@@ -176,6 +176,11 @@ struct Element {
     std::string type;
     std::string text;
     Bbox rect;
+    /// ISO 32000-1 §9.10.2 mapping-provenance label ("to_unicode"/"encoding"/
+    /// "predefined_cmap"/"embedded_cmap"/"actual_text"/"fallback"), or empty
+    /// when unknown. "fallback" means the text is a fabricated glyph-index echo,
+    /// not read from the file.
+    std::string provenance;
 };
 
 /// A single interactive AcroForm field (PHASE-8 forms).
@@ -899,6 +904,11 @@ class Document {
                 code = 0;
                 e.text = detail::take_string(pdf_oxide_element_get_text(list, i, &code),
                                              code, "Document::page_get_elements");
+                code = 0;
+                if (char *p = pdf_oxide_element_get_provenance(list, i, &code)) {
+                    e.provenance = detail::take_string(p, code,
+                                                       "Document::page_get_elements");
+                }
                 Bbox b{0, 0, 0, 0};
                 pdf_oxide_element_get_rect(list, i, &b.x, &b.y, &b.width, &b.height,
                                            &code);
--- a/csharp/PdfOxide/Internal/NativeMethods.cs
+++ b/csharp/PdfOxide/Internal/NativeMethods.cs
@@ -7097,6 +7097,18 @@ namespace PdfOxide.Internal
             out int errorCode);
 
         /// <summary>
+        /// Gets the §9.10.2 mapping-provenance label of an element
+        /// ("to_unicode"/"encoding"/"predefined_cmap"/"embedded_cmap"/
+        /// "actual_text"/"fallback"), or null when unknown.
+        /// </summary>
+        [LibraryImport(LibName, EntryPoint = "pdf_oxide_element_get_provenance", StringMarshalling = StringMarshalling.Utf8)]
+        [UnmanagedCallConv(CallConvs = new[] { typeof(CallConvCdecl) })]
+        public static partial IntPtr pdf_oxide_element_get_provenance(
+            NativeHandle elements,
+            int index,
+            out int errorCode);
+
+        /// <summary>
         /// Gets the bounding rectangle of an element.
         /// </summary>
         [LibraryImport(LibName, EntryPoint = "pdf_oxide_element_get_rect", StringMarshalling = StringMarshalling.Utf8)]
--- a/dart/lib/pdf_oxide.dart
+++ b/dart/lib/pdf_oxide.dart
@@ -1169,6 +1169,8 @@ class _Native {
             .lookupFunction<_ListStrC, _ListStrD>('pdf_oxide_element_get_type'),
         elementGetText = lib
             .lookupFunction<_ListStrC, _ListStrD>('pdf_oxide_element_get_text'),
+        elementGetProvenance = lib
+            .lookupFunction<_ListStrC, _ListStrD>('pdf_oxide_element_get_provenance'),
         elementGetRect = lib.lookupFunction<_ElemRectC, _ElemRectD>(
             'pdf_oxide_element_get_rect'),
         elementsFree = lib
@@ -1680,6 +1682,7 @@ class _Native {
   final _ElemCountD elementCount;
   final _ListStrD elementGetType;
   final _ListStrD elementGetText;
+  final _ListStrD elementGetProvenance;
   final _ElemRectD elementGetRect;
   final _ListFreeD elementsFree;
   final _ElemJsonD elementsToJson;
@@ -7311,7 +7314,7 @@ class OcrEngine implements Finalizable {
 
 /// A single layout element read from an [ElementList].
 class Element {
-  const Element(this.type, this.text, this.rect);
+  const Element(this.type, this.text, this.rect, [this.provenance = '']);
 
   /// The element type label (e.g. `Text`, `Image`, `Table`).
   final String type;
@@ -7321,6 +7324,12 @@ class Element {
 
   /// The element's bounding box in page user-space points.
   final Bbox rect;
+
+  /// ISO 32000-1 §9.10.2 mapping-provenance label (`to_unicode`/`encoding`/
+  /// `predefined_cmap`/`embedded_cmap`/`actual_text`/`fallback`), or `''` when
+  /// unknown. `fallback` means the text is a fabricated glyph-index echo, not
+  /// read from the file.
+  final String provenance;
 
   @override
   String toString() => 'Element($type, $rect)';
@@ -7372,6 +7381,8 @@ class ElementList implements Finalizable {
     _check();
     final type = _str(_n.elementGetType, index, 'elementGetType');
     final text = _str(_n.elementGetText, index, 'elementGetText');
+    final provenance =
+        _str(_n.elementGetProvenance, index, 'elementGetProvenance');
     final x = calloc<Float>();
     final y = calloc<Float>();
     final w = calloc<Float>();
@@ -7380,7 +7391,8 @@ class ElementList implements Finalizable {
     try {
       _n.elementGetRect(_handle, index, x, y, w, h, code);
       if (code.value != 0) throw PdfOxideError(code.value, 'elementGetRect');
-      return Element(type, text, Bbox(x.value, y.value, w.value, h.value));
+      return Element(
+          type, text, Bbox(x.value, y.value, w.value, h.value), provenance);
     } finally {
       calloc.free(x);
       calloc.free(y);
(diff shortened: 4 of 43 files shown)


tobocop2:
thank you for this!

PR #914, opened 2026-07-21
fix: Do operator misreads XObject name when stray operands precede it

### LINKED ISSUE #913: [Bug]: Text inside a Form XObject is missing from `to_plain_text()` / `extract_text()` / `extract_chars()`
### Bug Description

When a page draws text by invoking a **Form XObject** (`... /MyXObj Do`), the
text inside that XObject is **not** included in the output of any of the text
extraction APIs:

- `to_plain_text(page)`
- `extract_text(page)`
- `extract_chars(page)`

Only text drawn directly in the page's own content stream is returned. The
glyphs render correctly in `render_pixmap()`, so the text is genuinely present
in the document — it is only *extraction* that drops it.

Compositing content as a Form XObject (stamps, overlays, watermarks, reused
headers/footers, imposition, etc.) is entirely standard, and many real-world
PDFs draw part or all of their body text through XObjects, so this can make
extraction return partial or empty results for otherwise ordinary pages.

### Steps to Reproduce

Run the following code after downloading the two PDFs listed/linked in the PDF field below
```
import pdf_oxide

for name in ["case_a_overlay_xobject.pdf", "case_b_all_xobject.pdf"]:
    doc = pdf_oxide.PdfDocument.from_bytes(open(name, "rb").read())
    print(name)
    print("  extract_chars:", repr("".join(c.char for c in doc.extract_chars(0))))
    print("  to_plain_text:", repr(doc.to_plain_text(0)))
    print("  extract_text :", repr(doc.extract_text(0)))
```

### Expected Behavior

```
case_a_overlay_xobject.pdf
  extract_chars: 'base body textoverlay text'   # (both strings, in some order)
  to_plain_text: 'base body text\noverlay text\n'
  extract_text : 'base body text ... overlay text'
case_b_all_xobject.pdf
  extract_chars: 'base body textoverlay text'
  to_plain_text: 'base body text\noverlay text\n'
  extract_text : 'base body text ... overlay text'
```

### Actual Behavior

```
case_a_overlay_xobject.pdf
  extract_chars: 'base body text'
  to_plain_text: 'base body text\n'
  extract_text : 'base body text'
case_b_all_xobject.pdf
  extract_chars: ''
  to_plain_text: ''
  extract_text : ''
```

Two sample files are incliuded here:

| File                          | Structure                                                                 |
|-------------------------------|---------------------------------------------------------------------------|
| [case_a_overlay_xobject.pdf](https://github.com/user-attachments/files/30174491/case_a_overlay_xobject.pdf)  | Page with its own **direct** body text (`"base body text"`) **plus** an overlay Form XObject containing `"overlay text"`. |
| [case_b_all_xobject.pdf](https://github.com/user-attachments/files/30174493/case_b_all_xobject.pdf)      | **Blank** page; both `"base body text"` and `"overlay text"` are drawn via Form XObjects (nothing drawn directly).        |

Both use only the standard base-14 fonts (Helvetica / Times-Roman) — no
embedded/external fonts are required.

### Environment

- **OS**: Arch Linux up to date
- **Rust version**: 1.95.0
- **pdf_oxide version**: 0.3.74
- **Python version** (if using Python bindings): 3.14.6

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

The extraction pipeline appears to walk only the page's own content stream and
does not recurse into invoked Form XObjects (`Do` operators whose referenced
XObject has `/Subtype /Form`). To extract the text it must:

1. resolve the invoked XObject from the page (or parent XObject) `/Resources
   /XObject`,
2. parse its content stream, and
3. position the resulting glyphs using the CTM in effect at the `Do` (the
   `cm` matrix pushed before the `Do`, composed with the XObject's `/Matrix`).

## Suggested fixes

1. In the text-extraction pipeline, recurse into invoked Form XObjects and keep
   their glyphs/spans (applying the `Do`-site CTM and the XObject `/Matrix`) so
   `to_plain_text` / `extract_text` / `extract_chars` all include XObject text.
2. Handle **nested** XObjects (an XObject that itself invokes another), with a
   recursion guard against cyclic `Do` references.
3. Add regression fixtures for: (a) text drawn only via a Form XObject on an
   otherwise-blank page, (b) a page mixing direct text with an overlay XObject,
   and (c) nested XObjects.

### Checklist

yfedoseev:
Closes #913

## Root cause

Text drawn via a Form XObject (`Do`) went missing from `extract_chars` /
`extract_text` / `to_plain_text` / `to_markdown` / `to_html` alike — but not
because the extractor fails to recurse into Form XObjects (it already does,
with `/BBox` clipping, `/Matrix` composition, MCID scoping, and span/stream
caching). The actual bug is narrower.

Both of the issue's repro PDFs invoke their XObject as:

```
q 1 0 0 1 20 150 /Name Do Q
```

with **no `cm` token** — the six numbers are operands with no operator to
consume them (almost certainly a dropped/malformed `cm` in whatever produced
these files). `Do`'s operator-building code in `src/content/parser.rs` reads
`operands[0]` for the XObject name — correct when the accumulated operand
list holds exactly one element, but here operand 0 is one of the stray
numbers instead of the Name, so `Do` resolves to an empty name and the
XObject is silently never invoked.

Per ISO 32000-1:2008 §7.8.2: *"In PDF, all of the operands needed by an
operator shall immediately precede that operator... operands shall not be
left over when an operator finishes execution."* So `Do`'s operand here is
whatever immediately precedes it — the Name, the last element — not
whatever sits at index 0.

## Fix

Both `Do`-operand-building sites (the nom-based parser and the fast
in-BT/ET parser) now read the *last* accumulated operand instead of index 0.

## Verification

- **poppler** (`pdftotext -layout`) and **pymupdf** both extract the
  XObject's text from the same malformed shape, and neither applies a
  translate from the dangling numbers (pymupdf's word bbox for the overlay
  text sits at the form's own untranslated position) — confirming "discard
  the stray operands" is the right interpretation, not "treat them as an
  implicit `cm`".
- All five output surfaces (`extract_chars`, `extract_text`,
  `to_plain_text`, `to_markdown`, `to_html`) were equally broken before the
  fix and equally fixed after — they all route through the same
  content-stream operator parser, so one fix aligns all of them.
- 419-PDF corpus regression sweep (text/md/html) vs `main`: **0 byte-diffs,
  0 crash-state changes** — the fix is a no-op for every well-formed PDF in
  the corpus (a conformant `Do` always has exactly one operand, so index 0
  and the last index are the same value) and only changes behavior on the
  malformed shape this issue reports.
- `cargo fmt` / `cargo clippy --lib` clean; targeted `cargo test --lib`
  green (parser: 217, xobject: 23, content: 382, extractors: 611), plus 4
  new regression tests covering direct+overlay text, XObject-only/blank
  page, and nested XObjects — all via in-code minimal PDFs.

## Test plan

- [x] `cargo test --lib` (parser, xobject, content, extractors modules)
- [x] `cargo fmt --check` / `cargo clippy --lib`
- [x] 419-PDF corpus regression sweep vs `main` (0 diffs)
- [x] Cross-validated against poppler and pymupdf on the reporter's exact
      repro shape


yfedoseev:
## `/verify-pr` — VERDICT: READY ✅

Full verification pass (maintainer's own PR, so no formal approval button — posting as review).

**1. Bug is real (anti-slop revert-check).** Reverting *only* `src/content/parser.rs` to main (keeping the tests) makes all 3 e2e tests FAIL with the exact expected panics — `missing XObject text: "base body text"`, `chars == "overlay text"` assertion, nested text missing. Restore the fix → all pass. The defect (Form-XObject text silently dropped) reproduces on main.

**2. Fix is spec-correct.** Both parse paths (`parse_operator_with_operands` nom / `parse_text_operator_fast` fast) build a **fresh per-operator operand buffer and return after emitting one operator** — verified there's no bleed from a prior operator — so `.last()` is unambiguously *the operand immediately preceding* `Do`, per ISO 32000-1:2008 §7.8.2 / §8.10. For a conformant single-operand `/Name Do`, `last == operands[0]`, i.e. **zero behavior change on valid PDFs**. Fast path is panic-safe (fixed `[Option; 8]` array + `saturating_sub`, indexes `[0]`→`None`→`""` on a zero-operand `Do`). Nice touch: this now matches the existing `scn`/`SCN` last-operand-as-name idiom already in the same file.

**3. Regression — Rust-native, 419-corpus × 3 formats (text/markdown/html), main vs PR: 0 diff / 1257 files.** Ran the `dump_all_formats` example (`extract_text` / `to_markdown_all` / `to_html_all`) on both branches. No corpus PDF changes output — confirming no conformant document has a multi-operand `Do`, and empirically covering the markdown/html surfaces too.

**Tests:** revert-checkable, self-built minimal fixtures (no third-party PDFs), covering both parse paths + direct/overlay, xobject-only, and nested recursion.

Two optional non-blocking follow-ups:
- Add a `to_markdown`/`to_html` `contains("overlay text")` assertion to fully back the body's "all 5 surfaces" claim (the corpus sweep covers it, but a unit assertion is cheaper to keep honest).
- Worth a one-line note that the fix assumes stray operands *precede* the name (true for real dropped-`cm` corruption); a pathological `/Name 1 2 Do` would now resolve `""` — not emitted by any real producer, and the corpus 0-diff confirms it doesn't occur.

Adding to the merge queue ahead of the governance PRs.

Issue #1, opened 2025-11-26
A personal thank you


magnus-trent:
I've been grappling with this issue for a very long time now. And this just unlocked the entire pipeline I've been working on for nearly the same amount of time! Almost a month ago I started, weird timing. 

But thank you so much for making this! 😁

leonardodepaula:
I would also like to say this is a great initiative. Congratulations and the project looks very promising.

chaudharydeepanshu:
Great work and thanks! It helped me migrate [old Flutter package](https://github.com/whuppi/pdf_manipulator) from iText7 AGPL to MIT, with cross-platform support. Thanks for rust and wasm bindings.

PR #943, opened 2026-07-24
fix: preserve add_text() overlays through destructive redaction on the same page (supersedes #799)

### LINKED ISSUE #941: Redaction + add_text() overlay: multi-stream /Contents and font-name registration (tracking #799)
## Summary

Tracking issue for the redaction-overlay bugs diagnosed in #799, per the "anchor the work to an accepted issue" (rule-1) baseline. #799 was closed to keep the review queue current, not rejected — the diagnosis was accepted (all three bugs real, root-cause accurate, validated independently with Poppler `pdftotext`). Two must-fixes remain before the work can merge; this issue tracks them so a follow-up PR has something to close against.

Context: combining `add_text()` with `apply_redactions_destructive()` on the same page — the common PII pattern of redacting a sensitive value and inserting a replacement token at the same position.

### 1. (must fix) Multi-stream `/Contents` array not fully replaced by destructive redaction

The destructive-redaction block replaces the page `/Contents` wholesale with a single reference to the redacted stream, which drops any `add_text()` overlay stream already merged into the `/Contents` array. #799's fix replaced only `arr[0]`, keeping `arr[1..]` — but that assumes the original `/Contents` was a single stream.

A page whose **original** `/Contents` is already an array `[c1, c2]` (legal, ISO 32000-1 §7.7.3.3, and common) becomes `[redacted, c2]`, where `c2` is an *original, un-redacted* stream. The `redacted_orphan_ids` + write-skip backstop physically drops `c2` (so **no content leak** — the "remove all traces" guarantee §12.5.6.23 holds), but `/Contents[1]` is then a **dangling reference**: lenient readers render fine, strict validators (veraPDF / Poppler strict mode) flag the broken array.

**Fix shape:** track the original content object id(s) and drop *all* originals, keeping only genuine overlay/addition refs — rather than assuming array position 0.

### 2. (must fix) Overlay font registered under raw name, but `Tf` emits the `map_font_name`-transformed name

The overlay-font registration keys on the raw `t.font.name`, while the overlay stream emits `map_font_name(&t.font.name, style.weight.is_bold())` (`src/writer/content_stream.rs:861`). These diverge exactly in the cases the fix targets, leaving the "Unknown font tag" bug live for:

- **bold/italic overlays** — `Helvetica` + bold → stream emits `/Helvetica-Bold Tf`, but only `/Helvetica` is registered → dangling `Tf`;
- **generic families** (`"Arial"`, `"sans-serif"`) — not Base-14, so unregistered, but the stream still emits `/Helvetica Tf`;
- **Symbol / ZapfDingbats** — `map_font_name` routes them to Helvetica, so the referenced `/Helvetica` dangles while an unused font object is written.

**Fix shape:** key registration off the `map_font_name(...)` output (the name actually emitted). Symbol/ZapfDingbats can be dropped from the registration name set since the writer never emits them.

### 3. (nice to have) Inherited `/Resources` shadowing

When a page has no own `/Resources` key (legal — inheritable from the `/Pages` chain, §7.7.3.4), building a fresh font-only `/Resources` shadows the inherited dict and drops the page's original fonts/XObjects. `build_page_font_metrics` (`src/editor/document_editor.rs`) already walks `/Parent` for resources — reusing that pattern keeps it consistent.

## Tests to add

- multi-stream `/Contents` (original already `[c1, c2]`) round-trip
- indirect `/Resources` / indirect `/Resources/Font` round-trip (the headline complexity of #799, currently unexercised)
- a styled (bold) overlay font, to lock down the `map_font_name` keying
- promoting the manual Poppler `pdftotext` cross-check into CI if easy

No corpus regression sweep needed — this is confined to the editor/redaction write path, which the extraction sweep never touches. Unit + editor round-trip tests are the right gate.

A follow-up PR implementing must-fixes 1 & 2 (plus tests) will reference this issue.

thomnico:
Supersedes #799 (closed after the review queue caught up; couldn't be reopened after a rebase onto current `main`). Closes #941.

Thanks for the detailed review on #799 — this lands the must-fixes, now rebased onto current `main` (post-#942), plus the requested tests. The diagnosis is unchanged; this addresses the blockers.

## The two must-fixes

**1. Multi-stream `/Contents` — now covered upstream by #942.**
As you noted, #942 fixed the same defect class on `main`: destructive redaction now rewrites `/Contents` wholesale as `[redacted, overlay_additions]`, tracking the additions stream by object id, so no original entry can survive as a dangling reference. During the rebase this PR's own orphan-id-based reconciliation was dropped in favor of that implementation — this PR now contributes only the regression tests for the redaction-path trigger (`destructive_redaction_multi_stream_contents_no_dangling_ref`), which pass against the #942 code.

**2. Font registered under the emitted name, not the raw one.**
The Standard-14 name mapping is extracted from `ContentStreamBuilder::map_font_name` into a single `pub(crate) fn map_base14_font_name(name, bold)` source of truth; the writer method delegates to it, and the overlay-font registration now keys off `map_base14_font_name(&t.font.name, t.is_bold())` — the exact name the `Tf` operator emits. This closes the divergence for bold/italic, generic families (`Arial`/`sans-serif`), and Symbol/ZapfDingbats. The `BASE14_NAMES` allowlist is gone.

## Also addressed

- **Indirect `/Resources` / `/Resources/Font`** are resolved through references (the headline complexity), and pre-existing font entries (`/F1`, `/F3`) are preserved when `/Resources` is rebuilt.
- **Inherited `/Resources` (§7.7.3.4):** when a page has no own `/Resources`, we seed **only** the inherited `/Font` sub-dict — never the whole inherited dict. Copying an ancestor's `/XObject`/`/Pattern` onto the page would make those streams freshly reachable and let them survive a later orphan-drop, re-introducing a stray reference into a destructive-redaction output. (Caught while red-teaming this change.)
- **`save_page()` accumulation:** `overlay_additions.insert()` discarded overlays staged by a prior `save_page()` call for the same page; now accumulates via `.entry(page).or_default().extend()`.

## Scope of the guarantee

This removes traces from the page content streams that get rewritten — the §12.5.6.23 mechanism the redaction module already implements via `redacted_orphan_ids` + the write-skip, **unchanged here**. It does not add document-level sanitization of forms, annotations, XMP, or embedded files; that's out of scope for this PR.

## Tests

Five regression tests in `editor::document_editor::tests`, each proven to **fail on the un-fixed code** via negative controls before being kept:

- `destructive_redaction_preserves_add_text_overlay` (bugs 1 & 2)
- `destructive_redaction_multi_stream_contents_no_dangling_ref` — multi-stream `/Contents` + **indirect** `/Resources`+`/Font`; asserts no dangling ref by scanning the raw saved bytes (not `load_object`, which masks a missing object via file-scan/`Null` fallback)
- `styled_overlay_font_registered_under_emitted_name` — bold overlay, asserts `/BaseFont /Helvetica-Bold` is actually written
- `overlay_does_not_inline_inherited_xobject_into_page_resources` — inherited `/XObject` must not be copied onto the page
- `multiple_add_text_calls_on_same_page_all_survive` (bug 3)

Output independently validated with **Poppler** (`pdfinfo`/`pdftotext`, exit 0, no warnings): the redacted value is gone from the bytes, the replacement token and the unrelated second stream are both present.

```
cargo test          # full default-features suite on the rebased branch, exit 0
lib:  test result: ok. 5779 passed; 0 failed; 2 ignored
integration: 40 test binaries, all ok (418 tests, 0 failed)
```

`cargo fmt --check` and `cargo clippy -- -D warnings` clean. Poppler cross-check re-run on the rebased branch (pdftotext/pdfinfo 26.07.0, exit 0, no warnings): redacted value absent from the raw bytes, bold replacement token extracted, `/Helvetica-Bold` declared under the emitted name.

## AI assistance disclosure

Per CONTRIBUTING.md: Claude Code was used assistively throughout (diagnosis, drafting the fix and tests, this rebase). All changes were reviewed, executed and validated on real runs by the human author (full `cargo test`, Poppler cross-check on a real pandoc/xelatex PDF), who takes responsibility for correctness, licensing and provenance.

Happy to squash/rename or split further if you'd prefer.

### DIFF
--- a/src/editor/document_editor.rs
+++ b/src/editor/document_editor.rs
@@ -2459,6 +2459,45 @@ impl DocumentEditor {
                                         None
                                     };
 
+                                // Every Standard-14 base-font name that the overlay content
+                                // stream will actually emit via `Tf` needs a matching Font
+                                // object registered in this page's `/Resources/Font`, else the
+                                // resource name is dangling.
+                                //
+                                // The name to register is NOT the raw `font.name` the caller
+                                // passed — it's the transformed name the writer emits, i.e.
+                                // `map_base14_font_name(&font.name, is_bold)` (the same source
+                                // of truth used by `ContentStreamBuilder::map_font_name`). Keying
+                                // off the raw name diverges from the emitted `Tf` name for:
+                                //   - bold/italic overlays (raw "Helvetica" + bold → emits
+                                //     "/Helvetica-Bold");
+                                //   - generic families ("Arial", "sans-serif" → "/Helvetica");
+                                //   - Symbol/ZapfDingbats (routed to "/Helvetica").
+                                // Registering the emitted name closes all three; only Standard-14
+                                // names are ever emitted here (embedded fonts take the
+                                // ShowEmbeddedText path and never reach this registration).
+                                let mut overlay_font_ids: HashMap<String, u32> = HashMap::new();
+                                if let Some(added) = self.overlay_additions.get(&source_page_index)
+                                {
+                                    let mut needed: Vec<String> = added
+                                        .iter()
+                                        .filter_map(|el| match el {
+                                            ContentElement::Text(t) => {
+                                                Some(crate::writer::map_base14_font_name(
+                                                    &t.font.name,
+                                                    t.is_bold(),
+                                                ))
+                                            },
+                                            _ => None,
+                                        })
+                                        .collect();
+                                    needed.sort();
+                                    needed.dedup();
+                                    for name in needed {
+                                        overlay_font_ids.insert(name, self.allocate_object_id());
+                                    }
+                                }
+
                                 // Apply page property modifications if any
                                 let mut final_page_obj = if let Some(props) =
                                     self.modified_page_props.get(&source_page_index)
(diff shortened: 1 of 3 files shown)


yfedoseev:
Thanks for the rebase and the added tests, @thomnico — the diagnosis and fix shape both look right on a read-through. Two process items are blocking this from moving through review right now:

1. **CLA**: the `cla` check is failing — looks like the CLA hasn't been signed yet. Please sign via the link the CLA bot posts on this PR (see [CLA.md](../blob/main/CLA.md)) so that check goes green.
2. **DCO**: none of the commits carry a `Signed-off-by:` trailer, which `CONTRIBUTING.md` requires on every commit (enforced by the `dco` job once the full matrix runs). Easiest fix: `git commit --amend -s` and force-push (or `git rebase --exec 'git commit --amend --no-edit -s' <base>` if you split into multiple commits).

Also, so you don't spend time reconciling it blind: **must-fix #1 (multi-stream `/Contents` not fully replaced)** from #941 has already landed on `main` independently, in #942 — we hit the identical defect class through a different trigger (`set_text()`/`remove_element()` on a page whose original `/Contents` was already an array) while working a different issue, and merged a fix for it. That's why this PR is now showing a merge conflict. Worth rebasing onto current `main` and dropping/reconciling that hunk against what's already there; must-fix #2 (font registration keying off the emitted name, not the raw one) and the `overlay_additions` HashMap-overwrite fix plus the `/Resources` indirect-reference handling all still look like genuinely open, valuable fixes independent of that one.

Once CLA/DCO are sorted and it's rebased, happy to take another pass — will also run our corpus regression sweep against the editor-path changes before merge.

thomnico:

yfedoseev:
Follow-up after actually running the verification: read through the diff in full and ran our native corpus-signature sweep (`corpus_sig`, text hash + word-length + page-rotation per PDF) against a 419-PDF regression set, isolated at this PR's own fork point (v0.3.75) so it's not muddied by unrelated changes that have landed on `main` since — **zero text/word/rotation differences**. That's consistent with #941's own claim that this is confined to the editor/redaction write path.

On substance: bug #2 (font registration keyed off the emitted name, not the raw one) and bug #3 (`overlay_additions` HashMap overwrite discarding prior `save_page()` calls) both look like genuine, still-open fixes — not covered by anything already merged. Bug #1 remains the one to drop/reconcile per my comment above, since it's already fixed on `main`.

So: **happy to approve on substance** once the process items are sorted — CLA signature, DCO sign-off (`Signed-off-by:` trailer), and the rebase to resolve the current conflict with `main`. No further code concerns from this end beyond those.

thomnico:

thomnico:
Thanks @yfedoseev — all three items addressed:

1. **CLA**: signed (comment above).
2. **DCO**: the branch is now a single commit rebased onto current `main`, signed off (`Signed-off-by: thomnico <thomnico@googlemail.com>`).
3. **Rebase / #942 reconciliation**: I dropped my orphan-id-based `/Contents` hunk in favor of the #942 implementation (wholesale `[redacted, overlay_additions]` replacement tracked by object id) — it covers the redaction-path trigger too, and my multi-stream regression test (`destructive_redaction_multi_stream_contents_no_dangling_ref`) passes against it unchanged. What remains in this PR: must-fix #2 (font registration keyed off the emitted `map_base14_font_name` name), indirect + inherited `/Resources` handling, the `overlay_additions` accumulation fix, and the five regression tests.

Local gates on the rebased branch: `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean, full `cargo test` green (lib: 5779 passed / 0 failed; 40 integration binaries all ok, 418 tests, 0 failed). Poppler cross-check re-run (pdftotext/pdfinfo 26.07.0, exit 0, no warnings): redacted value absent from the raw bytes, bold replacement token extracted, `/Helvetica-Bold` declared under the emitted name.

PR description updated to reflect the post-rebase scope.

thomnico:
recheck

yfedoseev (approved):
Formal approval — code review, corpus verification (419-PDF sweep, zero regressions at this branch's own fork point), CLA, and DCO are all resolved per the comments above. Queued to merge once CI (currently rebuilding after the #944/#946 rebase in flight) is fully green.

PR #967, opened 2026-07-28
fix(965): remove_footers should persist when saving a file

### LINKED ISSUE #965: [Bug]: remove_footers does not persist when saving a file
### Bug Description

remove_footers (and remove_headers / remove_artifacts) claim that spans were erased (and they are for to_markdown), the caller reasonably believes the header/footer is gone, and saves the file — and the output file is byte-for-byte unmodified with respect to that content.

I found this by chasing down why a specific PDF that appeared to have correctly removed headers reported 0 were removed, which was, of course, because 0 were removed.

I'm working on a fix.  Reporting for visibility and welcome feedback.

### Steps to Reproduce

1. save a file with code below into examples with name `demo_erase_not_persisted.rs`
2. run with `cargo run --example demo_erase_not_persisted`

```
//! Demonstrates: `remove_footers`'s `erase_region`-based removal does
//! not actually remove anything in the PDF. It only filters spans returned by
//! `extract_spans`/`extract_text`/`to_markdown` on the SAME in-memory
//! `PdfDocument` — `DocumentEditor::from_document` does not copy over
//! `PdfDocument`'s `erase_regions` field, so nothing survives the round
//! trip through `save_to_bytes()`.
//!
//! Uses a plain, untagged, position-locked recurring footer ("Confidential
//! Draft Only" on every page) — the case pass 2's exact-literal heuristic
//! already handles correctly today (`removed > 0`)

use pdf_oxide::editor::DocumentEditor;
use pdf_oxide::PdfDocument;

fn obj(buf: &mut Vec<u8>, off: &mut [usize], id: usize, body: &str) {
    off[id] = buf.len();
    buf.extend_from_slice(format!("{id} 0 obj\n{body}\nendobj\n").as_bytes());
}

fn stream(buf: &mut Vec<u8>, off: &mut [usize], id: usize, data: &[u8]) {
    off[id] = buf.len();
    buf.extend_from_slice(format!("{id} 0 obj\n<< /Length {} >>\nstream\n", data.len()).as_bytes());
    buf.extend_from_slice(data);
    buf.extend_from_slice(b"\nendstream\nendobj\n");
}

fn build_pdf(page_count: usize) -> Vec<u8> {
    let mut buf: Vec<u8> = Vec::new();
    let mut off = vec![0usize; 4 + page_count * 2];
    buf.extend_from_slice(b"%PDF-1.7\n%\xE2\xE3\xCF\xD3\n");
    obj(&mut buf, &mut off, 1, "<< /Type /Catalog /Pages 2 0 R >>");
    let kids: String = (0..page_count)
        .map(|i| format!("{} 0 R", 5 + i * 2))
        .collect::<Vec<_>>()
        .join(" ");
    obj(
        &mut buf,
        &mut off,
        2,
        &format!("<< /Type /Pages /Kids [{kids}] /Count {page_count} >>"),
    );
    obj(
        &mut buf,
        &mut off,
        3,
        "<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica /Encoding /WinAnsiEncoding >>",
    );
    for i in 0..page_count {
        let content_id = 4 + i * 2;
        let page_id = 5 + i * 2;
        let content = format!(
            "BT /F1 12 Tf 1 0 0 1 72 400 Tm (Body text placeholder) Tj ET\n\
             BT /F1 10 Tf 1 0 0 1 72 30 Tm (Confidential Draft Only) Tj ET\n"
        );
        stream(&mut buf, &mut off, content_id, content.as_bytes());
        obj(
            &mut buf,
            &mut off,
            page_id,
            &format!(
                "<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792] \
                 /Resources << /Font << /F1 3 0 R >> >> /Contents {content_id} 0 R >>"
            ),
        );
    }
    let xref_off = buf.len();
    let total_objs = off.len();
    buf.extend_from_slice(format!("xref\n0 {}\n", total_objs).as_bytes());
    buf.extend_from_slice(b"0000000000 65535 f \n");
    for offset in &off[1..] {
        buf.extend_from_slice(format!("{:010} 00000 n \n", offset).as_bytes());
    }
    buf.extend_from_slice(
        format!(
            "trailer\n<< /Size {} /Root 1 0 R >>\nstartxref\n{}\n%%EOF\n",
            total_objs, xref_off
        )
        .as_bytes(),
    );
    buf
}

fn main() {
    let page_count = 5;
    println!("building PDF with {page_count} pages");
    let bytes = build_pdf(5);
    let doc = PdfDocument::from_bytes(bytes).expect("open");

    let removed = doc.remove_footers(0.5).expect("remove_footers");
    println!("remove_footers() reported {removed} spans erased");
    assert!(removed > 0, "expected a real erasure to happen here");

    let text_after_in_memory = doc.extract_text(1).unwrap();
    println!(
        "extract_text on the same PdfDocument instance, page 1 shows no footer: {:?}",
        text_after_in_memory
    );
    assert!(
        !text_after_in_memory.contains("Confidential Draft Only"),
        "sanity check: erase_region should have hidden it in-memory"
    );

    println!("\n--- now round-tripping through DocumentEditor::from_document(...).save_to_bytes() ---\n");

    let mut editor = DocumentEditor::from_document(doc).expect("from_document");
    let saved_bytes = editor.save_to_bytes().expect("save_to_bytes");

    let raw_contains_footer = String::from_utf8_lossy(&saved_bytes).contains("Confidential Draft Only");
    println!(
        "Raw saved-bytes contain the literal footer text: {}",
        raw_contains_footer
    );

    let reopened = PdfDocument::from_bytes(saved_bytes).expect("reopen saved bytes");
    let text_from_saved_file = reopened.extract_text(1).unwrap();
    println!("extract_text on the re-opened saved file, page 1: {:?}", text_from_saved_file);

    if text_from_saved_file.contains("Confidential Draft Only") {
        println!(
            "\n*** BUG: remove_footers() reported {removed} spans erased, \
             but the footer is still fully present in the saved/exported file. ***"
        );
    } else {
        println!("\n(Footer was NOT present in the saved file — erasure did persist.)");
    }
}
```

### Expected Behavior

output:
```
building PDF with 5 pages
remove_footers() reported 5 spans erased
extract_text on the same PdfDocument instance, page 1 shows no footer: "Body text placeholder"

--- now round-tripping through DocumentEditor::from_document(...).save_to_bytes() ---

Raw saved-bytes contain the literal footer text: false
extract_text on the re-opened saved file, page 1: "Body text placeholder"

*** BUG: remove_footers() reported 5 spans erased, but the footer is still fully present in the saved/exported file. ***
```

### Actual Behavior

output
```
building PDF with 5 pages
remove_footers() reported 5 spans erased
extract_text on the same PdfDocument instance, page 1 shows no footer: "Body text placeholder"

--- now round-tripping through DocumentEditor::from_document(...).save_to_bytes() ---

Raw saved-bytes contain the literal footer text: true
extract_text on the re-opened saved file, page 1: "Body text placeholder\n\n\nConfidential Draft Only"

*** BUG: remove_footers() reported 5 spans erased, but the footer is still fully present in the saved/exported file. ***
```

_No response_

### Environment

- **OS**: macOS 15.7.3
- **Rust version**: rustc 1.95.0 
- **pdf_oxide version**:  recent main 6bc3923bf56531f1bd2b04aac0a68a5ffae8c7c6

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

found while trying to refactor https://github.com/yfedoseev/pdf_oxide/pull/795/changes -- I pulled out a single test which has nothing to do with parity pages yet where the fix was required to get recto/verso header change to work. 

This test fails in main, passes on the branch, and I sought to address this in isolation
https://github.com/ultrasaurus/pdf_oxide/blob/a504587632914102e5ef3c07d16e56cacd504d12/tests/test_remove_footers.rs#L97

### Checklist

- [x] This report is written in my own words (not AI-generated)

ultrasaurus:
remove_footers (and remove_headers / remove_artifacts) claim that spans were erased (and they are for to_markdown), the caller reasonably believes the header/footer is gone, and saves the file — and the output file was byte-for-byte unmodified with respect to that content.

Now whatever was removed in the PDFDocument is removed when DocumentEditor::from_document is called.

## Linked issue

Closes https://github.com/yfedoseev/pdf_oxide/issues/965

## What and why

Initially I just wanted to use it for testing.  Practically, it could be useful in a production pipeline where I will assemble PDFs from different sources into a book / website and want to use PDFDocument to_html or to_markdown to present each PDF as a webpage and then use DocumentEditor to decorate the header/footer in some unified way.

Fundamentally, it's just how I expected it to work.

## Type of change

- [X] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [X] I added a test that **fails before this change and passes after**
- [X] Test named by defect **class**, not an issue/PR number; no
- [X] `cargo test` passes, **and** the affected feature tiers:
- [X] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Regression on real PDFs

This is a change of behavior.  Before (on main) you can remove_artifacts/footers/headers and that has no effect on the PDF written by DocumentEditor.  After the change, on this branch, artifacts will be removed.

This is in progress.

**Corpus I tested** Corpus: 31 PDFs total
* 20 IRS tax forms
* 10 personal-library documents
* 1 large real-world scan — warandpeace030164mbp.pdf, 725 pages, from archive.org

Format/source diversity: government forms (dense tables, form fields), academic/magazine layout (running headers/footers, multi-column), and a large OCR'd scan (noisy positioning) — three fairly different document-generation pipelines.

For this specific PR (persistence fix): only 6 of the 31 actually exercise the code path (had removed > 0 from remove_artifacts at threshold 0.8)

- [x] Ran `corpus_sig` and diffed my branch against **`main`** — no unintended
      scripts, or crashes).
- [x] Diffed my branch against the **latest release** (`v0.3.__`) — same.
      char-Levenshtein.

**Diff summary vs `main`:** 
zero diff against main

**Diff summary vs latest release:** 
zero diff against v0.3.77

## AI assistance disclosure

- [X] AI assistance:**assisted** — tool: Claude, extent: draft code and corpus testing
- [X] I understand and can explain every line; the description and my review
      predominantly AI-generated.

## Checklist

- [X] One logical change (no bundled refactor/perf/correctness).
- [X] Commits follow Conventional Commits and are **DCO signed-off** (`git commit -s`).
      CHANGELOG (not in code).
- [ ] Public-API changes considered for semver (`semver-checks` will run).

### DIFF
--- a/src/editor/document_editor.rs
+++ b/src/editor/document_editor.rs
@@ -2,6 +2,7 @@
 //!
 //! Provides the DocumentEditor type for modifying PDF documents.
 
+use crate::cache::MutexExt;
 use crate::document::PdfDocument;
 use crate::editor::form_fields::FormFieldWrapper;
 use crate::editor::resource_manager::ResourceManager;
@@ -480,8 +481,19 @@ pub struct DocumentEditor {
     modified_annotations: HashMap<usize, Vec<crate::editor::dom::AnnotationWrapper>>,
     /// Modified page properties (rotation, boxes)
     modified_page_props: HashMap<usize, ModifiedPageProps>,
-    /// Erase regions per page (whiteout overlays)
+    /// Erase regions per page (whiteout overlays). Populated by the public
+    /// `erase_region`/`erase_regions` API (cosmetic-only, by contract — see
+    /// their doc comments) AND, for any rect also present in
+    /// `inherited_erase_regions`, as the default/fallback rendering for a
+    /// region inherited via `from_document` until/unless save-time
+    /// destructive redaction removes it.
     erase_regions: HashMap<usize, Vec<[f32; 4]>>,
+    /// Regions inherited via `from_document` (see its doc comment) that
+    /// save-time destructive redaction should attempt to actually remove.
+    /// Kept separate from `erase_regions` so a direct caller of the public
+    /// `erase_region`/`erase_regions` API — which is documented as
+    /// cosmetic-only — is never silently upgraded to real content removal.
+    inherited_erase_regions: HashMap<usize, Vec<[f32; 4]>>,
     /// Pages where annotations should be flattened
     flatten_annotations_pages: std::collections::HashSet<usize>,
     /// Pages where redactions should be applied
@@ -509,6 +521,10 @@ pub struct DocumentEditor {
     remove_acroform: bool,
     /// Warnings collected during form flattening (e.g. widgets with no /AP stream)
     flatten_warnings: Vec<String>,
+    /// Structured warnings collected during editing/save (distinct from
+    /// `flatten_warnings`, which predates this and is a `&[String]`
+    /// form-flattening-only log — see [`Self::structured_warnings`]).
+    structured_warnings: Vec<crate::extractors::warnings::Warning>,
     /// Embedded files to add to the document
     embedded_files: Vec<crate::writer::EmbeddedFile>,
     /// Modified or new form fields (field name → wrapper)
@@ -628,6 +644,7 @@ impl DocumentEditor {
             modified_annotations: HashMap::new(),
             modified_page_props: HashMap::new(),
             erase_regions: HashMap::new(),
+            inherited_erase_regions: HashMap::new(),
             flatten_annotations_pages: std::collections::HashSet::new(),
             apply_redactions_pages: std::collections::HashSet::new(),
             redaction_regions: HashMap::new(),
@@ -637,6 +654,7 @@ impl DocumentEditor {
             flatten_forms_pages: std::collections::HashSet::new(),
             remove_acroform: false,
             flatten_warnings: Vec::new(),
+            structured_warnings: Vec::new(),
             embedded_files: Vec::new(),
             modified_form_fields: HashMap::new(),
             deleted_form_fields: HashSet::new(),
@@ -646,10 +664,38 @@ impl DocumentEditor {
     }
 
     /// Open a PDF document for editing from an existing PdfDocument object.
+    ///
+    /// Any regions already staged via `source.erase_region()` (e.g. by
+    /// `remove_headers`/`remove_footers`/`remove_artifacts`) are carried
+    /// over so they aren't silently lost — see [`Self::save_to_bytes`] for
+    /// how they're actually applied. This step is a plain data copy, so it
+    /// cannot fail: staging a region never makes `from_document` itself
+    /// fallible, even for a page whose redaction will later be refused.
     pub fn from_document(mut source: PdfDocument) -> Result<Self> {
         let page_count = source.page_count()?;
         let next_id = Self::find_max_object_id(&source) + 1;
         let page_order: Vec<i32> = (0..page_count as i32).collect();
+
+        // Capture staged erase regions before `source` moves into `Self`.
+        // Populate BOTH fields: `erase_regions` so the region renders as its
+        // existing cosmetic overlay by default (same as any other overlay
+        // entry) until/unless save-time redaction removes it, and
+        // `inherited_erase_regions` as the marker that this particular
+        // region — unlike one a caller adds directly via the public
+        // `erase_region`/`erase_regions` API — is eligible for that
+        // destructive upgrade at save time.
+        let inherited: HashMap<usize, Vec<[f32; 4]>> = source
+            .erase_regions
+            .lock_or_recover()
+            .iter()
+            .map(|(&page, rects)| {
+                let converted = rects
+                    .iter()
+                    .map(|r| [r.x, r.y, r.x + r.width, r.y + r.height])
+                    .collect();
+                (page, converted)
+            })
+            .collect();
 
         Ok(Self {
             source,
@@ -667,7 +713,8 @@ impl DocumentEditor {
             structure_modified: false,
             modified_annotations: HashMap::new(),
             modified_page_props: HashMap::new(),
-            erase_regions: HashMap::new(),
+            erase_regions: inherited.clone(),
+            inherited_erase_regions: inherited,
             flatten_annotations_pages: std::collections::HashSet::new(),
             apply_redactions_pages: std::collections::HashSet::new(),
             redaction_regions: HashMap::new(),
@@ -677,6 +724,7 @@ impl DocumentEditor {
             flatten_forms_pages: std::collections::HashSet::new(),
             remove_acroform: false,
             flatten_warnings: Vec::new(),
+            structured_warnings: Vec::new(),
             embedded_files: Vec::new(),
             modified_form_fields: HashMap::new(),
             deleted_form_fields: HashSet::new(),
@@ -712,6 +760,7 @@ impl DocumentEditor {
             modified_annotations: HashMap::new(),
             modified_page_props: HashMap::new(),
             erase_regions: HashMap::new(),
+            inherited_erase_regions: HashMap::new(),
             flatten_annotations_pages: std::collections::HashSet::new(),
             apply_redactions_pages: std::collections::HashSet::new(),
             redaction_regions: HashMap::new(),
@@ -721,6 +770,7 @@ impl DocumentEditor {
             flatten_forms_pages: std::collections::HashSet::new(),
             remove_acroform: false,
             flatten_warnings: Vec::new(),
+            structured_warnings: Vec::new(),
             embedded_files: Vec::new(),
             modified_form_fields: HashMap::new(),
             deleted_form_fields: HashSet::new(),
@@ -1888,6 +1938,74 @@ impl DocumentEditor {
         // silently produces a structurally valid but unreadable PDF.
         if self.source.is_encrypted() && !self.source.is_authenticated() {
             return Err(Error::EncryptedPdf);
+        }
+
+        // Apply any regions inherited via `from_document`, now as real,
+        // destructive redactions. If a page's redaction is refused, leave
+        // that page's cosmetic overlay in place as a fallback, other errors
+        // still fail the save.
+        if !self.inherited_erase_regions.is_empty() {
+            let staged: Vec<(usize, Vec<[f32; 4]>)> = self
+                .inherited_erase_regions
+                .iter()
+                .map(|(&page, rects)| (page, rects.clone()))
+                .collect();
+            let redact_opts = crate::redaction::RedactionOptions {
+                draw_overlay_when_no_ic: false,
+                ..crate::redaction::RedactionOptions::default()
+            };
+            for (source_page, rects) in staged {
+                let entry = self.redaction_regions.entry(source_page).or_default();
+                let before_len = entry.len();
+                for rect in &rects {
+                    entry.push(crate::redaction::RedactionRegion::from_rect(
+                        rect[0], rect[1], rect[2], rect[3], None,
+                    ));
+                }
+                let was_already_queued = self.apply_redactions_pages.contains(&source_page);
+                self.apply_redactions_pages.insert(source_page);
+
+                match self.apply_redactions_destructive_for_page(source_page, &redact_opts) {
+                    Ok(_) => {
+                        // Removal succeeded — drop exactly these rects from
+                        // the cosmetic-overlay field, leaving any rect a
+                        // caller added directly via the public
+                        // `erase_region`/`erase_regions` API untouched.
+                        if let Some(v) = self.erase_regions.get_mut(&source_page) {
+                            v.retain(|r| !rects.contains(r));
+                            if v.is_empty() {
+                                self.erase_regions.remove(&source_page);
+                            }
+                        }
+                    },
+                    Err(Error::Unsupported(reason)) => {
+                        // Refused. Undo exactly what this step added,
+                        // any caller-supplied redaction_regions/
+                        // apply_redactions_pages state is untouched, and
+                        // keep the cosmetic overlay as the fallback.
+                        if let Some(v) = self.redaction_regions.get_mut(&source_page) {
+                            v.truncate(before_len);
+                            if v.is_empty() {
+                                self.redaction_regions.remove(&source_page);
+                            }
+                        }
+                        if !was_already_queued {
+                            self.apply_redactions_pages.remove(&source_page);
+                        }
+                        self.structured_warnings.push(crate::extractors::warnings::Warning {
+                            category:
+                                crate::extractors::warnings::WarningCategory::RedactionOverlayFallback,
+                            page: Some(source_page),
+                            message: format!(
+                                "staged erasure could not be applied as a destructive \
+                                 redaction, falling back to a cosmetic overlay: {reason}"
+                            ),
+                            spec_section: None,
+                        });
+                    },
+                    Err(e) => return Err(e),
+                }
+            }
         }
 
         /// Compress a stream object with FlateDecode if it has no filter yet.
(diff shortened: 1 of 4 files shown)


yfedoseev (commented):
Thanks for this — the diagnosis is right, the write-up is clear, and the test is properly revert-checked. I verified it locally: your test fails on `main` and passes on the branch, `cargo test --lib` is 5785 green, and a 419-PDF extraction sweep against `main` shows zero change, so nothing that doesn't call `erase_region` is affected.

I also ran the save round trip your change actually touches, over the same 419-PDF corpus: open → `remove_artifacts(0.8)` → `DocumentEditor::from_document` → `save_to_bytes` → reopen, then compare the reopened text against what the in-memory (filtered) document reports. The invariant I'd expect from this PR is that those two agree.

**Control first:** the same round trip *without* any erasure staged is clean — 196 documents, zero text change, zero errors. So the differences below come from the new redaction path, not from round-trip noise.

**With erasures staged:** 21 documents exercised the path, 8 of them disagreed with the in-memory view. Two representative cases:

`long_tiny_text/20_pg18_pg1.pdf` (`removed=6`): unfiltered 1173 words, in-memory filtered 1168, saved file **1173** — the saved file matches the *unfiltered* count. The words present in the file but absent from the filtered view are exactly the erased watermark: `РЕЛИЗ ПОДГОТОВИЛА ГРУППА`, `"What's News"`, `VK.COM/WSNWS`. So on this document the erasure still doesn't survive the save.

`multi_column/027c390510b2fd3364f5a382f2bc304af018_page_4_pg1.pdf` (`removed=6`): unfiltered 796, filtered 779, saved **776** — three fewer than the filtered view, i.e. slightly more removed than was staged.

To be clear, the first case is **not a regression** — `main` leaves that watermark in the file too. It's a coverage gap: the promise "whatever was removed in the PdfDocument is removed when `from_document` is called" doesn't hold for every span `erase_region` hides. My guess is the redaction rect and the span geometry disagree for transformed/rotated runs, but I haven't chased it down and it's your change, so I'd rather report it than guess in your branch. Worth a look at whether `add_redaction` covers the same glyphs `erase_region` filters.

Some of the per-word diffs are hyphenation re-joins (`sig-` + `nificantly` vs `significantly`) from re-extracting a rewritten content stream, which is expected and not a problem — I filtered on the substantive cases above.

Two smaller notes:

1. The test asserts the footer is gone but never asserts the body survives, so a change that wiped the whole page would pass it. I checked and the body *does* survive (all 5 pages keep `Body text placeholder`), so this is just a missing assertion rather than a defect — worth adding.
2. `from_document` now performs a destructive content-stream rewrite during construction, and gains new failure modes as a result (`add_redaction`/`apply_redactions_destructive` both return `Result`). That's a real API-shape question rather than a code problem — I'd like @yfedoseev to weigh in on whether erasure persistence belongs in the constructor or in an explicit step before this merges.

Not blocking on my side beyond that; the direction looks right to me.

ultrasaurus:
Thanks for confirming the direction and pointing out the potential implication of an API signature change.  I'd like to suggest a middle ground...  keep the API, don't do the full redact and document the limitation.  It's a risk that someone thinks the text was fully redacted, so would be open to backing off to just a doc update.

`from_document` could simply ensure that `source`'s staged erasures aren't dropped. It can do that by copying `source.erase_regions` into the new `DocumentEditor`'s *own* `erase_regions` field (the cosmetic-overlay mechanism already exposed via `erase_region`/`erase_regions` — a plain data copy, not a content-stream rewrite, so it can't fail. 

The tradeoff: by default, `DocumentEditor::from_document(doc).save_to_bytes()` after `remove_footers()` now produces a file where the footer is visually covered but still present and extractable underneath — not the full removal I expected when I wrote up bug #965. Callers who need true removal call the existing fallible primitives themselves:

```rust
let mut editor = DocumentEditor::from_document(doc)?; // now infallible w.r.t. redaction
editor.apply_redactions_destructive(RedactionOptions {
    draw_overlay_when_no_ic: false,
    ..Default::default()
})?; // caller opts in, handles the Result
editor.save_to_bytes()?;
```

Adding a code snippet like that to the docs would be nice.

will wait to hear back from @yfedoseev

yfedoseev:
@ultrasaurus — following up on this one since it's been sitting in draft. Your PR body said the corpus verification was in progress, with only a handful of the PDFs you had actually exercising the path, which was a fair reason to keep it draft.

I've now covered that surface from our side. A text-only comparison can't see a write-path defect at all (the in-memory answer is right while the saved file is wrong), so I ran a save → **re-parse** round-trip over 112 PDFs from our corpus, comparing page count, extracted text and a content hash *after reopening the saved bytes*. Result: verdicts byte-identical to `main` — 97 `OK`, plus 15 encrypted documents that fail identically on both sides. Zero differing lines.

That's the surface this PR is actually about, so the gap that justified draft status looks closed from our end. Whether it's ready is still your call — happy to leave it draft if you have more you want to do, but if you're satisfied with it, mark it ready and I'll put it in the merge queue.

Separately, and not a defect in this PR: those 15 documents show that the write path can't round-trip an encrypted document at all — `DocumentEditor::from_document` and `save_to_bytes` both refuse with "requires a password" even after a successful `authenticate()`. That's pre-existing on `main` and unchanged by your change. I'll file it separately.

ultrasaurus:
@yfedoseev It's sitting in draft because of the "real API-shape question" -- I proposed an alternate approach to the code in this draft and was waiting to hear back on that

yfedoseev:
@ultrasaurus — you're right that this was waiting on me, and I'm sorry: you raised the API-shape question on 29 July and my last comment answered a different one (corpus coverage) that you hadn't asked about. Here's a decision.

**Short version:** I want your API fix and not your default. `from_document` should stay infallible — a constructor doing fallible content-stream rewriting is the wrong shape, and you were right to push back on it. But the regions shouldn't become cosmetic overlays either.

**The reason is an invariant, not a security position.** `PdfDocument::erase_region` is documented as *"content in this region will be excluded from all subsequent text extractions"* — it's an extraction-time mask. `DocumentEditor::erase_region` is documented as *"the original content is not removed but hidden beneath the white overlay."* Same method name, opposite meaning. Copying the regions across silently converts "not extracted" into "still extracted, visually hidden", so `extract_text()` on the reopened file returns the footer again.

That's what makes the overlay default worse than the status quo rather than a safe middle ground. Today the removal is lost on save and you can see it's lost. With the overlay the file looks right and extracts wrong — nobody goes looking. We've fixed that shape of bug twice recently (#1032, #1010) and both times the answer was to make the write path tell the truth.

**The shape I'd like:**

1. `from_document` copies the staged regions — plain data copy, infallible, exactly as you proposed.
2. `save()` / `save_to_bytes()` — already fallible — performs the actual removal. The fallibility lands where a `Result` already exists, so no signature changes anywhere.
3. Where removal genuinely can't be done safely (the composite/Type0 case `apply_redactions_destructive` already refuses), fall back to the overlay **and emit a warning** through `WarningCategory`. Degraded, never silent.
4. `apply_redactions_destructive` stays the explicit, fail-closed path for real redaction.

This isn't a new design — it's the two-phase model the format itself defines. ISO 32000-2 §12.5.6.23 separates marking from applying precisely so the reversible step and the irreversible one stay distinct, and that's the same split PyMuPDF ships as `add_redact_annot()` → `apply_redactions()`. We already have both halves: `remove_footers()` is the mark, `apply_redactions_destructive` is the apply. The only thing missing is the connection at save.

**Two practical notes.** #1013 is in the merge queue now and makes `WarningCategory` `#[non_exhaustive]`, so adding a variant becomes a non-breaking change once it lands — worth sequencing after it. And the acceptance test for #965 should be a save → **re-parse** → compare-extracted-text round-trip; checking the in-memory editor after save proves nothing about the bytes. I have a 112-document harness of exactly that shape and can point it at this branch when you're ready.

Separately: those two identically-named `erase_region` methods with opposite semantics are worth renaming or at least cross-documenting regardless of which way this goes. That collision is most of why this was ambiguous in the first place, and that's on the codebase, not on you.

ultrasaurus:
new approach:

If remove_headers/remove_footers/remove_artifacts are called before
`DocumentEditor::from_document` then copy regions that are staged
for removal (so `from_document` remains infallible), then do the
destructive redaction at save time (`write_full_to_writer`).

Where the redaction can be applied safely, the region is genuinely
gone from the saved bytes. Where it's refused the page falls back
to its existing cosmetic overlay instead of failing the whole save,
and a new `WarningCategory::RedactionOverlayFallback` is recorded via
`DocumentEditor::structured_warnings()` so callers can detect the
degraded case.

Regression tests re-parse the saved output rather than trusting
in-memory editor state:
- from_document stays Ok even when a page's redaction will later be
  refused
- a redactable page's erased region is confirmed gone after save +
  reopen, with the untouched body text confirmed to survive alongside it
- a refused page's save still succeeds and records the fallback warning

Still need to re-run corpus, skipping PDF that fails due to: https://github.com/yfedoseev/pdf_oxide/issues/1107

yfedoseev:
Short note so this does not sit on a stale assumption.

The intake automation that closed #1108 twice is fixed on `main` (188322f7). Two of those bugs would have applied here too: the gates were being applied retroactively to PRs opened before the policy existed, and this one dates from 28 July. It is now exempt, so **this draft is not at risk of being auto-closed**, including when you next push to it.

On sequencing: you said on #1107 that you wanted that bug fixed before submitting a revised version of this. That chain is unblocked at the automation level now — #1108 can be reopened whenever you like, and it no longer has to wait for #976 to resolve first, because it too predates the policy. Your original preference for #976 first still stands unless you say otherwise; I am not reordering it for you.

Nothing on my side is blocking this PR. When you do pick it back up, the round-trip test and the persistence probe you added are the right shape for it — a fix about what survives a save can only be proved by reopening the saved bytes, which is exactly what they do. That is the part most PRs in this area get wrong, so it is worth saying explicitly rather than leaving as an unremarked good decision.

yfedoseev:
Separately from the automation note above — I have now reviewed the implementation itself, since you had asked for a read on the approach rather than just the corpus.

The shape is right, and the part I would most have expected to get collapsed is the part you got right: keeping `erase_regions` and `inherited_erase_regions` separate, so a caller of the documented cosmetic-only API is never silently upgraded to real content removal. A three-line version of this fix would have merged those and quietly changed what a published API means.

**One thing to fix.** `save_to_bytes` reads the inherited map and, on success, removes the consumed rects from `erase_regions` — but never clears the inherited map itself. A second `save_to_bytes()` re-stages every region, pushes duplicate `RedactionRegion`s, and re-runs the destructive pass over content the first save already removed. Save-to-bytes-then-save-to-file, or a retry after a recoverable error, both hit it.

I had drafted a suggestion that you drain the map as you consume it, then talked myself out of it — worth saying why, since it is the obvious fix. Draining makes the first and second save produce *different files with no error*, which is a worse failure than the double-apply it cures. The property to want is idempotence: repeated saves, identical bytes. That points at keeping the staged regions as immutable intent and memoising the applied outcome, so applying twice is a cache hit rather than a second mutation. For this PR, making a second save a no-op is enough; the general property and the other places it is violated are #1122. A save-twice test asserting byte-equality would pin it — nothing currently saves twice.

**Smaller note.** `v.retain(|r| !rects.contains(r))` also drops a rect a caller staged directly if it happens to share coordinates with an inherited one, though the comment says those are left untouched. Benign in effect, since the content is gone either way — but `inherited_and_direct_erase_regions_on_same_page_are_handled_independently` uses deliberately distinct rects, so the collision case is not covered. Worth noting `Rect` is not `Eq`/`Hash` and `RedactionRegion::from_rect` can produce non-finite geometry, so value equality is fragile as identity here.

**On your probe:** I pointed it at our corpus and it does the job — it exercises the save path on a real population and finds no genuine mismatch. The two it flags are false positives from its own metric: it diffs word *sets* across two independent extractions, and saving re-chunks text runs, so a word the source emitted joined comes back split (`established` → `es-` + `tablished.`) and trips the set difference with nothing actually changed. Squashing whitespace and hyphen breaks before comparing clears both without weakening it. Worth fixing, since as written it reports a failure on a correct save.

On sequencing: you had said you wanted #1107 fixed before revising this. Your fix for it — #1108 — is still closed, and that was our gate, not anything about the change. #1114 fixed the automation, and I have checked that #1108 predates the policy epoch, so reopening it will not trip the same rule again even while #976 is open. One-click reopen whenever you want it; nothing further needed from you to make it stick.

PR #976, opened 2026-07-29
fix(794): partial, remove page number when span also has other text

ultrasaurus:
removes page numbers like "1 com", "2 com", test cases motivated by real world examples from
bug #794 and IRS_Form_1120_2024.pdf (which defeated https://github.com/yfedoseev/pdf_oxide/pull/795)

Fixed by adding position consistency logic to upstream heuristics used for tagging in
- ensure_running_artifact_signatures
- mark_running_artifact_spans

improved artifact tagging where omissions, which were part of what seemed needed
in order to fix the removal (was part of pass 1 in `remove_repeated_text`, now part of pass 2,
so in retrospect a different approach could have been taken)

`remove_repeated_text` ignored labeled artifacts that are removed by `extract_spans`
so they weren't included in the count, which caused a test to fail prematurely 
when artifacts were correctly removed
(`remove_footers_strips_combined_page_number_and_brand_span`).

updated `example/extract_markdown_clean.rs`:
- writes markdown to file, prints removal count

test_artifact_mistagging.rs:
- `distinct_form_line_item_labels_are_not_tagged_as_pagination`
- `coincidentally_colocated_form_line_item_labels_are_not_tagged_as_pagination`

test_remove_footer.rs:
- `remove_footers_strips_combined_page_number_and_brand_span`

New shared test helpers: `tests/common/mod.rs`

New probe:
`probe_remove_artifacts_corpus.rs`

## Linked issue

https://github.com/yfedoseev/pdf_oxide/issues/794

Closes #

## What and why

Geometric location that is persistent over pages is a clear signal for page chrome.  In order to solve for IRS_Form_1120_2024.pdf where real form content labels "1a" and "2a" were very close in position on different pages and there were other spans that normalized to "#a" causing them to meet the page threshold. By grouping them in clusters by position, I could use the page threshold to distinguish them.

Initially, I thought it was a requirement to fix the problem "upstream" in 
- ensure_running_artifact_signatures and
- mark_running_artifact_spans;
however, I later realized that it was safer to address in "pass 2" of remove_repeated_text. 

I sought to de-couple this fix from the parity page fix I had been working on (see https://github.com/yfedoseev/pdf_oxide/pull/795), where this fix had become a dependency. I started with one failing test and through reproducing cases found in corpus testing and isolating various failures, I added additional test cases and fixed various artifact tagging bugs. 

## Type of change

- [X] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [X] I added a test that **fails before this change and passes after**
- [X] Test named by defect **class**, not an issue/PR number; no
- [X] `cargo test` passes, **and** the affected feature tiers: I don't think these are affected (are they?)
- [X] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Regression on real PDFs

**Corpus I tested** Corpus: 32 PDFs total
* 20 IRS tax forms
* 9 personal-library documents
* 2 exported from Google Docs (public domain text w/ page headers, footers)
* 1 large real-world scan — warandpeace030164mbp.pdf, 725 pages, from archive.org

Format/source diversity: government forms (dense tables, form fields), academic/magazine layout (running headers/footers, multi-column), and a large OCR'd scan (noisy positioning) — multiple different document-generation pipelines.

- [X] Ran `corpus_sig` and diffed my branch against **`main`** — no unintended
      scripts, or crashes).
- [X] Diffed my branch against the **latest release** (`v0.3.77` == main) — same.
- [n/a] Judged with a structural metric (word-Jaccard + spacing outliers), not
      char-Levenshtein.

**Diff summary vs `main`:** none

**Diff summary vs latest release:** none

## AI assistance disclosure

- [X] AI assistance: **none**, or **assisted** — tool: Claud, extent: draft code, wrote probe and ran corpus testing
- [X] I understand and can explain every line; the description and my review
      predominantly AI-generated.

## Checklist

- [X] One logical change (no bundled refactor/perf/correctness).
- [X] Commits follow Conventional Commits and are **DCO signed-off** (`git commit -s`).
- [x] Docs/`CHANGELOG.md` updated if user-facing; reporters credited in the
      CHANGELOG (not in code). -- I don't think this applies?
- [n/a] Public-API changes considered for semver (`semver-checks` will run).


ultrasaurus:
updated probe to include `ConversionOptions { include_artifacts: false` since it was written before that default was changed in main.

Full re-run results with the corrected probe (`extract_text` with `include_artifacts: false`, isolating this tagging fix specifically):

| File | main chars tagged | HEAD chars tagged |
|---|---|---|
| `IRS_Form_1120_2024.pdf` | 6 | 0 (fixed — `"1a"`/`"2a"`) |
| `IRS_Form_990_2024.pdf` | 64 | 0 (fixed — `" (2025)"` position-drift, `"(14)"` one-off) |
| all other 30 corpus PDFs | — | 0 diff |

Two real false-positive fixes confirmed, zero regressions, on the corrected probe. 

close look at results from IRS_Form_990_2024.pdf:
- `main` wrongly tags `" (2025)"` as `Pagination(Other)` on 8 pages — but its y-position varies wildly across occurrences: `13.83`, `25.83`, `37.83`, `61.83` — clearly not position-locked chrome, just a form-footer annotation whose vertical position shifts with how much content sits above it on each page.
- `main` also wrongly tags `"(14)"` (page 6) — a one-off, almost certainly a footnote/citation number, not a page number.
- this branch correctly tags neither, since no single position-cluster among these scattered occurrences clears the recurrence threshold — exactly the fix we built.

yfedoseev:
@ultrasaurus — thanks for laying out the design trade-off so explicitly in the description; that made this easy to evaluate properly rather than guess at. I want to answer the open question directly, because I think the retrospective instinct in the PR body points the wrong way, and I'd rather say so with the evidence than leave it ambiguous.

**Keep the upstream fix.** You wrote that you initially thought the fix belonged upstream in `ensure_running_artifact_signatures` / `mark_running_artifact_spans`, then concluded it was safer in pass 2 of `remove_repeated_text`. Pass 2 is narrower, but narrower here means fixing the symptom and leaving a worse bug in place, because the mis-tag causes silent content loss on paths that never call `remove_repeated_text` at all:

1. **Markdown/HTML/IR drop the content outright.** `src/converters/pdf_to_ir.rs:132` does
   ```rust
   spans.retain(|s| !is_page_artifact(s) && !is_geometric_footer(s, page_h_for_filter));
   ```
   and `is_page_artifact` matches `Some(ArtifactType::Pagination(_))` — any subtype, including the `Other` these labels receive. So a mis-tagged `1a`/`2a` is deleted at the IR boundary: `to_markdown()` and `to_html()` lose real form rows, with no removal call in the picture.
2. **Structured output gets the wrong role.** `role_for_span` in `src/structured.rs` maps `Pagination(_)` to `RegionRole::Artifact`, so form line items are classified as page chrome.
3. **It's public API.** `pub artifact_type` is exposed on `TextSpan`, `TextBlock`, and paths, so consumers read the wrong answer regardless of internal removal behaviour.

Since `mark_running_artifact_spans` runs inside `postprocess_spans`, the mis-tag propagates to everything built on `extract_spans`.

This also matches how this codebase has been resolving the same shape of problem recently — fix the authority, not each consumer: #1010 (`extract_chars`/`extract_spans` now parse through the same parser), #1018 (the table path now reuses the span merger's word-boundary verdict instead of re-deriving it), #1011 (one glyph decoder serves extraction and rendering).

So the two changes are complementary, not alternatives. The upstream change fixes the fact that was wrong; the pass-2 change fixes a genuine separate counting bug. I'm noting this explicitly so the upstream half doesn't get stripped later as redundant — it isn't.

**On verification.** The review's one open item was the size of the sample this had been checked against. I've now run it across our own corpus — 445 PDFs we assembled from open-source documents — comparing text, markdown and HTML output against `main`:

```
docs_compared=445 docs_changed=1 surface_diffs=0 crash_changes=3
```

Zero surface differences. The single flagged document is one large file that sits right on our harness's per-document time limit and flips run to run independently of any PR — noise on our side, not a signal about this change. That closes the verification gap.

**One follow-up, not a blocker:** `ensure_position_consistency` keeps only a signature's largest cluster, which you documented yourself — a legitimate two-slot running head (recto/verso in different corners) would keep one slot and lose the other. Worth a tracking issue; it doesn't affect what this PR claims to fix.

The branch is `BEHIND` but merges cleanly against current main — a resync is all it needs.

ultrasaurus:
@yfedoseev Thanks for the detailed review! The corpus run at that scale is exactly the evidence I was hoping for, and I appreciate you providing clear use cases of why the upstream fix matters.

On the follow-up: this PR is a partial fix of [#794](https://github.com/yfedoseev/pdf_oxide/issues/794), since that issue also reports the recto/verso case. I'd planned to leave that open as a tracking item rather than fold it into this PR — the single-largest-cluster limitation is a real, separate piece of work (per-parity thresholding, not just a bigger cluster cap), and I'd rather layer it on once this lands than grow this PR's scope.

branch is now sync'd to main

yfedoseev:
@ultrasaurus — apologies. The automation got this wrong three separate ways in the last day, and all three are mine, not yours.

**On #1108.** It was closed twice by bots, and both closures were wrong.

The first said it "does not reference an issue". It did — you wrote `Closes https://github.com/yfedoseev/pdf_oxide/issues/1107`, which is GitHub's own supported closing syntax and would have closed #1107 on merge. The check only matches the short `Closes #NNN` form, so it could not see yours. Worse, after you pointed that out, the same bot posted the identical message again, because it re-evaluates on reopen and has no memory of having already been answered. That is a bug and I am fixing the pattern.

The second closure was the one-open-PR rule. That one fired as designed, and it still should not have hit you: I asked you to take #1107 — "please go ahead; the issue is yours" — and then a rule I merged the same week closed the PR you wrote because of it. Inviting work and then auto-rejecting it is not a policy, it is a trap, and you walked into one I set.

**Being straight about what happens next.** Fixing the link-detection pattern on its own would not unblock #1108, because the one-PR rule would still close it while this PR is open. So the honest sequence is this one first.

**And this PR is currently blocked by the same automation.** The gate wants a linked, approved, assigned issue; the description carries the template's empty `Closes #` and #794 has no assignee. That rule landed weeks after you opened this, so it is being applied retroactively to work that could not possibly have complied. Also mine to fix — please do not edit anything on your side for it.

So, concretely:

1. I clear the gate blocking this PR — you do not need to do anything.
2. This PR gets my attention next. I owe you a reply on the open design question from my 19 Aug comment either way; you have waited weeks and that is on me.
3. Once it resolves, reopen #1108 and I will support it through. The bug behind it is real — I traced a second instance of the same root cause in `replace_text_run` while working on something unrelated, so your report found more than the one call site.
4. The link-detection pattern, the duplicate-comment behaviour, and retroactive application to pre-existing PRs all get fixed regardless.

Thank you for pushing back rather than walking away. The rules exist to protect review capacity, and they are worth nothing if they cost the contributors who use the process properly.

yfedoseev:
@ultrasaurus — regression verification is done on this one, and it comes back clean. No issues found; it is in the queue to merge.

Worth saying what it actually took, because the first two runs were worthless and I nearly reported one of them as a pass.

The standard sweep runs conversion with default options, and `include_artifacts` defaults to **true** — so marked artifacts stay in the output and a change to the *marking* is structurally invisible to it. Two full runs came back with zero differences, which said nothing about your PR at all. Re-running with `include_artifacts: false`, so marked spans are actually dropped, is what made your change observable.

With that, the result is a **confirmed improvement**: the only documents that moved both **gained** text, and in both cases what came back is content `main` currently deletes.

- One is a scanned 19th-century magazine where `main` strips a figure out of a sentence, leaving it stating a measurement with no quantity in it. Your change keeps it.
- The other is a journal article where `main` strips the entry numbering out of the reference list — numbered bibliography entries recur in the same band on every page of a reference list, which is exactly the shape the detector mistakes for a folio. The position-consistency work here is what tells them apart.

Zero regressions and zero crash changes elsewhere. Nothing got smaller, which is the right direction for a change whose purpose is to stop over-stripping.

The word counts involved are small, but the kind of damage is not: a missing quantity makes a sentence assert nothing, and missing reference numbers break citation resolution. This is the false-positive class #794 describes.

Remaining before merge is mechanical — a sync onto current `main` and one clean CI cycle. Happy to push the sync for you if that helps; I would merge `main` locally and sign the merge commit rather than use GitHubs **Update branch** button, since that button produces an unsigned merge commit which then fails the DCO check here.

ultrasaurus:
@yfedoseev rebased and ran tests locally (all green), then force pushed.  Should be all ready!

yfedoseev:
Perfect — thank you, and that rebase is exactly right. One commit, your sign-off intact, same seven files. It also cleared the fourteen unsigned merge commits the branch had picked up from repeated `Update branch` clicks, which would have failed DCO; I had a rebase staged to do the same thing and you beat me to it, so I have thrown mine away.

Worth saying plainly: this is a genuinely good change. The discriminator you landed on — that a repeated span is only page furniture if its position is consistent across pages — is the right property to key on, and it is what lets the detector tell a folio apart from a form line-item label or a numbered heading that happens to recur. That distinction is what the corpus run rewarded: the two documents that moved both **gained** text, recovering content `main` currently deletes, with nothing lost anywhere across the rest.

On timing: it should land in the next few hours. There are two of mine ahead of it in the queue (#968 and #1115) that are mid-cycle, and this repo fires roughly 200 checks per push against a single self-hosted runner pool, so they go through one at a time rather than in parallel — several at once just starve each other. Yours is next after those, and nothing about it is waiting on you.

Thanks for sticking with this one through a long review and a fair amount of automation misbehaving at your expense.

PR #982, opened 2026-08-02
fix(extract): stop sideways text fusing across line breaks

### LINKED ISSUE #806: Rotated content-stream text (rotation_degrees ±90) still garbles: reversed reading, glued words, destroyed table/chart structure
## Summary

Text drawn with a **rotated text matrix** (`Tm = [0 1 -1 0 …]` / `[0 -1 1 0 …]`, so each glyph's `rotation_degrees` is `±90`) — vertical chart-axis labels, rotated table headers, and whole landscape tables placed sideways on a portrait page — is **not extracted correctly**. `#804` fixed the catastrophic *cross-column fusion* (whole columns returned as one 1000+ char token), but the underlying rotated-text handling remains broken in three ways:

1. **Reversed reading direction** — a rotated run's glyphs come out backwards.
2. **Lost word boundaries** — adjacent words glue together.
3. **Destroyed 2D structure** — rotated tables/matrices collapse into an unreadable stream with erratic spacing.

This affects both the current release and `#804`'s branch; `#804` only stopped the *worst* fusion. Filing to track the deeper fix.

## Evidence (public arXiv papers, PII-free)

From a 356-PDF regression sweep, **every** doc whose `extract_text` changed between builds had rotated content — and all of them still extract imperfectly. Representative case: **arXiv:2510.24987 page 16** is a full rotated table ("Table 3: Unscaled metric values in experiments").

`extract_chars(16)` — the glyph geometry is *correct* (2318/2320 chars at `rotation_degrees=90`, distinct `origin_y` down each vertical run), but the run text reads **reversed**:

```
x=162  text='stnemirepxe ni seulav cirtem delacsnU :3 elbaT'   # = "Table 3: Unscaled metric values in experiments" REVERSED
x=209  text='noitargetniscimo-owt'                             # = "two-omics integration" reversed + glued
```

`extract_spans(16)` — the span text is de-reversed but **words are glued** and the bbox is **flattened onto the x-axis** (horizontal `w×h` for vertical content):

```
bbox=(319,80,w117,h9)  text='large-scaletwo-omicsintegration'   # 3 words, no spaces; bbox is horizontal
bbox=(179,335,w61,h9)  text='bio-conservation'
```

`extract_text(16)` — the table is unusable (structure gone, spacing erratic):

```
'PCR 0.260 .090.370 .0 00 .0 00.610.810.950 .00 0.0 0 0 .970.980.980.930.670.950.99 ...'
```

Other corpus examples: arXiv:2510.25531 (rotated chart axis — p-value thresholds), arXiv:2510.26274 (rotated algorithm-name labels), arXiv:2510.21165 (rotated network-stats table).

## Minimal synthetic repro (no third-party file needed)

A 1-page PDF whose content stream draws two rotated runs:

```
BT /F1 12 Tf 0 1 -1 0 200 100 Tm (Alpha) Tj ET
BT /F1 12 Tf 0 1 -1 0 214 100 Tm (Bravo) Tj ET
```

`extract_chars` gives `rotation_degrees=90` with correct per-glyph positions, but `extract_words`/`extract_text_lines` mis-handle direction/word-grouping for the rotated runs.

## Root cause

The extractor represents a rotated glyph run with a **flattened, horizontal** span:

- **Span bbox is computed as if horizontal** — `x/y` = run start, `width = Σ glyph advances` laid on the **x-axis**, `height = font`. A vertical run therefore gets a wide-short bbox (`src/extractors/text.rs` — `flush_tj_buffer` / `flush_tj_span_buffer` / `cluster_to_span`). The true axis-aligned box (tall + narrow) is never produced.
- **`TextSpan` carries no 2D per-glyph geometry** — only `bbox`, a snapped `rotation_degrees`, and 1-D advance *magnitudes* (`char_widths`); no text matrix, no per-glyph `(x,y)`, and `char_x_offsets` is empty for rotated runs.
- **`TextSpan::to_chars` re-lays glyphs horizontally** along `bbox.x` at constant `y` (`src/layout/text_block.rs`), discarding the rotation — so every consumer that decomposes spans (`extract_words`, `extract_text_lines`) sees fake-horizontal chars.
- **Clustering + reading-order assume horizontal flow** — `cluster_chars_into_words` / `cluster_words_into_lines` group by y-band then x-gap; `assemble_text_from_spans` (used by `extract_text`, with table detection) reads left-to-right.

The genuine 2-D geometry *does* exist at extraction time — `extract_chars` computes each glyph's real `origin_x/origin_y` and full `matrix` in `show_text` — but it is **lost** the moment a run becomes a `TextSpan`. Reversed reading comes from ordering rotated glyphs by an axis-aligned comparator that doesn't know the run's true direction.

## Relationship to #804

`#804` added a firewall so rotated runs are **not merged/line-grouped across columns** (stops the mega-token fusion). It does **not** reconstruct correct rotated-text reading — words within a rotated run can still be reversed/glued, and rotated tables still collapse.

## Proposed fix direction

Carry the run's real geometry through the span layer instead of flattening it:

1. Give `TextSpan` a true (rotated) bbox and either the text matrix or per-glyph `(x,y)` (or have `to_chars` reconstruct positions along the rotation direction from `origin + advance·(cosθ,sinθ)` with the correct sign from the matrix, not the snapped angle).
2. Make word/line clustering operate in the run's **upright frame** (rotate coords by `-rotation_degrees`, cluster horizontally, rotate back) — mirrors what `order_rotated_blocks` already does for *ordering*.
3. Preserve reading direction from the matrix so runs don't reverse.

This is the "within-span character re-order for rotated multi-glyph spans" follow-up noted in the v0.3.58 `/Rotate` work.

## Impact

Rotated chart axes, rotated table headers, and sideways landscape tables (common in scientific PDFs) extract as reversed/glued/garbled text. Coordinates for rotated words are also wrong (flattened bbox), affecting bbox-based consumers.

tobocop2:
## Linked issue

Closes #806, the run-continuation half. The bounding-box half is #989.

Blocks #987 and #1012. #987's `a_rotate_page_keeps_one_line_per_rotated_run` fails until this lands. Its fixture's two `Tj`s are fused into one span by the continuation test this PR fixes. The carve-out that test pins is therefore never reached.

## What and why

**The bug.** The extractor assembled text drawn under a rotated text matrix as if it ran left-to-right. It judged `Tm` run continuation in unrotated page axes. Consecutive runs of one sideways line then fused across line breaks, giving "the quick brown foxjumps over the lazydog".

**The fix.** ISO 32000-1 §9.4.4 puts the glyph displacement along the text matrix's `(a, b)` row. Under a ±90° matrix a run advances along `f`, while successive lines separate along `e`. This PR judges continuation along the run's own writing axis instead.

**Compatibility.** For any `b = 0, a > 0` matrix the along test equals the existing `e − e0` test. The perpendicular tolerance is ≥ the existing `f`-band: equal when unskewed, looser for sheared matrices. The new test is ANDed onto the old ones, so unrotated output cannot change. Vertical writing (WMode 1) is exempt, since its displacement genuinely runs along `(c, d)`.

| what a reader saw | after |
|---|---|
| "the quick brown foxjumps over the lazydog" on sideways pages | run continuation judged along the writing axis; **392,109** upright pages byte-identical |

This PR was opened before and closed during a consolidation attempt. It is reopened per the one-issue-one-PR rule. The fix is identical, rebased onto current `main`.

## Type of change

- [x] Bug fix

## Tests

- [x] Revert-checked: `rotated_paragraph_keeps_words_and_lines` (the fused-lines scenario) and the quadrant unit test's veto cases fail without the fix. The new conjunct is veto-only, so the remaining tests are pins. The rotated-subscript test pins the perpendicular tolerance against tightening. Its fixture uses a 5° matrix, so the conjunct is actually reached. The rest pin pre-existing boundaries. Reproducers are minimal synthetic PDFs built in code.
- [x] Named by defect class (rotated-run word boundaries, vertical-writing probes)
- Affected feature tiers: none. core text extraction only
- [x] `cargo test --release`: 359 suites, 0 failures. `cargo fmt --check` is clean. Workspace clippy `-D warnings` was clean on the consolidated tree this was sliced from.

## Regression on real PDFs

**Corpus I tested:** 19,151 documents / ~470,000 pages, swept against `main` twice independently with matching results.

- [x] Differential sweep vs `main`, per page, all text surfaces: **392,109 upright pages byte-identical**; only rotated-content pages change
- [x] Judged with word-multiset + structural signals, not char-Levenshtein

**Full-corpus footprint.** The original bundled branch changed 7,718 pages; that measurement covered both this fix and the right-aligned column-stack rule, which review split into #1086/#1087. The reduced branch was re-swept against a fresh base from current `main`: its movement is the rotated-run surface plus pages attributable to #990, which this branch predates, and nothing else. The whitespace, hyphen-join and de-duplication accounting from the bundled measurement (every glyph conserved; 232 hyphen pages; 35 de-duplication pages such as the 19-character spurious line on 018146.pdf page 13) belongs to the rotation rule and still holds.
- [x] Agreement with PyMuPDF, pypdf and pdfminer.six improves on rotated buckets (e.g. rot90 0.7920 → 0.7925 vs PyMuPDF) and is unchanged upright
- [ ] Diffed vs the latest release: not run separately. The sweep ran against `main`, which is this branch's merge base and also the v0.3.77 release commit.

**Diff summary vs `main`:** rotated-content pages regroup as intended; zero upright movement.

| gate | result |
|---|---|
| upright pages byte-identical | 392,109 / 392,109 |
| reference agreement (PyMuPDF / pypdf / pdfminer.six), rotated buckets | improves in every bucket (rot90 vs PyMuPDF 0.7920 → 0.7925), unchanged upright |
| null experiment (same binary twice) | 1-page floor (a font-stats tie, fixed separately) |

|---|---|---|

|---|---|---|
| [008672.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/008.zip) (008.zip) | 4, 18 | base emits bogus concatenations ("PhenobarbitalRifampicinClotrimazole") on top of the correct lines; this PR drops exactly those |
| [027417.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/027.zip) (027.zip) | 3 | base emits all three bullet items twice (once joined, once split); fixed output emits each once |
| [020599.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/020.zip) (020.zip) | 24 | rotated chart axis: "1990199219941996..." becomes "1990 1992 1994 1996 ..." |
| [014644.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/014.zip) (014.zip) | 9 | base line "UNIVERSITYINDUSTRIALINDUSTRIAL" is spurious duplication; fixed output emits "INDUSTRIAL" once |
| [018146.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/018.zip) (018.zip) | 13 | the single flagged "character loss" is exactly a duplicated line removed |

(the last two rows land with the pending glue commit)

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: extensive (implementation, corpus validation, and description drafting)
- [ ] ~~written by me, not generated~~. See disclosure above. Every claim in this description is machine-verified. The evidence is reproducible with the commands in the linked issue thread.

## Checklist

- [x] One logical change
- [x] No public-API change


tobocop2:

tobocop2:
There was a failure related to the Julia package registry that returned a 404 while installing JuliaFormatter, and this PR touches no Julia files. I can open an issue about this if it makes sense, but I just amended the comimt and re-triggered CI this way instead.

yfedoseev:
Thanks for the writeup on the writing-axis fix — that reasoning is solid and it checked out clean against our test corpus, including a golden-reference dataset with known-correct output. The bundled second change (right-aligned column stack) isn't covered by the linked issue though, and its tolerances look tuned to one specific measurement rather than derived generically. Can you split that into its own issue+PR? The rotation fix on its own is close to mergeable.

tobocop2:
Done — the right-aligned column-stack commit is out of this branch. The head is now 9e5b194b: the writing-axis fix, its tests, and the rustfmt cleanup, nothing else. The full library suite and the word-boundary tests pass on the reduced branch.

The column stack gets its own issue and PR. One data point from re-validating the split: the fusion the removed commit targeted ("13,286" + "753" extracting as "13,286753") now reproduces on current main directly — page 15 of govdocs 014094 — since #990 landed, so the follow-up issue will carry a main-based repro rather than a stacked one.

tobocop2:
The column-stack follow-up is up as its own pair: issue #1086 and draft PR #1087, with the right-edge tolerance derived from the run's own glyph advance (half the narrower run's mean advance) instead of a point constant, and a doubled-scale regression test that fails on any point-constant form. Its corpus sweep fills into #1087 today.

tobocop2:
Corpus validation of the reduced branch (9e5b194b): 470,208 pages against a fresh base from today's `main`. The branch's own surface is the expected rotated-run movement; every remaining flagged page attributes to #990, which this branch predates. Nothing new appeared from dropping the column-stack commit.

PR #987, opened 2026-08-02
fix(extract): group rotated runs into whole lines

tobocop2:
> **Blocked by #982.** Merge that one first. This branch is rebuilt on the reduced #982 (the column-stack commit that review split out is no longer in this history), so the diff carries #982's commits and this PR's own on top - review the delta rather than the full file list. The test `a_rotate_page_keeps_one_line_per_rotated_run` depends on #982's `advances_along_writing_axis`; the changes to `src/extractors/text.rs` are #982's, not this PR's. The head also adds the writing-axis ordering fix for merged rotated lines, described in the review thread.

## Linked issue

Closes #983. Blocks #1012.

## What and why

**The bug.** `extract_text_lines` grouped words into lines by banding them on y. A rotated line advances along y, so every word of one visual line landed in its own band. A landscape table page of 458 words came back as 458 one-word lines. `extract_text` read the same page correctly, so two surfaces disagreed about the same document.

**The fix.** The grouper now bands rotated runs along their own writing axis. Runs that share a perpendicular offset become one line, however many `Tm`s drew them. A rotated table row is typically one run per cell.

**Cost.** A `BTreeMap` keyed to 1/100 pt holds those offsets, instead of rescanning them per run. That keeps grouping linear in the run count, on exactly the pages that have the most runs. The quantization saturates, so a crafted PDF with coordinates near the `i64` ceiling degrades to extra lines rather than panicking.

**Carve-out.** A page with a nonzero `/Rotate` keeps one line per rotated run. Its span bboxes are already mapped into the displayed frame, while `rotation_degrees` still describes the pre-display frame. Reading the offset there would use the wrong axis. A test pins this.

No public-API change.

## Type of change

- [x] Bug fix
- [ ] New feature (has an accepted issue: #___)
- [x] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [x] I added a test that **fails before this change and passes after**
      (revert-checked). For bug fixes the reproducer is a **minimal synthetic
      PDF built in code**. No third-party/reporter PDF is committed.
- [x] Test named by defect **class**, not an issue/PR number; no
      contributor/company names in code or fixtures.
- [x] `cargo test` passes, **and** the affected feature tiers:
      `rendering` / `fips,icc` / `ml` / binding (list which): ____
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

[...]

PR #988, opened 2026-08-02
fix(extract): to_markdown and to_html read rotated pages differently from extract_text

### LINKED ISSUE #984: [Bug]: A sideways page reads correctly as text but scrambled as Markdown or HTML
### Bug Description

The same sideways page reads correctly through `extract_text` and scrambled through `to_markdown`, `to_html` and `to_plain_text`. Whichever surface you happen to use decides whether you get sensible text.

On a page whose text is mostly rotated:

`extract_text` gives:

```
Engine oil capacity / six point zero quarts / Motorcraft synthetic blend
```

`to_markdown` on the same page gives the words interleaved across the columns instead, so a line of the table is spliced together from three different rows.

For a conversion or retrieval pipeline the markdown or HTML path is usually the one in production, so in practice the good behaviour is the one nobody is using. `extract_text_filtered` and its region variant have the same problem.

I think one entry point knows how to turn the page and read it upright, and the others assemble in raw page coordinates without that step.

### Steps to Reproduce

Any page whose text is predominantly rotated will do. A minimal one:

```rust
let doc = PdfDocument::from_bytes(rotated_page_pdf())?;
let opts = ConversionOptions::default();
println!("text:     {:?}", doc.extract_text(0)?);
println!("markdown: {:?}", doc.to_markdown(0, &opts)?);
```

The two disagree on reading order. `to_html` and `to_plain_text` agree with markdown, not with `extract_text`.

### Expected Behavior

Every text-producing surface reads a page the same way. If `extract_text` can read a sideways page, so should markdown, HTML and plain text.

### Actual Behavior

Only `extract_text` reads it correctly. The converters and the filtered paths return scrambled reading order for the same page.

### Environment

- **Rust version**: stable

### Checklist

**AI disclosure:** drafted with AI assistance and reviewed before filing. The behaviour was reproduced, not inferred, so I'm leaving the box above unticked rather than claiming otherwise.

tobocop2:
## Linked issue

Closes #984.

## What and why

The rotated reading frame was applied at one call site: `extract_text` entered it, while `to_markdown`, `to_html`, `to_plain_text`, `extract_text_filtered` and `extract_text_in_rect` assembled in raw page space — six surfaces disagreeing about the same page. Every surface now reads a rotated page in the same frame.

**Merge order:** builds on #982 and #987 (its converter-parity tests assert agreement with `extract_text`, which requires those fixes to read rotated pages correctly); this branch includes them and rebases to a pure delta as they land.

## Evidence

| what a reader saw | after |
|---|---|
| `extract_text` read a rotated page correctly, `to_markdown` did not | **1,620** rotated pages read identically across text, markdown, HTML and plain text |

**Corpus:** 14,911 documents / 454,905 pages per page. Harness: https://gist.github.com/tobocop2/e053b98e486eee76185d01dafe628f2f

| gate | result |
|---|---|
| upright pages, markdown and HTML | byte-identical across all **392,109** — the signal that word geometry did not move |
| converter parity on rotated pages | text/markdown/HTML/plain read the same page the same way |
| this branch (with its #982+#987 base), `cargo test --release` | full-suite revalidation running; parity tests green on the consolidated tree this was sliced from |

## Type of change

- [x] Bug fix

## AI assistance disclosure

- [x] AI assistance: **assisted** — tool: Claude Code, extent: extensive
- [ ] ~~written by me, not generated~~ — see disclosure

## Checklist

- [x] One logical change
- [x] No public-API change

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -5393,13 +5393,27 @@ impl PdfDocument {
         // portrait page): the row-major assembler groups lines
         // in the portrait frame and interleaves every rotated row. Assemble
         // such pages in their rotated reading frame instead.
-        let base_spans = match self.map_dominant_rotation_into_reading_frame(page_index, base_spans)
-        {
-            Ok(mapped) => mapped,
-            Err(original) => original,
-        };
+        let base_spans = self.spans_in_reading_frame(page_index, base_spans);
         let text = self.assemble_text_from_spans(page_index, base_spans, options)?;
         Ok(Self::apply_mixed_rtl_line_pass(text))
+    }
+
+    /// [`Self::map_dominant_rotation_into_reading_frame`] with the
+    /// mapped/unchanged distinction collapsed, for the call sites that only
+    /// want "spans as a reader sees them".
+    ///
+    /// Every text-producing surface goes through this. Applying the frame at
+    /// one call site made the same page read correctly through `extract_text`
+    /// and incorrectly through `to_markdown` / `to_html` / `to_plain_text`.
+    fn spans_in_reading_frame(
+        &self,
+        page_index: usize,
+        spans: Vec<crate::layout::TextSpan>,
+    ) -> Vec<crate::layout::TextSpan> {
+        match self.map_dominant_rotation_into_reading_frame(page_index, spans) {
+            Ok(mapped) => mapped,
+            Err(original) => original,
+        }
     }
 
     /// Map a dominant-rotation page's spans into their rotated reading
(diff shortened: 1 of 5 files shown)


tobocop2:
Consolidated into #993.

The reading-frame work is one defect across several surfaces — merging any subset leaves `extract_text` and the converters disagreeing about the same page — so it is easier to review as a single change with one commit per issue. This PR's reproducer, test and corpus gate are carried over there, and its description is inlined in #993 under *Per-issue detail*.

#993 also carries three determinism fixes found while measuring this work.

PR #992, opened 2026-08-02
fix(rendering): warn when a glyph is dropped instead of failing silently

### LINKED ISSUE #991: [Bug]: Characters vanish from rendered pages with no warning
### Bug Description

Characters vanish from rendered pages and nothing says so. The cursor still moves, so the glyph leaves a gap that looks like a space, and the page comes back looking plausible while being wrong.

I  hit this in [lilbee](https://github.com/tobocop2/lilbee) with a font where every `i`, `j`, `.` and `:` disappeared from the rendered output. Nothing in the logs, no fallback character, no error. OCR read the gap-filled images faithfully and the corrupted text reached a search index before anyone noticed.

What makes it hard to catch is that a blank region is indistinguishable from real whitespace. There is no way for a caller to tell that the page it just rendered is missing characters.

There are a few different ways a glyph ends up dropped: the font gives back no outline for it, the character has no mapping in the font's encoding and never gets painted (not even as `.notdef`), or the character fails to decode and leaves the shaping input entirely, which also shifts everything after it.

### Steps to Reproduce

Render any page using a font that fails to produce an outline for some glyph. The case we hit was a font using the deprecated CFF `dotsection` operator, which makes the outline parser abort the charstring and return nothing.

Compare the rendered image against the extracted text: the text has the characters, the image does not, and nothing is logged at warn level.

### Expected Behavior

When a glyph produces nothing but the cursor still advances, say so. A warning naming the font and the character, once per font rather than once per glyph, would have caught this immediately.

Painting `.notdef` for an unmapped character would also match what most viewers do, though the logging is the part that matters.

### Actual Behavior

The glyph is silently skipped, the cursor advances, and the page renders with an invisible gap. Nothing is reported at warn level, so a caller has no signal that anything went wrong.

### Environment

- **Rust version**: stable

### Checklist

**AI disclosure:** drafted with AI assistance and reviewed before filing, based on a failure we hit in production. Leaving the box above unticked rather than claiming otherwise.

tobocop2:
## Linked issue

Closes #991.

## What and why

A glyph that failed to paint vanished silently while the cursor advanced. Four rasterizer sites swallowed the drop: a `None` outline from ttf-parser painted nothing; an unmapped non-whitespace char never painted, not even `.notdef`; chars resolving to U+FFFD left the shaping input (no glyph, no advance, index skew); and total loss was reported at `debug` only. All four now report. Drops are counted per text run, so a broken font is named once rather than once per glyph, and one document's warning cannot suppress the next document's. Includes the follow-up that makes the ignored reporting test call the real render API.

## Evidence

| what a reader saw | after |
|---|---|
| glyphs missing from output with nothing in the logs | every drop counted and attributed; **raster output byte-identical** — logging only |

**Corpus:** 14,911 documents / 454,905 pages. Harness: https://gist.github.com/tobocop2/e053b98e486eee76185d01dafe628f2f

| gate | result |
|---|---|
| upright pages changed | **0** — rendering behaviour unchanged, reporting only |
| this branch alone, `cargo test --release` | full-suite revalidation running; component suites green standalone on the tree this was sliced from |

## Type of change

- [x] Bug fix (observability; no rendering behaviour change)

## AI assistance disclosure

- [x] AI assistance: **assisted** — tool: Claude Code, extent: extensive
- [ ] ~~written by me, not generated~~ — see disclosure

## Checklist

- [x] One logical change
- [x] No public-API change

### DIFF
--- a/src/extractors/text.rs
+++ b/src/extractors/text.rs
@@ -5564,6 +5564,11 @@ impl<'doc> TextExtractor<'doc> {
                 // scale-relative (0.5× the text-space glyph height, ≥0.5pt
                 // floor) so it is correct at any font size and still
                 // splits genuine line breaks.
+                //
+                // The `f`/`e` tests below only mean "same line" and "forward"
+                // while the run advances along +x; under a rotated matrix the
+                // two axes swap. The added conjunct re-checks both along the
+                // run's own writing axis (ISO 32000-1 §9.4.4).
                 let cur_font_size = self.state_stack.current().font_size;
                 let is_continuation = self.merging_config.merge_tm_tj_runs
                     && match self.tj_span_buffer {
@@ -5576,7 +5581,14 @@ impl<'doc> TextExtractor<'doc> {
                                 && b == buffer.start_matrix.b
                                 && c == buffer.start_matrix.c
                                 && d == buffer.start_matrix.d
-                                && e >= buffer.start_matrix.e =>
+                                && e >= buffer.start_matrix.e
+                                && Self::advances_along_writing_axis(
+                                    buffer.start_matrix,
+                                    buffer.wmode,
+                                    e,
+                                    f,
+                                    cur_font_size,
+                                ) =>
                         {
                             // Same line, same transform, LTR progression →
                             // update width to reflect actual visual extent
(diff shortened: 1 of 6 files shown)


tobocop2:
Consolidated into #993.

The reading-frame work is one defect across several surfaces — merging any subset leaves `extract_text` and the converters disagreeing about the same page — so it is easier to review as a single change with one commit per issue. This PR's reproducer, test and corpus gate are carried over there, and its description is inlined in #993 under *Per-issue detail*.

#993 also carries three determinism fixes found while measuring this work.

PR #990, opened 2026-08-02
fix(extract): keep words that the table grid does not re-emit

tobocop2:
## Linked issue

Closes #981.

## What and why

Turning table extraction on deleted words the grid never re-emitted. Three ownership checks overshot.

**Multiplicity.** Cell text was tracked as a `HashSet`. That records *that* a text is a cell, but never *how many* cells hold it. Repeated values lost their surplus. `")("` was emitted 7 times instead of 28.

**Containment.** A value column without row-separator rules became one tall empty-text cell. Its bbox swallowed every figure in the column. It re-emitted none of them.

**Double claims.** The containment path and the text-fallback path kept separate accounts. One cell could therefore be claimed twice. Three identical labels dropped against two rendering cells.

`render_text()` emits each cell's text exactly once. Ownership is now a per-cell token budget, drawn on by both paths. The budget is denominated in the cell's member spans. Filler cells with no members use rendered text instead, because builders glue sub-em gaps and split column-spanning decimals.

**Three rules follow.** A covered span leaves the flow only by consuming its tokens from a covering cell. A same-text span outside every cell claims a cell only while that cell's budget is whole. Spans no cell can absorb stay in the flow: duplication over silent deletion.

The new drop set is a strict subset of the old one by construction. No page can emit less content than before. This also enforces the documented unflushed-table invariant on the MCID branch, which previously held only on the bbox path.

**Rejected alternatives**, both measured on the full corpus. A multiplicity budget without ownership scoping fixed 4,243 pages but regressed 48, by resurrecting fragments of merged cells. Substring-containment ownership reintroduced deletion through rowspan-merged cells on 660 pages.

## Type of change

- [x] Bug fix
- [ ] New feature (has an accepted issue: #___)
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [x] I added a test that **fails before this change and passes after**
      (revert-checked). For bug fixes the reproducer is a **minimal synthetic
      PDF built in code**. No third-party/reporter PDF is committed.
- [x] Test named by defect **class**, not an issue/PR number; no
      contributor/company names in code or fixtures.
- [ ] `cargo test` passes, **and** the affected feature tiers:
      `rendering` / `fips,icc` / `ml` / binding (list which): ____
- [ ] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Regression on real PDFs

[...]

PR #989, opened 2026-08-02
fix(extract): report the correct bounding box for sideways runs

tobocop2:
> [!IMPORTANT]
> **Merge order: merge #1063 before this PR.** #996 closed in favour of its splits; the table-cells split #1063 carries the `table_extractor.rs` changes this note used to point at. Once #1063 is in, this PR needs a small rebase there, about 14 lines, easy to reapply. #1057 stacks on this PR and merges after it.

## Linked issue

Part of #806. This is the bounding-box half. The run-continuation half is #982. The issue stays open until both land.

## What and why

A run drawn sideways reported a wide, short box for text that is tall and narrow.

**Why not redefine `bbox`.** The run-local extents are load-bearing. Two internal consumers depend on them, and `TextSpan` is serialized to JS. Redefining `bbox` would move every word box on every rotated page in every binding.

**What this does instead.** `page_bbox()` is a derived accessor. It rotates the extents about the run origin into page space, mirroring `PathContent::rendered_bbox()`. At rotation 0 it returns `bbox` unchanged. That is why upright pages cannot move.

**Pages with a nonzero `/Rotate`.** Span postprocessing has already rewritten `bbox` into the displayed frame, while `rotation_degrees` still holds the raw content angle. Rotating again would double-transform exactly the sideways-landscape-in-a-portrait-MediaBox case the issue documents. Two runtime-only (`serde(skip)`) fields prevent that structurally: the mapper records `page_rotation_applied`, and `page_bbox()` builds the hull in the frame `bbox` actually lives in. Mirrored runs (negative determinant) reflect across the axis instead of rotating.

**Bindings.** The C export and the Python property moved to their own pair per review: issue #1056 and PR #1057, which stacks on this one. This PR now carries only the core fix.

The PR also carries the rustdoc intra-doc-link fix for the accessor. Doc builds run `-D warnings` on stable.

## Evidence

| what a reader saw | after |
|---|---|
| sideways run: wide, short box | `page_bbox()` returns the page-space rectangle; `bbox` keeps its meaning; **0 pages move** |

| gate | result |
|---|---|
| pages changed by this commit | **0**. Accessor is derived, no stored geometry changes |
| this branch alone, `cargo test --release` | 359 suites, 0 failures; fmt clean; rustdoc `-D warnings` clean |
| API surface | additive only (`page_bbox()`); semver-checks green |

## Type of change

- [x] New feature (accepted issue: #806). A derived accessor, with no behavior change to existing fields

## Tests

[...]

PR #993, opened 2026-08-03
fix: read rotated pages the same way on every text surface

### LINKED ISSUE #806: Rotated content-stream text (rotation_degrees ±90) still garbles: reversed reading, glued words, destroyed table/chart structure

tobocop2:
## Linked issues

Closes #806, #981, #983, #984, #991.

Supersedes #982, #987, #988, #989, #992 — each was opened per-issue and carries that
issue's own reproducer and corpus gate. They are consolidated here because splitting them
left the crate internally inconsistent: `extract_text` would read a rotated page correctly
while `to_markdown`, `to_html` and `to_plain_text` read it wrongly.

## What this fixes

Every row is measured on the same 454,905-page corpus, per page rather than per document.

| issue | what a reader saw | after |
|---|---|---|
| **#806** | a run drawn sideways reported a wide, short box for text that is tall and narrow | `page_bbox()` returns the page-space rectangle; `bbox` keeps its meaning and **0 pages move** |
| **#983** | a landscape table came back as one line per cell — 458 words, 458 lines | **6,097** rotated-content pages regrouped; **392,109** upright pages byte-identical |
| **#984** | `extract_text` read a rotated page correctly and `to_markdown` did not | **1,620** rotated pages now read identically across text, markdown, HTML and plain text |
| **#991** | a glyph that failed to paint vanished silently while the cursor advanced | four rasterizer sites now report the drop; **0** upright pages change, rendering behaviour unchanged |
| **#981** | turning table extraction on deleted words the grid never re-emitted | losing pages **7,657 → 3,916** and lost token instances **78,019 → 17,995**; on every one of 454,930 pages the fix arm emits at least the characters the base arm does |
| — | `extract_text` returned different text for the same input, run to run | a font-size tie was a per-process coin flip: **6 of 12** runs disagreed, now **16 of 16** identical |
| — | a redacted PDF's bytes differed between runs | dictionary emission sorted; `ObjectSerializer` already did this, redaction was a second copy that missed it |
| — | table rule families ordered by hash iteration | ordered by geometry; latent, no corpus page reached it |
| — | `extract_chars` read a sideways chart axis flat while `extract_spans` read it rotated | both surfaces share one parser; 1,134 of 454,899 pages change, and agreement with PyMuPDF, pypdf and pdfminer improves in every rotation bucket |
| — | on pages over 256 KB, text relying on spacing, leading or colour set earlier on the page drifted, piled onto one line, or turned black | the fast-path parser now carries the full persistent graphics state into each text region |
| — | a Type0 font with a UTF-8 CMap extracted right and rendered garbage; a ligature rendered right but extracted lossy | one glyph decoder for both paths, carrying the union of the two feature sets |

## Problem

Text drawn under a rotated text matrix was assembled as if it ran left-to-right. Six
surfaces disagreed about the same page, and three unrelated defects surfaced while
measuring: glyphs dropped without a log line, output that was not reproducible between
runs of the same binary, and redacted PDFs that were not byte-reproducible.

```
before:  "the quick brown foxjumps over the lazydog and keeps going"
after:   "the quick brown fox" / "jumps over the lazy" / "dog and keeps going"
```

## Fixes

One commit each.

| # | commit | what it fixes |
|---|---|---|
| #982 | `fix(extract): sideways text joins consecutive lines together` | `Tm` run continuation judged along the run's own writing axis |
| #983 | `fix(extract): group a rotated line drawn as several runs into one line` | `extract_text_lines` returned one line per run — 458 words became 458 lines |
| #984 | `fix(convert): read a rotated page in the same frame on every text surface` | the rotated frame was applied in `extract_text` only |
| #806 | `feat(layout): expose where a rotated run sits on the page` | a sideways run's page-space rectangle, as a derived accessor |
| #991 | `fix(render): report glyphs that paint nothing` | four rasterizer sites swallowed a failed glyph while advancing the cursor |
| — | `fix(layout): break modal-statistic ties deterministically` | see below |
| — | `fix(redaction,compliance): emit hash-backed dictionaries in sorted order` | see below |
| — | `fix(tables): order rule families deterministically` | see below |
| — | `fix(content): reconstruct inherited graphics state in prescan text regions` | leading, spacing and fill colour were silently reset at synthesized region boundaries on pages over 256 KB; a follow-up hardens the colour replay — a stale name operand no longer fabricates a pattern (`/GS0 gs 0.5 scn`), and a colour with more components than the scanner can faithfully replay injects nothing rather than a truncated wrong value. Isolated against the corpus: 89 of 455,053 pages change, char-level colour only, every one in the over-256 KB prescan path, extracted text byte-identical everywhere |
| — | `fix(extract): parse chars and spans with the same parser` | `extract_chars` drove a second parser that rebuilt graphics state by a different route |
| — | `refactor(fonts): unify the duplicated glyph decoder` | see below |
| #981 | `fix(extract): stop table extraction deleting words the grid does not re-emit` | ownership of a span becomes a per-cell token budget instead of set membership and bbox containment |

### Why the reading frame is one change, not four

ISO 32000-1 §9.4.4 puts the glyph displacement along the text matrix's `(a, b)` row, so
under a ±90° matrix a run advances along `f` while successive **lines** separate along
`e` — the two axes are swapped. Every surface that assembled text made the same
assumption in a different place. Fixing one leaves the others wrong on the same document.

**Unrotated text cannot move.** For any `b = 0, a > 0` matrix the projection is termwise
the existing test: `along` **is** `e − e0`, `perp` **is** `f − f0`. Vertical writing
(WMode 1) is exempt — the displacement genuinely runs along `(c, d)` there.

### The determinism fixes

Found while measuring, not sought. `max_by_key` returns the **last** maximal element, and
over a `HashMap` that is the per-process-randomized iteration order:

- **`PageFontStats`** — two font-size buckets tying made `dominant_em` a coin flip.
  It feeds the multi-column gate that selects the reading-order branch, so the whole
  page's text reordered. Measured on `govdocs_046_046432.pdf` p3: the histogram is
  `8pt:32, 41pt:32`, an exact tie, and the page's extracted text alternated between two
  values across runs of the same binary — 6 of 12 runs each. With the tie-break, 16 of 16
  identical.
- **`redaction::serialize`** — `Object::Dictionary` is a `HashMap`, so a redacted PDF's
  bytes differed between runs. Unconditional, not tie-dependent.
  `writer::ObjectSerializer` already sorts its keys for this reason; this was a second
  copy that missed it.
- **PDF/X and font-embedding validators**, **table rule families** — same class.

Ties resolve to the smaller value and the lexicographically smaller name. For
`dominant_em` that is the conservative direction: it is meant to be the *body* em, and a
tighter cluster gap makes the multi-column gate stricter rather than looser.

### Corpus

| property | value |
|---|---|
| documents | 14,911 |
| **pages compared** | **454,905** |
| sources | govdocs1 `.gov` crawl (60 slices), arXiv, PubMed Central OA, IRS forms, pdf.js corpus, PoC‖GTFO |
| named fixtures | vehicle owner's manual, 4 rotated arXiv papers, 2 vertical-Japanese novels, FAA Airplane Flying Handbook |
| machine | 32 vCPU, sharded 32 ways |

Compared **per page**: a single garbled page in a 200-page report barely moves a
whole-document score.

### Noise floor

The same binary swept twice over the whole corpus, before the determinism fixes:
**1 page** differed. That page is `govdocs_046_046432.pdf` p3 — the font-stats tie. Every
gate below is read against that floor, so a one-page difference is not evidence of
anything until it is reproduced.

> The corpus figures below were measured at `c5be50ff`. One later commit
> (`perf(extract): index rotated line offsets…`) is behaviour-preserving by
> construction but has not been re-swept yet, so these numbers describe the tree
> one commit back.

### Differential gate — every page, both arms

454,899 pages compared (8,832 excluded as empty in both arms). Signals compared per page:
extracted text, words+bboxes, spans+bboxes, lines, whitespace-free text, markdown, HTML,
table counts and `extract_chars`.

| bucket | pages | byte-identical | changed |
|---|---|---|---|
| **upright** | **392,109** | **392,109** | **0** |
| upright + rotated minority | 17,831 | 11,701 | 6,130 |
| rot90 | 34,355 | 30,241 | 4,114 |
| rot270 | 1,704 | 1,623 | 81 |
| skew | 42 | 31 | 11 |
| mixed | 25 | 15 | 10 |
| rot180 | 1 | 0 | 1 |

**Blockers: 0.** Not one upright page moves. Markdown and HTML are measured because table
detection consumes word geometry — a change confined to the assembler moves `text` alone,
one that perturbed geometry moves those too.

### Agreement with reference extractors

A token-multiset diff cannot distinguish "text got worse" from "text got reorganized", and
this change reorganizes rotated pages by design. Three independent extractors can. Both
arms scored on how closely they agree with each, 6,000 sampled pages.

| bucket | pages | reference | base | fix | |
|---|---|---|---|---|---|
| rot90 | 2,859 | PyMuPDF | 0.7920 | **0.7925** | improved |
| rot90 | 2,859 | pypdf | 0.7865 | **0.7871** | improved |
| rot90 | 2,859 | pdfminer.six | 0.7387 | **0.7392** | improved |
| upright | 2,869 | PyMuPDF | 0.9449 | 0.9449 | unchanged |
| upright | 2,869 | pypdf | 0.9244 | 0.9244 | unchanged |
| upright | 2,869 | pdfminer.six | 0.9389 | 0.9389 | unchanged |
| rot270 | 140 | all three | — | — | unchanged |
| upright + minority | 130 | all three | — | — | unchanged |

The fix moves toward all three references simultaneously on rotated pages and away from
none anywhere. The sample spans all rotated pages rather than only changed ones, so
unchanged pages dilute these means — the per-changed-page effect is larger.

### Unit tests

`cargo test` is green on the consolidated branch. Each fix carries its own tests, and the
tie-breaks are tested in pairs — one test pinning the tie resolution, one asserting a
unique winner is unaffected, because the failure mode of a tie-break is changing the
ordinary no-tie case.

### Harness trust

| check | result |
|---|---|
| null experiment (same binary twice) | 1 page — the flap, now fixed |
| determinism after the fix | 16/16 identical on the flapping page |
| arms differ | yes — a null result elsewhere is earned, not an artifact |
| shard completion | 32/32 reconciled against the corpus count |
| corpus manifest verified per machine | 14,911 files, sha256 |

### #981 — table extraction deleting words the grid does not re-emit

Three deletion mechanisms shared one root: ownership tests that overshoot. A cell-text
`HashSet` records *that* a text is a cell but never *how many* cells hold it, so repeated
values lost their surplus (`")(" 28→7`). Bbox containment is not ownership: a value column
without row rules becomes one tall empty-text cell whose box swallows every figure in the
column and re-emits none of them, and a rowspan-merged cell's box covers far more spans
than its text was built from. And with the containment and text-fallback paths keeping
separate accounts, one cell could be claimed twice — three identical labels dropped
against two rendering cells.

`render_text()` emits each cell's text exactly once, so ownership is now a per-cell token
budget drawn on by both paths. A covered span leaves the flow only by consuming its tokens
from a covering cell; a same-text span outside every cell claims a cell only while its
budget is whole; spans no cell can absorb stay in the flow — better to duplicate than to
silently drop. The drop set is a strict subset of the old one by construction.

| check | result |
|---|---|
| losing pages, table extraction on vs off | 7,657 → 3,916 of 454,905 |
| lost token instances | 78,019 → 17,995 |
| per-page character multisets, fix arm vs base arm, all 454,930 pages | fix ⊇ base everywhere; 27 pages differ only by hyphens consumed joining wrapped words |
| pages newly flagged by the token metric | 21, every one losing **zero** characters — kept spans shift token boundaries |
| reference agreement on a 1,500-page stride sample of the 7,329 changed pages | Jaccard 0.8425 → **0.8467**, recall 0.9160 → **0.9357** against PyMuPDF, pypdf and pdfminer |
| instrument determinism | two independently built binaries produced byte-identical loss reports across the corpus |

The regression test builds the failing geometry synthetically — header band ruled across
every column, row rules stopping short of the value column, a figure inside the resulting
tall empty cell — and fails on the old drop pass by deleting the figure. A second fixture
pins the opposite failure: an ownership rule that exempts unmatched spans resurrects
fragments of merged cells and duplicates them (the earlier counting attempt regressed 48
pages exactly that way; its per-page reference scores also fall, which is what killed it).

## Per-issue detail

Inlined from the superseded PRs.

### #983 — `extract_text_lines` bands words by y

A rotated run advances along y, so every word of one visual line lands in its own band. A
page of 458 words came back as 458 lines while `extract_text` read it correctly, so the
two surfaces disagreed about the same page.

### #984 — the frame was applied at one call site

| surface | before |
|---|---|
| `extract_text` | entered the rotated frame |
| `to_markdown`, `to_html`, `to_plain_text` | assembled in raw page space |
| `extract_text_filtered`, `..._in_rect` | assembled in raw page space |

Markdown and HTML byte-identity across all 392,109 upright pages is the signal that word
geometry did not move.

### #806 — why an accessor, not a redefinition

The run-local encoding is load-bearing: `document.rs:5452` and `page_order.rs:164` both
depend on the extents being run-local, and `TextSpan` is serialized to JS in `wasm.rs`, so
the field meaning is a public contract. Redefining `bbox` would move every word box on
every rotated page and change that contract in all bindings. `page_bbox()` is derived,
mirrors `Path::rendered_bbox()`, and returns `bbox` unchanged at rotation 0 — which is why
**0 pages move**.

### #991 — four sites that swallowed a glyph

| site | what was swallowed |
|---|---|
| `render_cid_direct` (`text_rasterizer.rs:1537`) | a `None` outline from ttf-parser paints nothing |
| encoding lookup → GID 0 (`:1488-1500`), `gid != 0 \|\| whitespace` gate (`:1533`) | an unmapped non-whitespace char never painted, not even `.notdef` |
| `decode_text_to_unicode` (`:653-668`) | chars resolving to U+FFFD leave the shaping input — no glyph, no advance, index skew |
| `render_unicode_text` total loss (`:1395-1401`) | reported at `debug` only |

Logging only; raster output is byte-identical. Drops are counted per text run, so a broken
font is named once rather than once per glyph, and one document's warning cannot suppress
the next document's.

## Consolidated from

Each of these was opened per-issue with its own reproducer and its own corpus gate. They
are closed in favour of this one because the reading-frame fixes are a single defect
across several surfaces: merging any subset leaves `extract_text` and the converters
disagreeing about the same page.

| PR | issue | what it carried | where it is now |
|---|---|---|---|
| #982 | #806 | `Tm` run continuation judged along the run's own writing axis | commit 1 |
| #987 | #983 | rotated line grouping — 458 words no longer become 458 lines | commit 2 |
| #988 | #984 | the rotated frame applied on every text surface | commit 3 |
| #989 | #806 | `page_bbox()` — where a sideways run sits on the page | commit 4 |
| #992 | #991 | four rasterizer sites that dropped glyphs silently | commit 5 |
| #990 | #981 | table extraction deleting words | closed — superseded by the per-cell budget commit here |

The three determinism commits (6–8) have no predecessor PR: they were found while
measuring this work, not filed beforehand.

<details>
<summary><b>Validation harness (click to expand)</b></summary>

**`rot_sweep.rs`**

```rust
//! Per-page extraction signatures for differential regression testing.
//!
//! Emits one TSV row per page so a single garbled page in a long document is
//! visible; whole-document scores hide exactly that. Each page carries a
//! rotation bucket so upright and rotated populations can be gated differently.
//!
//! ```text
//! rot_sweep <corpus_dir> [--text-out <dir>] [--shard <i>/<n>]
//! ```

use std::collections::hash_map::DefaultHasher;
use std::fmt;
use std::hash::{Hash, Hasher};
use std::path::{Path, PathBuf};

use pdf_oxide::converters::ConversionOptions;
use pdf_oxide::document::PdfDocument;
use pdf_oxide::layout::{TextChar, TextSpan, Word};

const HEADER: &str = "relpath\tpage\tbucket\tnchars\tchars_hash\trot_frac\ttext_hash\twords_hash\tspans_hash\tlines_hash\tnwords\tmaxword\tnlines\tnspans\tntables\twsfree_hash\tmd_hash\thtml_hash\tntable_tags";

/// Glyph-rotation classification of a page.
#[derive(Clone, Copy, PartialEq, Eq)]
enum Bucket {
    Empty,
    Upright,
    UprightMinorityRotated,
    Rot90,
    Rot180,
    Rot270,
    Mixed,
    Skew,
}

impl fmt::Display for Bucket {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(match self {
            Self::Empty => "empty",
            Self::Upright => "upright",
            Self::UprightMinorityRotated => "upright_minority_rot",
            Self::Rot90 => "rot90",
            Self::Rot180 => "rot180",
            Self::Rot270 => "rot270",
            Self::Mixed => "mixed",
            Self::Skew => "skew",
        })
    }
}

/// Command-line inputs.
struct Args {
    corpus: PathBuf,
    text_out: Option<PathBuf>,
    shard_index: usize,
    shard_count: usize,
}

impl Args {
    fn parse() -> Self {
        let mut argv = std::env::args().skip(1);
        let corpus = argv
            .next()
            .map(PathBuf::from)
            .expect("usage: rot_sweep <corpus_dir> [--text-out <dir>] [--shard i/n]");
        let mut text_out = None;
        let mut shard = (0usize, 1usize);
        while let Some(flag) = argv.next() {
            match flag.as_str() {
                "--text-out" => text_out = argv.next().map(PathBuf::from),
                "--shard" => shard = parse_shard(&argv.next().unwrap_or_default()),
                _ => {},
            }
        }
        Self {
            corpus,
            text_out,
            shard_index: shard.0,
            shard_count: shard.1,
        }
    }

    /// Indices this shard owns, striped so every shard sees a similar file mix.
    fn owns(&self, index: usize) -> bool {
        index % self.shard_count == self.shard_index
    }
}

/// Parse an `i/n` shard selector, defaulting to the whole corpus.
fn parse_shard(spec: &str) -> (usize, usize) {
    let (index, count) = spec.split_once('/').unwrap_or(("0", "1"));
    let count = count.parse().unwrap_or(1).max(1);
    (index.parse().unwrap_or(0).min(count - 1), count)
}

/// Every PDF under `dir`, in a stable order so shards are reproducible.
fn collect_pdfs(dir: &Path, out: &mut Vec<PathBuf>) {
    let Ok(entries) = std::fs::read_dir(dir) else {
        return;
    };
    let mut paths: Vec<PathBuf> = entries.flatten().map(|entry| entry.path()).collect();
    paths.sort();
    for path in paths {
        if path.is_dir() {
            collect_pdfs(&path, out);
        } else if path.extension().and_then(|e| e.to_str()) == Some("pdf") {
            out.push(path);
        }
    }
}

fn hash_of(text: &str) -> u64 {
    let mut hasher = DefaultHasher::new();
    text.hash(&mut hasher);
    hasher.finish()
}

/// Quadrant a glyph rotation snaps to, or `None` for a free angle.
fn quadrant(degrees: f32) -> Option<u16> {
    if !degrees.is_finite() {
        return None;
    }
    let normalized = degrees.rem_euclid(360.0);
    [0u16, 90, 180, 270].into_iter().find(|&q| {
        let delta = (normalized - f32::from(q)).abs();
        delta < 0.5 || delta > 359.5
    })
}

/// Glyphs as the text assembler sees them.
///
/// Deliberately NOT `extract_chars`: on some pages that API returns a strict
/// subset of the glyphs `extract_spans` returns, and the rotated ones are
/// exactly what it omits. Bucketing from it labelled genuinely rotated pages
/// `upright` and held them to the byte-identity gate — measuring the wrong
/// population. Buckets must come from the glyphs the code under test consumes.
fn assembler_glyphs(spans: &[TextSpan]) -> Vec<TextChar> {
    spans.iter().flat_map(TextSpan::to_chars).collect()
}

/// Classify a page from its glyph rotations.
fn classify(chars: &[TextChar]) -> Bucket {
    if chars.is_empty() {
        return Bucket::Empty;
    }
    let mut quadrant_counts = [0usize; 4];
    let mut skewed = 0usize;
    for c in chars {
        match quadrant(c.rotation_degrees) {
            Some(0) => quadrant_counts[0] += 1,
            Some(90) => quadrant_counts[1] += 1,
            Some(180) => quadrant_counts[2] += 1,
            Some(270) => quadrant_counts[3] += 1,
            _ => skewed += 1,
        }
    }
    if skewed * 2 > chars.len() {
        return Bucket::Skew;
    }
    if quadrant_counts[1..].iter().sum::<usize>() + skewed == 0 {
        return Bucket::Upright;
    }
    let (index, &dominant) = quadrant_counts
        .iter()
        .enumerate()
        .max_by_key(|(_, &count)| count)
        .expect("quadrant_counts is non-empty");
    if dominant * 2 <= chars.len() {
        return Bucket::Mixed;
    }
    match index {
        0 => Bucket::UprightMinorityRotated,
        1 => Bucket::Rot90,
        2 => Bucket::Rot180,
        _ => Bucket::Rot270,
    }
}

/// Fraction of glyphs that are not upright.
fn rotated_fraction(chars: &[TextChar]) -> f64 {
    if chars.is_empty() {
        return 0.0;
    }
    let upright = chars
        .iter()
        .filter(|c| quadrant(c.rotation_degrees) == Some(0))
        .count();
    (chars.len() - upright) as f64 / chars.len() as f64
}

fn words_signature(words: &[Word]) -> String {
    let parts: Vec<String> = words
        .iter()
        .map(|w| {
            format!(
                "{}|{:.2},{:.2},{:.2},{:.2}",
                w.text, w.bbox.x, w.bbox.y, w.bbox.width, w.bbox.height
            )
        })
        .collect();
    parts.join("\u{1}")
}

fn chars_signature(chars: &[TextChar]) -> String {
    let parts: Vec<String> = chars
        .iter()
        .map(|c| {
            format!(
                "{}|{:.2},{:.2}|{:.1}",
                c.char, c.bbox.x, c.bbox.y, c.rotation_degrees
            )
        })
        .collect();
    parts.join("\u{1}")
}

fn spans_signature(spans: &[TextSpan]) -> String {
    let parts: Vec<String> = spans
        .iter()
        .map(|s| {
            format!(
                "{}|{:.2},{:.2},{:.2},{:.2}|{:.1}",
                s.text, s.bbox.x, s.bbox.y, s.bbox.width, s.bbox.height, s.rotation_degrees
            )
        })
        .collect();
    parts.join("\u{1}")
}

/// One page's measured signature.
struct PageRow {
    bucket: Bucket,
    chars: usize,
    chars_hash: u64,
    rotated_fraction: f64,
    text: u64,
    words: u64,
    spans: u64,
    lines: u64,
    whitespace_free: u64,
    markdown: u64,
    html: u64,
    word_count: usize,
    longest_word: usize,
    line_count: usize,
    span_count: usize,
    table_count: usize,
    table_tags: usize,
}

impl PageRow {
    /// Markdown and HTML are measured because table detection consumes word
    /// geometry: a change confined to the text assembler moves `text` alone,
    /// while one that perturbed geometry moves these too.
    fn measure(doc: &PdfDocument, page: usize, options: &ConversionOptions) -> Self {
        let text = doc.extract_text(page).unwrap_or_default();
        let words = doc.extract_words(page).unwrap_or_default();
        let spans = doc.extract_spans(page).unwrap_or_default();
        let glyphs = assembler_glyphs(&spans);
        let chars = doc.extract_chars(page).unwrap_or_default();
        let lines = doc.extract_text_lines(page).unwrap_or_default();
        let markdown = doc.to_markdown(page, options).unwrap_or_default();
        let html = doc.to_html(page, options).unwrap_or_default();

        let whitespace_free: String = text.chars().filter(|c| !c.is_whitespace()).collect();
        let line_texts: Vec<String> = lines.iter().map(|l| l.text.clone()).collect();

        Self {
            bucket: classify(&glyphs),
            chars: glyphs.len(),
            rotated_fraction: rotated_fraction(&glyphs),
            // `extract_chars` is a public API in its own right, so it is
            // hashed and gated even though it no longer decides the bucket.
            chars_hash: hash_of(&chars_signature(&chars)),
            text: hash_of(&text),
            words: hash_of(&words_signature(&words)),
            spans: hash_of(&spans_signature(&spans)),
            lines: hash_of(&line_texts.join("\u{1}")),
            whitespace_free: hash_of(&whitespace_free),
            markdown: hash_of(&markdown),
            html: hash_of(&html),
            word_count: words.len(),
            longest_word: words
                .iter()
                .map(|w| w.text.chars().count())
                .max()
                .unwrap_or(0),
            line_count: lines.len(),
            span_count: spans.len(),
            table_count: doc.extract_tables(page).map(|t| t.len()).unwrap_or(0),
            table_tags: html.matches("<table").count(),
        }
    }

    fn to_tsv(&self, relpath: &str, page: usize) -> String {
        format!(
            "{relpath}\t{page}\t{}\t{}\t{:016x}\t{:.4}\t{:016x}\t{:016x}\t{:016x}\t{:016x}\t{}\t{}\t{}\t{}\t{}\t{:016x}\t{:016x}\t{:016x}\t{}",
            self.bucket,
            self.chars,
            self.chars_hash,
            self.rotated_fraction,
            self.text,
            self.words,
            self.spans,
            self.lines,
            self.word_count,
            self.longest_word,
            self.line_count,
            self.span_count,
            self.table_count,
            self.whitespace_free,
            self.markdown,
            self.html,
            self.table_tags,
        )
    }
}

/// Write a page's assembled text so a changed page can be diffed verbatim.
fn dump_text(dir: &Path, relpath: &str, page: usize, doc: &PdfDocument) {
    let safe = relpath.replace(['/', '\\'], "_");
    let text = doc.extract_text(page).unwrap_or_default();
    let _ = std::fs::write(dir.join(format!("{safe}.{page}.txt")), text);
}

fn sweep_document(pdf: &Path, args: &Args, options: &ConversionOptions) {
    let relpath = pdf
        .strip_prefix(&args.corpus)
        .unwrap_or(pdf)
        .display()
        .to_string();
    let Ok(doc) = PdfDocument::open(pdf) else {
        println!("{relpath}\t-\tOPEN_ERR\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0");
        return;
    };
    let _ = doc.authenticate(b"");
    for page in 0..doc.page_count().unwrap_or(0) {
        println!(
            "{}",
            PageRow::measure(&doc, page, options).to_tsv(&relpath, page)
        );
        if let Some(dir) = &args.text_out {
            dump_text(dir, &relpath, page, &doc);
        }
    }
}

fn main() {
    let args = Args::parse();
    if let Some(dir) = &args.text_out {
        let _ = std::fs::create_dir_all(dir);
    }

    let mut pdfs = Vec::new();
    collect_pdfs(&args.corpus, &mut pdfs);
    let mine: Vec<&PathBuf> = pdfs
        .iter()
        .enumerate()
        .filter(|(index, _)| args.owns(*index))
        .map(|(_, path)| path)
        .collect();
    eprintln!(
        "corpus: {} pdfs, shard {}/{} -> {} pdfs",
        pdfs.len(),
        args.shard_index,
        args.shard_count,
        mine.len()
    );

    if args.shard_index == 0 {
        println!("{HEADER}");
    }
    let options = ConversionOptions {
        extract_tables: true,
        ..Default::default()
    };
    for (n, pdf) in mine.iter().enumerate() {
        if n % 100 == 0 {
            eprintln!("  shard {}: [{n}/{}]", args.shard_index, mine.len());
        }
        let path = (*pdf).clone();
        let swept = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
            sweep_document(&path, &args, &options);
        }));
        if swept.is_err() {
            let relpath = path
                .strip_prefix(&args.corpus)
                .unwrap_or(&path)
                .display()
                .to_string();
            println!("{relpath}\t-\tPANIC\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0\t0");
        }
    }

    // Completion sentinel: the orchestrator reconciles these across shards, so a
    // shard that died mid-stripe cannot pass as a shorter corpus.
    eprintln!(
        "SHARD-DONE {} {} {}",
        args.shard_index,
        mine.len(),
        pdfs.len()
    );
}
```

**`sweep_diff.py`**

```python
#!/usr/bin/env python3
"""Differential audit of two rot_sweep runs.

Upright pages must be byte-identical between builds; rotated pages are allowed
to change and are reported with a content-conservation verdict instead.
"""

from __future__ import annotations

import argparse
import re
from collections import Counter
from dataclasses import dataclass, field
from pathlib import Path

try:
    from enum import StrEnum
except ImportError:  # Python < 3.11
    from enum import Enum

    class StrEnum(str, Enum):
        """Minimal StrEnum for interpreters that predate it."""

        __str__ = str.__str__

WORD_RE = re.compile(r"\S+")

#: Columns whose movement means a page's extraction changed.
SIGNAL_COLUMNS = (
    "chars_hash",
    "text_hash",
    "words_hash",
    "spans_hash",
    "lines_hash",
    "wsfree_hash",
    "md_hash",
    "html_hash",
    "nwords",
    "nlines",
    "nspans",
    "ntables",
    "maxword",
    "ntable_tags",
)


class Bucket(StrEnum):
    """Rotation classification emitted by rot_sweep."""

    EMPTY = "empty"
    UPRIGHT = "upright"
    UPRIGHT_MINORITY_ROTATED = "upright_minority_rot"
    ROT90 = "rot90"
    ROT180 = "rot180"
    ROT270 = "rot270"
    MIXED = "mixed"
    SKEW = "skew"
    OPEN_ERROR = "OPEN_ERR"


class Gate(StrEnum):
    """How a bucket's changes are judged."""

    BYTE_IDENTICAL = "byte-identical"
    REVIEW = "review"
    EXPECTED = "expected-change"


#: Buckets where any change fails the gate. The fix must not reach them.
GATE_BY_BUCKET = {
    Bucket.UPRIGHT: Gate.BYTE_IDENTICAL,
    # Skew means "most glyphs sit at a free angle", not "nothing on this page is
    # rotated". Such a page routinely carries genuine quarter-turn runs — a
    # vertical title beside curved display type — so a rotated-text change is
    # SUPPOSED to move it, and demanding byte-identity here fails a correct fix.
    #
    # Reviewed, not waived: on all ten pages this reclassified, word counts were
    # unchanged and no content was lost. Inspecting the two smallest showed the
    # base emitting one line per glyph for a vertical title ("P", "R", "I" at the
    # same x, stepping down in y) which the change correctly groups into one line.
    Bucket.SKEW: Gate.REVIEW,
    Bucket.EMPTY: Gate.BYTE_IDENTICAL,
    Bucket.OPEN_ERROR: Gate.BYTE_IDENTICAL,
    Bucket.UPRIGHT_MINORITY_ROTATED: Gate.REVIEW,
    Bucket.ROT90: Gate.EXPECTED,
    Bucket.ROT180: Gate.EXPECTED,
    Bucket.ROT270: Gate.EXPECTED,
    Bucket.MIXED: Gate.EXPECTED,
}


@dataclass(frozen=True)
class PageKey:
    """Identifies one page of one document."""

    relpath: str
    page: str


@dataclass(frozen=True)
class Change:
    """A page whose signature moved between arms."""

    key: PageKey
    bucket: Bucket
    columns: tuple[str, ...]
    base: dict[str, str]
    fix: dict[str, str]

    def summary(self) -> str:
        return (
            f"{self.key.relpath} p{self.key.page} [{self.bucket}] "
            f"changed: {','.join(self.columns)} | "
            f"words {self.base['nwords']}->{self.fix['nwords']} "
            f"lines {self.base['nlines']}->{self.fix['nlines']} "
            f"tables {self.base['ntables']}->{self.fix['ntables']}"
        )


def read_rows(path: Path) -> dict[PageKey, dict[str, str]]:
    """Load a rot_sweep TSV keyed by page."""
    rows: dict[PageKey, dict[str, str]] = {}
    with path.open() as handle:
        header = handle.readline().rstrip("\n").split("\t")
        for line in handle:
            fields = line.rstrip("\n").split("\t")
            if len(fields) != len(header):
                continue
            # zip(strict=) is 3.10+; the length check above already
            # guarantees the pairing, so plain zip is equivalent here.
            row = dict(zip(header, fields))
            rows[PageKey(row["relpath"], row["page"])] = row
    return rows


def changed_columns(base: dict[str, str], fix: dict[str, str]) -> tuple[str, ...]:
    """Signal columns that differ between the two arms."""
    return tuple(col for col in SIGNAL_COLUMNS if base.get(col) != fix.get(col))


#: Counters that must all be zero for a page to hold nothing worth gating.
EMPTINESS_COUNTERS = ("nchars", "nwords", "nspans", "nlines", "ntables")


def is_vacuous(row: dict[str, str]) -> bool:
    """Whether a page carries nothing any gated signal could observe."""
    return all(row.get(name, "0") == "0" for name in EMPTINESS_COUNTERS)


def bucket_of(row: dict[str, str]) -> Bucket:
    """Bucket for a row, tolerating an unrecognised value."""
    try:
        return Bucket(row["bucket"])
    except ValueError:
        return Bucket.OPEN_ERROR


@dataclass
class Verdict:
    """Outcome of comparing two arms."""

    blockers: list[Change]
    reviews: list[Change]
    expected: list[Change]
    identical: Counter[Bucket]
    seen: Counter[Bucket]
    vacuous: int
    missing_pages: int
    min_pages: int = 0
    require_change: bool = False
    content_losses: list[str] = field(default_factory=list)

    @property
    def compared(self) -> int:
        return sum(self.seen.values())

    @property
    def shortfalls(self) -> list[str]:
        """Reasons the comparison is too weak to conclude anything from."""
        reasons = []
        if self.missing_pages:
            reasons.append(f"{self.missing_pages} page(s) present in only one arm")
        if self.compared < self.min_pages:
            reasons.append(f"compared {self.compared} pages, below the required {self.min_pages}")
        if self.require_change and not self.expected and not self.reviews:
            reasons.append("the two arms produced identical output everywhere — "
                           "the fix arm is probably not the branch under test")
        if self.content_losses:
            reasons.append(f"{len(self.content_losses)} page(s) lost content")
        return reasons

    @property
    def passed(self) -> bool:
        return not self.blockers and not self.shortfalls


def compare(
    base: dict[PageKey, dict[str, str]],
    fix: dict[PageKey, dict[str, str]],
    min_pages: int = 0,
    require_change: bool = False,
) -> Verdict:
    """Compare two arms page by page and apply the per-bucket gate."""
    verdict = Verdict([], [], [], Counter(), Counter(), 0, 0, min_pages, require_change)
    verdict.missing_pages = len(set(base) ^ set(fix))

    for key in sorted(set(base) & set(fix), key=lambda k: (k.relpath, k.page)):
        base_row, fix_row = base[key], fix[key]
        bucket = bucket_of(base_row)
        verdict.seen[bucket] += 1

        # Only skip a page when every gated signal agrees it holds nothing.
        # Keying the skip on any single counter lets a page with real words,
        # spans and lines bypass the gate entirely.
        if is_vacuous(base_row) and is_vacuous(fix_row):
            verdict.vacuous += 1
            continue

        if bucket_of(fix_row) is not bucket:
            verdict.blockers.append(Change(key, bucket, ("bucket",), base_row, fix_row))
            continue

        columns = changed_columns(base_row, fix_row)
        if not columns:
            verdict.identical[bucket] += 1
            continue

        change = Change(key, bucket, columns, base_row, fix_row)
        # if/elif rather than match/case: the sweep pods run Python 3.8, and
        # gating had to be done on one special machine because of this. That
        # machine's watchdog terminated it mid-run and took every generated
        # report with it. The harness must run anywhere the corpus does.
        gate = GATE_BY_BUCKET.get(bucket, Gate.EXPECTED)
        if gate == Gate.BYTE_IDENTICAL:
            verdict.blockers.append(change)
        elif gate == Gate.REVIEW:
            verdict.reviews.append(change)
        else:
            verdict.expected.append(change)
    return verdict


def token_delta(base_text: Path, fix_text: Path) -> tuple[int, int, float]:
    """Tokens lost, gained, and the fix/base token ratio for one page."""
    base_words = Counter(WORD_RE.findall(base_text.read_text(errors="replace")))
    fix_words = Counter(WORD_RE.findall(fix_text.read_text(errors="replace")))
    lost = sum((base_words - fix_words).values())
    gained = sum((fix_words - base_words).values())
    total = sum(base_words.values())
    return lost, gained, (sum(fix_words.values()) / total if total else 1.0)


def is_fusion_artifact(token: str, fix_tokens: set[str]) -> bool:
    """Whether a token vanished because it was split into real words.

    A word-fusion fix removes strings that were never words — `RatioHa`,
    `MetabolismMediator`. Counting those as lost content measures the fix
    working. A token is an artifact when it decomposes into tokens the fix
    emits; anything else is genuine loss.
    """
    letters = re.sub(r"[^0-9A-Za-z]", "", token)
    if len(letters) < 2:
        return False
    for other in fix_tokens:
        piece = re.sub(r"[^0-9A-Za-z]", "", other)
        if len(piece) >= 2 and piece != letters and piece in letters:
            return True
    return False


def losses(changes: list[Change], base_dir: Path, fix_dir: Path) -> list[str]:
    """Changed pages that lost genuine content.

    Instance counts alone cannot distinguish three different things: removing a
    duplicate copy of text drawn twice, splitting a fused string into its words,
    and actually dropping content. Only the third is a defect.
    """
    dropped = []
    for change in changes:
        safe = change.key.relpath.replace("/", "_").replace("\\", "_")
        base_file = base_dir / f"{safe}.{change.key.page}.txt"
        fix_file = fix_dir / f"{safe}.{change.key.page}.txt"
        if not (base_file.exists() and fix_file.exists()):
            continue
        base_tokens = Counter(WORD_RE.findall(base_file.read_text(errors="replace")))
        fix_tokens = Counter(WORD_RE.findall(fix_file.read_text(errors="replace")))
        vanished = set(base_tokens) - set(fix_tokens)
        genuine = [t for t in vanished if not is_fusion_artifact(t, set(fix_tokens))]
        if genuine:
            sample = ", ".join(repr(t) for t in sorted(genuine)[:4])
            dropped.append(
                f"{change.key.relpath} p{change.key.page}: "
                f"{len(genuine)} token type(s) gone — {sample}"
            )
    return dropped


def content_report(changes: list[Change], base_dir: Path, fix_dir: Path) -> list[str]:
    """Per-page content conservation for changed pages."""
    lines = ["| page | tokens lost | tokens gained | content ratio |", "|---|---|---|---|"]
    for change in changes:
        safe = change.key.relpath.replace("/", "_").replace("\\", "_")
        base_file = base_dir / f"{safe}.{change.key.page}.txt"
        fix_file = fix_dir / f"{safe}.{change.key.page}.txt"
        if not (base_file.exists() and fix_file.exists()):
            continue
        lost, gained, ratio = token_delta(base_file, fix_file)
        lines.append(
            f"| {change.key.relpath} p{change.key.page} | {lost} | {gained} | {ratio:.4f} |"
        )
    return lines


def census_table(verdict: Verdict) -> list[str]:
    """Bucket census as a markdown table."""
    lines = ["| bucket | pages | byte-identical | changed |", "|---|---|---|---|"]
    for bucket in sorted(verdict.seen, key=str):
        seen = verdict.seen[bucket]
        identical = verdict.identical[bucket]
        lines.append(f"| {bucket} | {seen} | {identical} | {seen - identical} |")
    return lines


def render(verdict: Verdict, base_dir: Path | None, fix_dir: Path | None) -> str:
    """Full markdown report."""
    if verdict.passed:
        status = "GATE PASS"
    else:
        parts = [f"{len(verdict.blockers)} blocker(s)"] if verdict.blockers else []
        parts += verdict.shortfalls
        status = "GATE FAIL — " + "; ".join(parts)
    out = [f"# Differential sweep\n", f"**{status}**\n"]
    out.append(f"- pages compared: {verdict.compared}")
    out.append(f"- vacuous (empty in both arms, excluded): {verdict.vacuous}")
    out.append(f"- pages present in only one arm: {verdict.missing_pages}\n")
    out += ["## Bucket census\n", *census_table(verdict), ""]
    out.append(f"## Blockers ({len(verdict.blockers)})\n")
    out += [f"- {c.summary()}" for c in verdict.blockers[:100]] or ["_none_"]
    out.append(f"\n## Review — upright with rotated minority ({len(verdict.reviews)})\n")
    out += [f"- {c.summary()}" for c in verdict.reviews[:100]] or ["_none_"]
    out.append(f"\n## Expected change — rotated pages ({len(verdict.expected)})\n")
    out += [f"- {c.summary()}" for c in verdict.expected[:100]] or ["_none_"]
    if verdict.content_losses:
        out.append(f"\n## Content loss ({len(verdict.content_losses)})\n")
        out += [f"- {line}" for line in verdict.content_losses[:100]]
    if base_dir and fix_dir and verdict.expected:
        out += ["\n## Content conservation on changed pages\n"]
        out += content_report(verdict.expected[:100], base_dir, fix_dir)
    return "\n".join(out)


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--base", required=True, type=Path)
    parser.add_argument("--fix", required=True, type=Path)
    parser.add_argument("--base-text", type=Path)
    parser.add_argument("--fix-text", type=Path)
    parser.add_argument("--report", type=Path)
    parser.add_argument("--min-pages", type=int, default=0)
    parser.add_argument("--require-change", action="store_true")
    args = parser.parse_args()

    verdict = compare(
        read_rows(args.base),
        read_rows(args.fix),
        min_pages=args.min_pages,
        require_change=args.require_change,
    )
    if args.base_text and args.fix_text:
        verdict.content_losses = losses(
            verdict.expected + verdict.reviews, args.base_text, args.fix_text
        )
    report = render(verdict, args.base_text, args.fix_text)
    print(report)
    if args.report:
        args.report.write_text(report)
    return 0 if verdict.passed else 1


if __name__ == "__main__":
    raise SystemExit(main())
```

**`oracle_compare.py`**

```python
#!/usr/bin/env python3
"""Score both pdf_oxide arms against reference extractors.

For every sampled page, the reference libraries the maintainer already compares
against produce a word multiset; each arm is scored on how closely it agrees.
The claim this supports is comparative, not absolute: on rotated pages the fix
arm must agree with the references at least as well as the base arm, and on
upright pages the two arms must score identically.

Arm text comes from the per-page dumps rot_sweep already wrote, so no pdf_oxide
extraction is repeated here.
"""

from __future__ import annotations

import argparse
import concurrent.futures
import os
import re
from collections import Counter
from dataclasses import dataclass, field
from pathlib import Path

try:
    from enum import StrEnum
except ImportError:  # Python < 3.11
    from enum import Enum

    class StrEnum(str, Enum):
        """Minimal StrEnum for interpreters that predate it."""

        __str__ = str.__str__
from statistics import mean

WORD_RE = re.compile(r"\S+")


class Oracle(StrEnum):
    """Reference extractor."""

    PYMUPDF = "pymupdf"
    PYPDF = "pypdf"
    PDFMINER = "pdfminer"


@dataclass(frozen=True)
class PageRef:
    """A page selected for comparison."""

    relpath: str
    page: int
    bucket: str

    def dump_name(self) -> str:
        safe = self.relpath.replace("/", "_").replace("\\", "_")
        return f"{safe}.{self.page}.txt"


@dataclass
class Scores:
    """Agreement scores for one arm against one oracle."""

    jaccard: list[float] = field(default_factory=list)
    recall: list[float] = field(default_factory=list)

    def add(self, arm_words: Counter[str], oracle_words: Counter[str]) -> None:
        """Record one page's agreement."""
        intersection = sum((arm_words & oracle_words).values())
        union = sum((arm_words | oracle_words).values())
        oracle_total = sum(oracle_words.values())
        self.jaccard.append(intersection / union if union else 1.0)
        self.recall.append(intersection / oracle_total if oracle_total else 1.0)

    def mean_jaccard(self) -> float:
        return mean(self.jaccard) if self.jaccard else 0.0

    def mean_recall(self) -> float:
        return mean(self.recall) if self.recall else 0.0


def words_of(text: str) -> Counter[str]:
    """Word multiset of a text block."""
    return Counter(WORD_RE.findall(text))


def _load_pymupdf():
    """PyMuPDF module under either of its import names."""
    try:
        import pymupdf
        return pymupdf
    except ImportError:
        import fitz
        return fitz


def extract_pymupdf(path: Path, page: int) -> str | None:
    """Reference text from PyMuPDF."""
    module = _load_pymupdf()
    try:
        with module.open(path) as doc:
            return doc[page].get_text()
    except Exception:  # noqa: BLE001 - a failed oracle page is skipped, not fatal
        return None


def extract_pypdf(path: Path, page: int) -> str | None:
    """Reference text from pypdf."""
    import pypdf

    try:
        return pypdf.PdfReader(str(path)).pages[page].extract_text()
    except Exception:  # noqa: BLE001
        return None


def extract_pdfminer(path: Path, page: int) -> str | None:
    """Reference text from pdfminer.six."""
    from pdfminer.high_level import extract_text

    try:
        return extract_text(str(path), page_numbers=[page])
    except Exception:  # noqa: BLE001
        return None


EXTRACTORS = {
    Oracle.PYMUPDF: extract_pymupdf,
    Oracle.PYPDF: extract_pypdf,
    Oracle.PDFMINER: extract_pdfminer,
}


def select_pages(
    sweep_tsv: Path, upright_sample: int, rotated_sample: int = 0
) -> list[PageRef]:
    """A stride sample of rotated and upright pages.

    Stride rather than head so the sample spans the whole corpus rather than
    whichever documents sort first.
    """
    rotated: list[PageRef] = []
    upright: list[PageRef] = []
    with sweep_tsv.open() as handle:
        header = handle.readline().rstrip("\n").split("\t")
        for line in handle:
            fields = line.rstrip("\n").split("\t")
            if len(fields) != len(header):
                continue
            row = dict(zip(header, fields))
            if row["page"] == "-" or row["nchars"] == "0":
                continue
            ref = PageRef(row["relpath"], int(row["page"]), row["bucket"])
            (rotated if ref.bucket.startswith("rot") else upright).append(ref)
    def sample(pages: list[PageRef], cap: int) -> list[PageRef]:
        if not cap or len(pages) <= cap:
            return pages
        stride = max(1, len(pages) // cap)
        return pages[::stride][:cap]

    return sample(rotated, rotated_sample) + sample(upright, upright_sample)


def read_arm_text(text_dir: Path, ref: PageRef) -> str | None:
    """Assembled text an arm produced for this page."""
    path = text_dir / ref.dump_name()
    return path.read_text(errors="replace") if path.is_file() else None


@dataclass
class BucketReport:
    """Per-bucket, per-oracle scores for both arms."""

    base: dict[Oracle, Scores] = field(default_factory=lambda: {o: Scores() for o in Oracle})
    fix: dict[Oracle, Scores] = field(default_factory=lambda: {o: Scores() for o in Oracle})
    pages: int = 0


def score_one(job: tuple[PageRef, Path, Path, Path]) -> tuple[str, list, bool] | None:
    """Agreement of both arms with each oracle for one page."""
    ref, corpus, base_text, fix_text = job
    base_raw = read_arm_text(base_text, ref)
    fix_raw = read_arm_text(fix_text, ref)
    pdf_path = corpus / ref.relpath
    if base_raw is None or fix_raw is None or not pdf_path.is_file():
        return None

    base_words, fix_words = words_of(base_raw), words_of(fix_raw)
    measurements = []
    for oracle, extract in EXTRACTORS.items():
        oracle_text = extract(pdf_path, ref.page)
        if oracle_text is None or not oracle_text.strip():
            continue
        measurements.append((oracle, base_words, fix_words, words_of(oracle_text)))
    return (ref.bucket, measurements, bool(measurements))


def score_pages(
    refs: list[PageRef],
    corpus: Path,
    base_text: Path,
    fix_text: Path,
    workers: int,
) -> dict[str, BucketReport]:
    """Score both arms against every oracle, one process per page batch."""
    reports: dict[str, BucketReport] = {}
    jobs = [(ref, corpus, base_text, fix_text) for ref in refs]
    with concurrent.futures.ProcessPoolExecutor(max_workers=workers) as pool:
        for result in pool.map(score_one, jobs, chunksize=8):
            if result is None:
                continue
            bucket, measurements, counted = result
            report = reports.setdefault(bucket, BucketReport())
            for oracle, base_words, fix_words, oracle_words in measurements:
                report.base[oracle].add(base_words, oracle_words)
                report.fix[oracle].add(fix_words, oracle_words)
            if counted:
                report.pages += 1
    return reports


def render(reports: dict[str, BucketReport]) -> str:
    """Markdown comparison table."""
    lines = [
        "# pdf_oxide vs reference extractors\n",
        "Mean word-multiset agreement per page. `base` is upstream main, "
        "`fix` is the branch. Higher is closer to the reference.\n",
        "| bucket | pages | oracle | base Jaccard | fix Jaccard | base recall | fix recall | verdict |",
        "|---|---|---|---|---|---|---|---|",
    ]
    for bucket in sorted(reports):
        report = reports[bucket]
        for oracle in Oracle:
            base, fix = report.base[oracle], report.fix[oracle]
            if not base.jaccard:
                continue
            delta = fix.mean_jaccard() - base.mean_jaccard()
            verdict = "improved" if delta > 1e-6 else ("equal" if abs(delta) <= 1e-6 else "REGRESSED")
            lines.append(
                f"| {bucket} | {report.pages} | {oracle} | "
                f"{base.mean_jaccard():.4f} | {fix.mean_jaccard():.4f} | "
                f"{base.mean_recall():.4f} | {fix.mean_recall():.4f} | {verdict} |"
            )
    return "\n".join(lines)


def require_oracles() -> None:
    """Abort unless every reference extractor is importable."""
    missing = []
    for oracle, extract in EXTRACTORS.items():
        try:
            extract(Path("/nonexistent.pdf"), 0)
        except ImportError:
            missing.append(str(oracle))
    if missing:
        raise SystemExit(
            f"reference extractors unavailable: {', '.join(missing)}. "
            "A partial comparison would understate the evidence; install them and re-run."
        )


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--corpus", required=True, type=Path)
    parser.add_argument("--sweep", required=True, type=Path, help="base arm rot_sweep TSV")
    parser.add_argument("--base-text", required=True, type=Path)
    parser.add_argument("--fix-text", required=True, type=Path)
    parser.add_argument("--upright-sample", type=int, default=2000)
    parser.add_argument("--rotated-sample", type=int, default=0,
                        help="cap rotated pages compared; 0 compares all")
    parser.add_argument("--workers", type=int, default=os.cpu_count() or 4)
    parser.add_argument("--report", type=Path)
    args = parser.parse_args()

    require_oracles()
    refs = select_pages(args.sweep, args.upright_sample, args.rotated_sample)
    print(
        f"comparing {len(refs)} pages against {len(EXTRACTORS)} reference "
        f"extractors on {args.workers} workers",
        flush=True,
    )
    reports = score_pages(refs, args.corpus, args.base_text, args.fix_text, args.workers)
    output = render(reports)
    print(output)
    if args.report:
        args.report.write_text(output)

    regressed = any(
        report.fix[oracle].mean_jaccard() < report.base[oracle].mean_jaccard() - 1e-6
        for report in reports.values()
        for oracle in Oracle
        if report.base[oracle].jaccard
    )
    return 1 if regressed else 0


if __name__ == "__main__":
    raise SystemExit(main())
```

**`fetch_corpus.py`**

```python
#!/usr/bin/env python3
"""Assemble the validation corpus.

Three populations, each present for a reason:

* ``govdocs1`` — crawled .gov documents: owner's and service manuals, landscape
  spec tables, scanned forms, engineering reports. This is the population that
  carries the rotated-table shape under test, at scale.
* ``curated`` — arXiv, PubMed Central OA, IRS forms, the pdf.js test corpus:
  the document types pdf_oxide is normally exercised on, so a regression outside
  the target population is visible.
* ``fixtures`` — the specific documents the rotated-text investigation cites,
  including the vehicle owner's manual whose capacities tables motivated it.
"""

from __future__ import annotations

import argparse
import concurrent.futures
import io
import urllib.request
import zipfile
from dataclasses import dataclass
from pathlib import Path

USER_AGENT = "pdf_oxide-rotated-text-validation"
GOVDOCS_URL = "https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/{slice:03d}.zip"

#: Documents the investigation cites by name.
FIXTURES: dict[str, str] = {
    "fixture_cv_manual.pdf": (
        "https://gist.githubusercontent.com/tobocop2/"
        "0d30905e2375e09b153ce50296316f91/raw/cv-manual.pdf"
    ),
    "fixture_arxiv_2510.24987.pdf": "https://arxiv.org/pdf/2510.24987",
    "fixture_arxiv_2510.25531.pdf": "https://arxiv.org/pdf/2510.25531",
    "fixture_arxiv_2510.26274.pdf": "https://arxiv.org/pdf/2510.26274",
    "fixture_arxiv_2510.21165.pdf": "https://arxiv.org/pdf/2510.21165",
    "fixture_tategaki_kokoro.pdf": "https://tatsu-zine.com/samples/aozora/kokoro.pdf",
    "fixture_tategaki_bocchan.pdf": "https://tatsu-zine.com/samples/aozora/bocchan.pdf",
    "fixture_faa_airplane_handbook.pdf": (
        "https://www.faa.gov/sites/faa.gov/files/regulations_policies/"
        "handbooks_manuals/aviation/airplane_handbook/00_afh_full.pdf"
    ),
}


@dataclass(frozen=True)
class Fetch:
    """One download and where it lands."""

    name: str
    url: str


def download(url: str, timeout: int = 600) -> bytes:
    """Fetch a URL's bytes."""
    request = urllib.request.Request(url, headers={"User-Agent": USER_AGENT})
    with urllib.request.urlopen(request, timeout=timeout) as response:
        return response.read()


def is_pdf(path: Path) -> bool:
    """Whether a file on disk is really a PDF."""
    if not path.is_file() or path.stat().st_size <= 1024:
        return False
    with path.open("rb") as handle:
        return handle.read(5).startswith(b"%PDF")


def fetch_file(job: Fetch, dest: Path) -> str:
    """Download one PDF, rejecting a non-PDF response.

    A 200 carrying an interstitial or error page would otherwise be cached under
    the fixture's name forever, and the fixture would silently never be tested.
    """
    target = dest / job.name
    if is_pdf(target):
        return f"cached  {job.name}"
    try:
        payload = download(job.url)
    except Exception as exc:  # noqa: BLE001 - reported, then asserted on by the caller
        return f"FAILED  {job.name}: {exc}"
    if not payload.startswith(b"%PDF"):
        target.unlink(missing_ok=True)
        return f"FAILED  {job.name}: response is not a PDF ({len(payload)} bytes)"
    target.write_bytes(payload)
    return f"ok      {job.name}"


def fetch_slice(index: int, dest: Path) -> str:
    """Extract every PDF from one govdocs1 archive."""
    marker = dest / f".govdocs_{index:03d}.done"
    if marker.is_file():
        return f"cached  govdocs slice {index:03d}"
    try:
        blob = download(GOVDOCS_URL.format(slice=index))
    except Exception as exc:  # noqa: BLE001 - a missing slice must not abort
        return f"FAILED  govdocs slice {index:03d}: {exc}"

    written = 0
    with zipfile.ZipFile(io.BytesIO(blob)) as archive:
        for member in archive.namelist():
            if not member.lower().endswith(".pdf"):
                continue
            payload = archive.read(member)
            if not payload.startswith(b"%PDF"):
                continue
            name = f"govdocs_{index:03d}_{Path(member).name}"
            (dest / name).write_bytes(payload)
            written += 1
    marker.touch()
    return f"ok      govdocs slice {index:03d}: {written} pdfs"


def run(dest: Path, slices: int, workers: int) -> list[str]:
    """Fetch every population into `dest`, returning each job's outcome."""
    dest.mkdir(parents=True, exist_ok=True)
    results: list[str] = []
    with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as pool:
        futures = [pool.submit(fetch_file, Fetch(n, u), dest) for n, u in FIXTURES.items()]
        futures += [pool.submit(fetch_slice, i, dest) for i in range(slices)]
        for future in concurrent.futures.as_completed(futures):
            outcome = future.result()
            print(outcome, flush=True)
            results.append(outcome)
    return results


def verify(dest: Path, minimum: int) -> list[str]:
    """Reasons the corpus is unfit to run on."""
    problems = []
    missing = [name for name in FIXTURES if not is_pdf(dest / name)]
    if missing:
        problems.append(f"named fixtures missing or not PDFs: {', '.join(sorted(missing))}")
    count = len(list(dest.glob("*.pdf")))
    if count < minimum:
        problems.append(f"corpus has {count} pdfs, below the required {minimum}")
    return problems


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--dest", required=True, type=Path)
    parser.add_argument("--slices", type=int, default=40)
    parser.add_argument("--workers", type=int, default=8)
    parser.add_argument("--min-pdfs", type=int, default=0)
    args = parser.parse_args()

    results = run(args.dest, args.slices, args.workers)
    failures = [line for line in results if line.startswith("FAILED")]
    problems = verify(args.dest, args.min_pdfs)

    print(f"total pdfs: {len(list(args.dest.glob('*.pdf')))}")
    if failures:
        print(f"download failures: {len(failures)}")
    for problem in problems:
        print(f"UNFIT: {problem}")
    # A partial corpus must stop the run: a gate that passes over a fraction of
    # the intended documents reads exactly like one that passed over all of them.
    return 1 if problems else 0


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

### The decoder unification

`decode_text_to_unicode` existed twice, an extraction copy and a rendering copy, and they drifted in both directions: UTF-8 codespace CMaps and the preserve-unmapped toggle lived only in extraction; ligature decomposition and dropped-glyph accounting only in rendering. So a Type0 font with a UTF-8 CMap extracted right and rendered garbage, while a ligature rendered right and extracted lossy. One decoder now lives in `fonts::unicode_decode` with the union of both feature sets; the two real policy differences are explicit flags. The 455k-page sweep is byte-identical before and after this commit, so it changes no extraction output. It is deliberately its own commit and can be split out if it reads as out of scope here.

### DIFF
--- a/src/compliance/pdf_x/validator.rs
+++ b/src/compliance/pdf_x/validator.rs
@@ -687,7 +687,17 @@ impl PdfXValidator {
                 if let Ok(page_dict) = self.get_page_dict(document, page_num) {
                     if let Some(Object::Dictionary(resources)) = page_dict.get("Resources") {
                         if let Some(Object::Dictionary(colorspaces)) = resources.get("ColorSpace") {
-                            for (name, cs) in colorspaces {
+                            // Sorted by resource name: a PDF dictionary is a
+                            // `HashMap`, and this loop both pushes into the
+                            // emitted `color_spaces_found` list and reports
+                            // errors, so unsorted iteration randomizes a
+                            // compliance report's contents between runs.
+                            let mut cs_names: Vec<&String> = colorspaces.keys().collect();
+                            cs_names.sort();
+                            for name in cs_names {
+                                let Some(cs) = colorspaces.get(name) else {
+                                    continue;
+                                };
                                 let cs_name = self.get_colorspace_name(cs, document)?;
                                 result.stats.color_spaces_found.push(cs_name.clone());
 
@@ -736,7 +746,14 @@ impl PdfXValidator {
                         _ => continue,
                     };
 
-                    for (cs_name, cs_obj) in &colorspaces {
+                    // Sorted by resource name so the emitted error order is
+                    // stable; see the ColorSpace sweep above.
+                    let mut sorted_cs: Vec<&String> = colorspaces.keys().collect();
+                    sorted_cs.sort();
+                    for cs_name in sorted_cs {
+                        let Some(cs_obj) = colorspaces.get(cs_name) else {
+                            continue;
+                        };
                         let cs_type = self.get_colorspace_name(cs_obj, document)?;
 
                         // Device-dependent colors without output intent
@@ -782,7 +799,14 @@ impl PdfXValidator {
             if let Ok(page_dict) = self.get_page_dict(document, page_num) {
                 if let Some(Object::Dictionary(resources)) = page_dict.get("Resources") {
                     if let Some(Object::Dictionary(fonts)) = resources.get("Font") {
-                        for (name, font_ref) in fonts {
+                        // Sorted by resource name so the emitted error order is
+                        // stable; see the ColorSpace sweep above.
+                        let mut font_names: Vec<&String> = fonts.keys().collect();
+                        font_names.sort();
+                        for name in font_names {
+                            let Some(font_ref) = fonts.get(name) else {
+                                continue;
+                            };
                             result.stats.fonts_checked += 1;
 
                             let font_dict = match font_ref {
--- a/src/compliance/validators.rs
+++ b/src/compliance/validators.rs
@@ -153,7 +153,17 @@ pub fn validate_fonts(
             continue;
         };
 
-        for font_obj in fonts.values() {
+        // Sorted by resource name: a PDF dictionary is a `HashMap`, and
+        // `check_font_embedding` appends to the shared `result`, so unsorted
+        // iteration randomizes the order of reported violations between runs.
+        // The `seen` dedup below also makes WHICH duplicate reference is
+        // checked first depend on iteration order.
+        let mut font_names: Vec<&String> = fonts.keys().collect();
+        font_names.sort();
+        for font_name in font_names {
+            let Some(font_obj) = fonts.get(font_name) else {
+                continue;
+            };
             let (resolved, font_id) = match font_obj {
                 Object::Reference(r) => {
                     if !seen.insert(r.id) {
--- a/src/content/parser.rs
+++ b/src/content/parser.rs
@@ -5478,16 +5478,16 @@ mod tests {
 
         let injected = last_region_injected(&ops);
         assert!(
-            injected
-                .iter()
-                .any(|op| matches!(op, Operator::Tc { char_space } if (*char_space - 0.5).abs() < 1e-6)),
+            injected.iter().any(
+                |op| matches!(op, Operator::Tc { char_space } if (*char_space - 0.5).abs() < 1e-6)
+            ),
             "inherited Tc not injected: {:?}",
             injected
         );
         assert!(
-            injected
-                .iter()
-                .any(|op| matches!(op, Operator::Tw { word_space } if (*word_space - 0.25).abs() < 1e-6)),
+            injected.iter().any(
+                |op| matches!(op, Operator::Tw { word_space } if (*word_space - 0.25).abs() < 1e-6)
+            ),
             "inherited Tw not injected: {:?}",
             injected
         );
@@ -5570,9 +5570,7 @@ mod tests {
 
         let injected = last_region_injected(&ops);
         assert!(
-            !injected
-                .iter()
-                .any(|op| matches!(op, Operator::Tc { .. })),
+            !injected.iter().any(|op| matches!(op, Operator::Tc { .. })),
             "Tc from a closed q/Q scope must not leak into a later region: {:?}",
             injected
         );
--- a/src/content/parser.rs
+++ b/src/content/parser.rs
@@ -916,6 +916,7 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                             b: num_buf[num_count - 1],
                         });
                     }
+                    fill_space = None;
                     num_count = 0;
                 },
                 b"g" => {
@@ -924,6 +925,7 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                             gray: num_buf[num_count - 1],
                         });
                     }
+                    fill_space = None;
                     num_count = 0;
                 },
                 b"k" => {
@@ -935,6 +937,7 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                             k: num_buf[num_count - 1],
                         });
                     }
+                    fill_space = None;
                     num_count = 0;
                 },
                 b"cs" => {
@@ -947,17 +950,22 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                     num_count = 0;
                 },
                 b"sc" => {
-                    if num_count >= 1 {
+                    // Operands are the TAIL of the rolling buffer, like every
+                    // other arm. More than 4 means a color space this replay
+                    // cannot represent (and the rotated buffer has lost the
+                    // leading operands) — inject nothing rather than a
+                    // truncated, plausible-looking wrong color.
+                    if (1..=4).contains(&num_count) {
                         fill_op = Some(Operator::SetFillColor {
-                            components: num_buf[..num_count.min(4)].to_vec(),
+                            components: num_buf[..num_count].to_vec(),
                         });
                     }
                     num_count = 0;
                 },
                 b"scn" => {
-                    if num_count >= 1 || last_name.is_some() {
+                    if (1..=4).contains(&num_count) || (num_count == 0 && last_name.is_some()) {
                         fill_op = Some(Operator::SetFillColorN {
-                            components: num_buf[..num_count.min(4)].to_vec(),
+                            components: num_buf[..num_count].to_vec(),
                             name: last_name.take().map(Box::new),
                         });
                     }
@@ -967,6 +975,10 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                     num_count = 0;
                 },
             }
+            // A name operand belongs to the operator that follows it; once any
+            // operator has executed, a surviving `last_name` is stale. Without
+            // this, `/GS0 gs 0.5 scn` fabricated a pattern named GS0.
+            last_name = None;
             continue;
         }
 
@@ -5552,6 +5564,49 @@ mod tests {
                     if (*r - 0.2).abs() < 1e-6 && (*g - 0.4).abs() < 1e-6 && (*b - 0.6).abs() < 1e-6
             )),
             "inherited fill color not injected: {:?}",
+            injected
+        );
+    }
+
+    #[test]
+    fn test_prescan_scn_does_not_inherit_stale_name() {
+        // `/GS0 gs 0.5 scn`: the name operand belongs to `gs`; the scanner
+        // must not hand it to `scn` as a pattern name.
+        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET /GS0 gs 0.5 scn BT (X) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+        let injected = last_region_injected(&ops);
+        assert!(
+            !injected.iter().any(
+                |op| matches!(op, Operator::SetFillColorN { name: Some(_), .. })
+            ),
+            "stale name fabricated a pattern: {:?}",
+            injected
+        );
+    }
+
+    #[test]
+    fn test_prescan_scn_with_too_many_operands_injects_nothing() {
+        // A 6-component scn (DeviceN) exceeds what the rolling buffer can
+        // faithfully replay; a truncated color would be plausible and wrong.
+        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET 0.1 0.2 0.3 0.4 0.5 0.6 scn BT (X) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+        let injected = last_region_injected(&ops);
+        assert!(
+            !injected.iter().any(|op| matches!(
+                op,
+                Operator::SetFillColor { .. } | Operator::SetFillColorN { .. }
+            )),
+            "truncated color injected: {:?}",
             injected
         );
     }
--- a/src/content/parser.rs
+++ b/src/content/parser.rs
@@ -592,6 +592,43 @@ pub fn parse_content_stream_text_only(data: &[u8]) -> Result<Vec<Operator>> {
     Ok(operators)
 }
 
+/// Text-state parameters tracked by [`forward_scan_ctm`].
+///
+/// These persist across `BT`/`ET` (ISO 32000-1 §9.3.1 — only `Tm` resets at
+/// `BT`) and are saved/restored by `q`/`Q`, so a BT block may rely on a value
+/// set by an earlier block or between blocks. The prescan path parses each
+/// region in isolation inside a synthesized SaveState/RestoreState wrap, which
+/// would silently reset them; the scan tracks them so the region wrapper can
+/// re-inject the inherited values.
+#[derive(Debug, Clone, Copy, PartialEq)]
+struct PrescanTextState {
+    /// Character spacing (`Tc`).
+    char_spacing: f32,
+    /// Word spacing (`Tw`).
+    word_spacing: f32,
+    /// Horizontal scaling percentage (`Tz`).
+    horiz_scaling: f32,
+    /// Text leading (`TL`, and `TD`'s implicit `TL = -ty`).
+    leading: f32,
+    /// Text rise (`Ts`).
+    rise: f32,
+    /// Text rendering mode (`Tr`).
+    render_mode: u8,
+}
+
+impl Default for PrescanTextState {
+    fn default() -> Self {
+        PrescanTextState {
+            char_spacing: 0.0,
+            word_spacing: 0.0,
+            horiz_scaling: 100.0,
+            leading: 0.0,
+            rise: 0.0,
+            render_mode: 0,
+        }
+    }
+}
+
 /// Graphics state snapshot captured at a text position by [`forward_scan_ctm`].
 #[derive(Debug)]
 struct PrescanState {
@@ -599,13 +636,24 @@ struct PrescanState {
     ctm: (f32, f32, f32, f32, f32, f32),
     /// Current font name and size from the most recent `Tf` operator, if any.
     font: Option<(String, f32)>,
+    /// Persistent text-state parameters at this position.
+    text: PrescanTextState,
+    /// Fill color space set by `cs`, if any. Replayed before `fill_op` so the
+    /// components are interpreted in the right space.
+    fill_space: Option<String>,
+    /// Most recent fill-color operator (`rg`/`g`/`k`/`sc`/`scn`), if any.
+    /// Glyph color comes from the fill state, so losing this across region
+    /// boundaries turns colored text black (`chars_hash`-only sweep diffs).
+    fill_op: Option<Operator>,
 }
 
 /// Lightweight forward scan that tracks graphics state across the full content stream.
 ///
-/// Scans the stream recognizing only `q`, `Q`, `cm`, and `Tf` operators, skipping
-/// all path, color, and text operators. Records the accumulated CTM and font state
-/// at each position in `text_positions`.
+/// Scans the stream recognizing only `q`, `Q`, `cm`, `Tf`, and the persistent
+/// text-state operators (`Tc`, `Tw`, `Tz`, `TL`, `Ts`, `Tr`, plus `TD`'s
+/// implicit leading), skipping all path, color, and text-showing operators.
+/// Records the accumulated CTM, font, and text state at each position in
+/// `text_positions`.
 ///
 /// This is much cheaper than full parsing. Numeric operands are tracked in a
 /// rolling buffer so `cm` operands are always available when the operator is
@@ -631,8 +679,18 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
     // in ctm_stack to avoid String cloning on every q/Q.
     let mut font_table: Vec<(String, f32)> = Vec::new();
     let mut current_font_idx: Option<usize> = None;
-    let mut ctm_stack: Vec<(Matrix, Option<usize>)> = Vec::with_capacity(32);
+    #[allow(clippy::type_complexity)]
+    let mut ctm_stack: Vec<(
+        Matrix,
+        Option<usize>,
+        PrescanTextState,
+        Option<String>,
+        Option<Operator>,
+    )> = Vec::with_capacity(32);
     let mut ctm = Matrix::identity();
+    let mut text_state = PrescanTextState::default();
+    let mut fill_space: Option<String> = None;
+    let mut fill_op: Option<Operator> = None;
 
     // Rolling buffer of recent numeric operands (for cm's 6 floats)
     let mut num_buf: [f32; 6] = [0.0; 6];
@@ -652,6 +710,9 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
         .map(|_| PrescanState {
             ctm: (0.0, 0.0, 0.0, 0.0, 0.0, 0.0),
             font: None,
+            text: PrescanTextState::default(),
+            fill_space: None,
+            fill_op: None,
         })
         .collect();
 
@@ -665,6 +726,9 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
             results[orig_idx] = PrescanState {
                 ctm: (ctm.a, ctm.b, ctm.c, ctm.d, ctm.e, ctm.f),
                 font: current_font_idx.map(|idx| font_table[idx].clone()),
+                text: text_state,
+                fill_space: fill_space.clone(),
+                fill_op: fill_op.clone(),
             };
             next_tp_idx += 1;
         }
@@ -752,13 +816,24 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                     continue;
                 },
                 b"q" => {
-                    ctm_stack.push((ctm, current_font_idx));
+                    ctm_stack.push((
+                        ctm,
+                        current_font_idx,
+                        text_state,
+                        fill_space.clone(),
+                        fill_op.clone(),
+                    ));
                     num_count = 0;
                 },
                 b"Q" => {
-                    if let Some((saved_ctm, saved_font_idx)) = ctm_stack.pop() {
+                    if let Some((saved_ctm, saved_font_idx, saved_text, saved_space, saved_fill)) =
+                        ctm_stack.pop()
+                    {
                         ctm = saved_ctm;
                         current_font_idx = saved_font_idx;
+                        text_state = saved_text;
+                        fill_space = saved_space;
+                        fill_op = saved_fill;
                     }
                     num_count = 0;
                 },
@@ -789,6 +864,104 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
                     }
                     num_count = 0;
                     last_name = None;
+                },
+                b"Tc" => {
+                    if num_count >= 1 {
+                        text_state.char_spacing = num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"Tw" => {
+                    if num_count >= 1 {
+                        text_state.word_spacing = num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"Tz" => {
+                    if num_count >= 1 {
+                        text_state.horiz_scaling = num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"TL" => {
+                    if num_count >= 1 {
+                        text_state.leading = num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"Ts" => {
+                    if num_count >= 1 {
+                        text_state.rise = num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"Tr" => {
+                    if num_count >= 1 {
+                        text_state.render_mode = num_buf[num_count - 1] as u8;
+                    }
+                    num_count = 0;
+                },
+                b"TD" => {
+                    // `tx ty TD` also sets the leading to -ty (§9.4.2).
+                    if num_count >= 2 {
+                        text_state.leading = -num_buf[num_count - 1];
+                    }
+                    num_count = 0;
+                },
+                b"rg" => {
+                    if num_count >= 3 {
+                        fill_op = Some(Operator::SetFillRgb {
+                            r: num_buf[num_count - 3],
+                            g: num_buf[num_count - 2],
+                            b: num_buf[num_count - 1],
+                        });
+                    }
+                    num_count = 0;
+                },
+                b"g" => {
+                    if num_count >= 1 {
+                        fill_op = Some(Operator::SetFillGray {
+                            gray: num_buf[num_count - 1],
+                        });
+                    }
+                    num_count = 0;
+                },
+                b"k" => {
+                    if num_count >= 4 {
+                        fill_op = Some(Operator::SetFillCmyk {
+                            c: num_buf[num_count - 4],
+                            m: num_buf[num_count - 3],
+                            y: num_buf[num_count - 2],
+                            k: num_buf[num_count - 1],
+                        });
+                    }
+                    num_count = 0;
+                },
+                b"cs" => {
+                    // `cs` resets the fill color to the space's initial value,
+                    // so any earlier color operator no longer applies.
+                    if let Some(name) = last_name.take() {
+                        fill_space = Some(name);
+                        fill_op = None;
+                    }
+                    num_count = 0;
+                },
+                b"sc" => {
+                    if num_count >= 1 {
+                        fill_op = Some(Operator::SetFillColor {
+                            components: num_buf[..num_count.min(4)].to_vec(),
+                        });
+                    }
+                    num_count = 0;
+                },
+                b"scn" => {
+                    if num_count >= 1 || last_name.is_some() {
+                        fill_op = Some(Operator::SetFillColorN {
+                            components: num_buf[..num_count.min(4)].to_vec(),
+                            name: last_name.take().map(Box::new),
+                        });
+                    }
+                    num_count = 0;
                 },
                 _ => {
                     num_count = 0;
@@ -879,6 +1052,9 @@ fn forward_scan_ctm(data: &[u8], text_positions: &[usize]) -> Option<Vec<Prescan
         results[orig_idx] = PrescanState {
             ctm: (ctm.a, ctm.b, ctm.c, ctm.d, ctm.e, ctm.f),
             font: current_font_idx.map(|idx| font_table[idx].clone()),
+            text: text_state,
+            fill_space: fill_space.clone(),
+            fill_op: fill_op.clone(),
         };
         next_tp_idx += 1;
     }
@@ -1272,6 +1448,16 @@ where
                         let (a, b, c, d, e, f) = state.ctm;
                         handler(Operator::SaveState)?;
                         handler(Operator::Cm { a, b, c, d, e, f })?;
+                        // Re-inject inherited fill color: glyph color reads the
+                        // fill state, which the region wrap would otherwise
+                        // reset to black. Space first so components are
+                        // interpreted in the right color space.
+                        if let Some(ref name) = state.fill_space {
+                            handler(Operator::SetFillColorSpace { name: name.clone() })?;
+                        }
+                        if let Some(ref op) = state.fill_op {
+                            handler(op.clone())?;
+                        }
                         // Inject font state if the forward scan tracked one.
                         // This handles BT blocks that inherit Tf from a prior
                         // scope instead of setting their own.
@@ -1279,6 +1465,41 @@ where
                             handler(Operator::Tf {
                                 font: font_name.clone(),
                                 size: font_size,
+                            })?;
+                        }
+                        // Re-inject inherited text state the same way. These
+                        // parameters persist across BT/ET, so a region may rely
+                        // on a value set by an earlier region — which the
+                        // SaveState/RestoreState wrap would otherwise reset.
+                        // Only non-default values are injected, keeping the
+                        // operator stream unchanged for the common case.
+                        let ts = &state.text;
+                        if ts.char_spacing != 0.0 {
+                            handler(Operator::Tc {
+                                char_space: ts.char_spacing,
+                            })?;
+                        }
+                        if ts.word_spacing != 0.0 {
+                            handler(Operator::Tw {
+                                word_space: ts.word_spacing,
+                            })?;
+                        }
+                        if ts.horiz_scaling != 100.0 {
+                            handler(Operator::Tz {
+                                scale: ts.horiz_scaling,
+                            })?;
+                        }
+                        if ts.leading != 0.0 {
+                            handler(Operator::TL {
+                                leading: ts.leading,
+                            })?;
+                        }
+                        if ts.rise != 0.0 {
+                            handler(Operator::Ts { rise: ts.rise })?;
+                        }
+                        if ts.render_mode != 0 {
+                            handler(Operator::Tr {
+                                render: ts.render_mode,
                             })?;
                         }
                         parse_region_text_only(&data[*start..*end], &mut handler)?;
@@ -5198,6 +5419,162 @@ mod tests {
             tf_ops.len() >= 2,
             "expected Tf for both BT blocks (explicit + inherited), got {}",
             tf_ops.len()
+        );
+    }
+
+    /// Build a stream with two BT regions separated by >256KB of path filler,
+    /// forcing the prescan's forward-scan path. `first_region` is the body of
+    /// the first BT block (state setters + show ops).
+    fn two_region_prescan_stream(first_region: &[u8]) -> Vec<u8> {
+        let mut cs = Vec::new();
+        cs.extend_from_slice(b"BT ");
+        cs.extend_from_slice(first_region);
+        cs.extend_from_slice(b" ET\n");
+        for i in 0..14000u32 {
+            let line = format!(
+                "{}.0 {}.0 m {}.0 {}.0 l n\n",
+                i % 500,
+                (i * 7) % 500,
+                (i * 3) % 500,
+                (i * 11) % 500
+            );
+            cs.extend_from_slice(line.as_bytes());
+        }
+        assert!(cs.len() > 256 * 1024, "stream must exceed 256KB prescan threshold");
+        cs.extend_from_slice(b"BT (B) Tj T* (C) Tj ET\n");
+        cs
+    }
+
+    /// The operators injected for the LAST region: everything between the
+    /// final SaveState and the final BeginText.
+    fn last_region_injected(ops: &[Operator]) -> &[Operator] {
+        let last_bt = ops
+            .iter()
+            .rposition(|op| matches!(op, Operator::BeginText))
+            .expect("no BeginText");
+        let save = ops[..last_bt]
+            .iter()
+            .rposition(|op| matches!(op, Operator::SaveState))
+            .expect("no SaveState before last BeginText");
+        &ops[save..last_bt]
+    }
+
+    #[test]
+    fn test_prescan_injects_inherited_text_state() {
+        // Tc, Tw, and the leading persist across BT/ET (ISO 32000-1 §9.3.1;
+        // only Tm resets at BT), so the second region legitimately runs with
+        // the first region's values. The prescan path parses each region in
+        // an isolated SaveState/RestoreState wrap; without reconstruction the
+        // second region's T* would step by zero leading and every advance
+        // would lose Tc — glyphs land on the wrong line, slightly compressed
+        // (govdocs_040_040921.pdf p22, govdocs_055_055555.pdf p23).
+        let cs = two_region_prescan_stream(b"/F1 10 Tf 0.5 Tc 0.25 Tw 0 -12 TD (A) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+
+        let injected = last_region_injected(&ops);
+        assert!(
+            injected
+                .iter()
+                .any(|op| matches!(op, Operator::Tc { char_space } if (*char_space - 0.5).abs() < 1e-6)),
+            "inherited Tc not injected: {:?}",
+            injected
+        );
+        assert!(
+            injected
+                .iter()
+                .any(|op| matches!(op, Operator::Tw { word_space } if (*word_space - 0.25).abs() < 1e-6)),
+            "inherited Tw not injected: {:?}",
+            injected
+        );
+        assert!(
+            injected
+                .iter()
+                .any(|op| matches!(op, Operator::TL { leading } if (*leading - 12.0).abs() < 1e-6)),
+            "leading set via TD not injected: {:?}",
+            injected
+        );
+    }
+
+    #[test]
+    fn test_prescan_does_not_inject_default_text_state() {
+        // A first region that never touches text state must not cause any
+        // text-state operator to be synthesized for the second region.
+        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+
+        let injected = last_region_injected(&ops);
+        assert!(
+            !injected.iter().any(|op| matches!(
+                op,
+                Operator::Tc { .. }
+                    | Operator::Tw { .. }
+                    | Operator::Tz { .. }
+                    | Operator::TL { .. }
+                    | Operator::Ts { .. }
+                    | Operator::Tr { .. }
+            )),
+            "default text state must not be injected: {:?}",
+            injected
+        );
+    }
+
+    #[test]
+    fn test_prescan_injects_inherited_fill_color() {
+        // Glyph color reads the fill state, which persists across BT/ET like
+        // the rest of the graphics state. A region relying on a color set
+        // before or in an earlier region must get it re-injected, or colored
+        // text silently turns black on the prescan path (the old vec parser
+        // preserved color operators, so this shows up as chars_hash-only
+        // sweep diffs).
+        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET 0.2 0.4 0.6 rg BT (X) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+
+        let injected = last_region_injected(&ops);
+        assert!(
+            injected.iter().any(|op| matches!(
+                op,
+                Operator::SetFillRgb { r, g, b }
+                    if (*r - 0.2).abs() < 1e-6 && (*g - 0.4).abs() < 1e-6 && (*b - 0.6).abs() < 1e-6
+            )),
+            "inherited fill color not injected: {:?}",
+            injected
+        );
+    }
+
+    #[test]
+    fn test_prescan_text_state_respects_q_restore() {
+        // Text state set inside a q/Q scope is restored at Q, so it must not
+        // be injected into a region that begins after the Q.
+        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET q 3 Tc BT (X) Tj ET Q BT (Y) Tj");
+        let mut ops = Vec::new();
+        parse_and_execute_text_only(&cs, |op| {
+            ops.push(op);
+            Ok(())
+        })
+        .unwrap();
+
+        let injected = last_region_injected(&ops);
+        assert!(
+            !injected
+                .iter()
+                .any(|op| matches!(op, Operator::Tc { .. })),
+            "Tc from a closed q/Q scope must not leak into a later region: {:?}",
+            injected
         );
     }
 
--- a/src/content/parser.rs
+++ b/src/content/parser.rs
@@ -5581,9 +5581,9 @@ mod tests {
         .unwrap();
         let injected = last_region_injected(&ops);
         assert!(
-            !injected.iter().any(
-                |op| matches!(op, Operator::SetFillColorN { name: Some(_), .. })
-            ),
+            !injected
+                .iter()
+                .any(|op| matches!(op, Operator::SetFillColorN { name: Some(_), .. })),
             "stale name fabricated a pattern: {:?}",
             injected
         );
@@ -5593,7 +5593,8 @@ mod tests {
     fn test_prescan_scn_with_too_many_operands_injects_nothing() {
         // A 6-component scn (DeviceN) exceeds what the rolling buffer can
         // faithfully replay; a truncated color would be plausible and wrong.
-        let cs = two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET 0.1 0.2 0.3 0.4 0.5 0.6 scn BT (X) Tj");
+        let cs =
+            two_region_prescan_stream(b"/F1 10 Tf (A) Tj ET 0.1 0.2 0.3 0.4 0.5 0.6 scn BT (X) Tj");
         let mut ops = Vec::new();
         parse_and_execute_text_only(&cs, |op| {
             ops.push(op);
--- a/src/converters/docx_layout.rs
+++ b/src/converters/docx_layout.rs
@@ -984,11 +984,7 @@ mod heading_level_tests {
             );
         }
         for s in spans.iter().skip(3) {
-            assert_eq!(
-                s.heading_level, None,
-                "body-size span must not be a heading: {:?}",
-                s.text
-            );
+            assert_eq!(s.heading_level, None, "body-size span must not be a heading: {:?}", s.text);
         }
     }
 
--- a/src/converters/docx_layout.rs
+++ b/src/converters/docx_layout.rs
@@ -739,9 +739,14 @@ fn annotate_heading_levels(spans: &mut [crate::layout::text_block::TextSpan]) {
         let key = (s.font_size * 10.0).round() as i32;
         *buckets.entry(key).or_default() += 1;
     }
+    // Explicit tie-break, not `max_by_key`: it returns the LAST maximal
+    // element, and over a `HashMap` that is the per-process-randomized
+    // iteration order, so a page whose two font sizes carry equal span counts
+    // picked a different modal body size per process — which flips the
+    // heading/body ratio test below. Ties resolve to the smaller size.
     let body_size_tenth = buckets
         .iter()
-        .max_by_key(|(_, c)| *c)
+        .max_by(|(a_key, a_c), (b_key, b_c)| a_c.cmp(b_c).then_with(|| b_key.cmp(a_key)))
         .map(|(k, _)| *k)
         .unwrap_or(120);
     let body = (body_size_tenth as f32) / 10.0;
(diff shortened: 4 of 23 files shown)


yfedoseev:
@tobocop2 please read and follow Contributing.md. All recent changes that were merged followed it: one issue - one PR. If it's hard for you to follow, please, just report issues then and I am happy to work on changes.

tobocop2:
> @tobocop2 please read and follow Contributing.md. All recent changes that were merged followed it: one issue - one PR. If it's hard for you to follow, please, just report issues then and I am happy to work on changes.

It's not a difficulty thing at all. I assumed that this rule was a guideline and not a blocker

The rule was not followed when I made my [first contribution here either](https://github.com/yfedoseev/pdf_oxide/pull/814) and since then I had assumed this pattern was acceptable here

In my mind, these issues all correspond to one feature and make sense to be in one branch, but I will happily split everything up back to be how it was when I started this.

tobocop2:
Closing this in favor of multiple PR's, 1 issue each:

| PR | issue | what it fixes |
|---|---|---|
| #982 | #806 | sideways `Tm` run continuation — "brown foxjumps over" |
| #987 | #983 | rotated line grouping (458 words no longer 458 lines) + the O(runs²) scan it introduced |
| #1012 | #984 | the rotated reading frame applied on every text surface, not just `extract_text` |
| #989 | #806 | `page_bbox()` — where a sideways run sits on the page |
| #1013 | #991 | four rasterizer sites that dropped glyphs silently |
| #990 | #981 | table extraction deleting words the grid does not re-emit |
| #1008 | #1004 | hash iteration order reaching extraction, redaction and table output |
| #1009 | #1005 | text state and fill colour reset at region boundaries on pages over 256 KB |
| #1010 | #1006 | `extract_chars` and `extract_spans` parsed by two different parsers |
| #1011 | #1007 | the duplicated glyph decoder (merges after #1013) |

The split is content-exact: merging all ten onto `main` reproduces this branch's tree byte for byte (`git rev-parse` tree `2180d3ff…` on both sides, zero merge conflicts). Each PR carries its own reproducer, corpus evidence, and standalone green suite; evidence tables are in each description.

tobocop2:
@yfedoseev pinging here since I don't have a better way to reach you.

All of my open PR's are now ready for review just in time for the weekend :)

Thanks again for this awesome library and it's been a blast to work with. Looking forward to your reviews.

@Goldziher things are officially now very stable across many more cases, and all of my use cases that I know of are largely solved. this all came out of driving xberg with lilbee.

For some unsolicited context, I talk to car manuals a lot. In fact, this is my primary usecase when talking to PDFs. My 2011 Ford Crown Victoria manual was the catalyst that drove this all. I could not ask it what my oil capacity was anymore (my favorite test question to ask) and this is what drove this entire effort and it's been largely successful and it is faster than pdfium by far which is why I want to express my gratitude to this project again.

We don't hit https://github.com/yfedoseev/pdf_oxide/issues/994 in xberg so I'm leaving it to @yfedoseev.

Also, i haven't gotten to https://github.com/yfedoseev/pdf_oxide/issues/1003 yet and I'll try to take it on if I have some time this weekend.

Here are my PR's, I followed the 1 issue to 1 PR rule.

| PR | Issue | What it does |
|---|---|---|
| #982 | #806 | Stops sideways text fusing across line breaks |
| #987 | #983 | Groups rotated runs into whole lines |
| #989 | #806 | Reports the correct bounding box for sideways runs |
| #990 | #981 | Keeps words the table grid does not re-emit |
| #996 | #995 | Twelve extraction defects across text, tables, images and forms |
| #998 | #997 | Handles malformed geometry, and makes two read paths deterministic |
| #1000 | #999 | Decodes a sampled shading stream once, not per grid point |
| #1002 | #1001 | Skips draws whose device coordinates exceed f32 pixel precision |
| #1008 | #1004 | Makes extraction, redaction and table output order-independent |
| #1009 | #1005 | Restores text state in the prescan for streams over 256 KB |
| #1010 | #1006 | Parses chars and spans with the same parser |
| #1011 | #1007 | Unifies the duplicated glyph decoder |
| #1012 | #984 | Reads rotated pages in one frame on every text surface |
| #1013 | #991 | Reports glyphs that paint nothing instead of dropping them |
| #1020 | #1019 | Skips zero-dimension image blits instead of panicking |
| #1021 | #1017 | Makes inline-image extraction deterministic for duplicate dictionary keys |
| #1025 | #1018 | Reuses the merger's word boundaries when building cell text |

yfedoseev:
@tobocop2 please check CI CD, I see there are issues with formating. I can start reviewing these issues not early than mid of the next week

yfedoseev:
@tobocop2 — I've gone through all 17 PRs now: code review, our test corpus, and a golden-reference dataset with known-correct output, plus the lower-level extraction surfaces. Individual comments are going up on each PR now.

Headline: 4 of 17 are clean and merging today. The other 13 have real value in them — several are genuinely better-diagnosed than I'd have found myself — but as submitted, most are blocked on things like a missing DCO sign-off, bundling several fixes into one PR, or an open correctness question I need resolved before merge.

I hold my own commits to the same rules I'm asking you to follow: one logical change per PR, signed off, one thing reviewed at a time. I don't think that's an unreasonable bar, and honestly, the quality of your root-causing tells me you don't need me to explain any of this — you found and explained real bugs in this codebase faster than most contributors would.

It looks like our process might have some friction for a batch this size — if it's tough to get a full PR through everything on our end, you can always just report the issue and I'll fix it myself. #809 already worked out that way and it was genuinely no trouble.

Always a pleasure working with people who care about quality as much as you and @Goldziher do.

tobocop2:
> @tobocop2 — I've gone through all 17 PRs now: code review, our test corpus, and a golden-reference dataset with known-correct output, plus the lower-level extraction surfaces. Individual comments are going up on each PR now.
> 
> Headline: 4 of 17 are clean and merging today. The other 13 have real value in them — several are genuinely better-diagnosed than I'd have found myself — but as submitted, most are blocked on things like a missing DCO sign-off, bundling several fixes into one PR, or an open correctness question I need resolved before merge.
> 
> I hold my own commits to the same rules I'm asking you to follow: one logical change per PR, signed off, one thing reviewed at a time. I don't think that's an unreasonable bar, and honestly, the quality of your root-causing tells me you don't need me to explain any of this — you found and explained real bugs in this codebase faster than most contributors would.
> 
> It looks like our process might have some friction for a batch this size — if it's tough to get a full PR through everything on our end, you can always just report the issue and I'll fix it myself. #809 already worked out that way and it was genuinely no trouble.
> 
> Always a pleasure working with people who care about quality as much as you and @Goldziher do.

Thank you! I'll aim to address everything tonight.

tobocop2:
@yfedoseev everything is addressed. Since all of 27 of my CI jobs were canceled, I will leave triggering CI to you as you see fit. 

Since CI was green before everything was split up, I'm confident things will be green when the time comes. 

I recommend hardening some of the infrastructure in the future so that only maintainers can invoke the full CI sweep given how intensive it is. 

Thanks again for the reviews and feedback!

yfedoseev:
@tobocop2 great point, and thank you for it. I've hardened the CI approval policy so it now requires my sign-off before any external contributor's workflow run starts, not just first-time ones — exactly what you suggested. That should keep the pipeline from getting overwhelmed on batches like this going forward.

Thanks again for the work here and for thinking about the project's health, not just your own PRs. Appreciate it.

PR #996, opened 2026-08-04
fix(extract): correct twelve defects in text, tables, images and forms

### LINKED ISSUE #995: [Bug]: extraction-correctness bugs on the text/table/image path (each verified against pdftotext / PyMuPDF / pdfminer)
*A focused pass over the code reached by `from_bytes` → `to_markdown` / `to_plain_text`, table extraction, image extraction, forms, and link annotations. Companion to #994, which covered the non-extraction subsystems.*

> [!NOTE]
> One issue as an epic, sorry for the format — these are all independent extraction bugs, split off whatever's useful. Every finding is wrong or lost **text / table / image** output on a **valid** PDF.

> [!IMPORTANT]
> **Runnable reproducers for all 12 findings are in this gist: https://gist.github.com/tobocop2/3c1388fbdb338c62efca164d4db872e2** — the integration tests plus the PDF generators that build each crafted fixture, keyed to the table below.

## How each was verified
Each finding was reproduced **differentially**: the same PDF is extracted with pdf_oxide and with a reference extractor (`pdftotext`/poppler, PyMuPDF, and/or pdfminer/pypdf), and reported only where **pdf_oxide produces the wrong output and a reference produces the right one** — or, for crafted inputs, where pdf_oxide disagrees with the file's known-correct content. Where every extractor agreed, or the file was undecodable, it was dropped. Run against **v0.3.77**.

## Comparison table
Read each row as: given the **Trigger**, pdf_oxide outputs the **pdf_oxide** column, but the correct output (per the reference extractors and the PDF spec) is the **Correct** column.

| # | Sev | Location | Trigger | pdf_oxide output | Correct output |
|---|---|---|---|---|---|
| 1 | High | `src/fonts/cid_mappings/adobe_japan1.rs:2333` | Japanese 青 shown via an Adobe-Japan1 CIDFont (CID 2664), no ToUnicode | `⻘` (U+2ED8, a radical *presentation* form) | `青` (U+9752, the real ideograph) |
| 2 | Medium | `src/fonts/font_dict.rs:5154` | any radical-form char in U+2E80–2EFF (e.g. `⻘`) passing the "normalize radicals" guard | `⻘` unchanged — the guard silently no-ops | `青` (U+9752) |
| 3 | Low | `src/extractors/text.rs:8427` | Type0/Identity-H run `50 Tw <0020 0021> Tj` (the 2-byte code 0x0020 is a real glyph, not a space) | span right edge `x1 = 148` (word-spacing wrongly applied) | `x1 = 98` — matches the `TJ` form, PyMuPDF, and ISO 32000-1 §9.3.3 |
| 4 | High | `src/pipeline/converters/mod.rs:359` | 20pt heading "Revenue" then 12pt paragraph "2024 was a record year." (two baselines) | `## Revenue 2024 was a record year.` (heading swallows the paragraph) | `Revenue` heading + separate `2024 was a record year.` paragraph |
| 5 | Medium | `src/pipeline/converters/markdown.rs:350` | three distinct 20pt headings "Sales", "Marketing", "Engineering" | `## Sales Marketing Engineering` (one heading) | three separate headings |
| 6 | Medium | `src/pipeline/converters/markdown.rs:610` | bold text "A gitHub repo" (single upper word + CamelCase word + lower word) | `**AgitHubrepo**` (both spaces deleted) | `A gitHub repo` |
| 7 | High | `src/structure/table_extractor.rs:829` | a table cell containing two runs "Hello" then "World" under one MCID | cell = `Hello`; "World" is dropped out of the table entirely | cell keeps `Hello World` |
| 8 | High | `src/extractors/images.rs:954` | a 1-bit DeviceGray 8×8 image | 8 *packed* bytes labelled 8-bit `Grayscale` (buffer too short to be a valid 8×8 plane) | 64 unpacked samples `[0,255,0,255,…]` |
| 9 | Medium | `src/extractors/images.rs:942` | an 8-bit gray image with `/Decode [1 0]` (tonal invert) | raw samples `[0,1,2,3,…]` — `/Decode` ignored | inverted `[255,254,253,…]` |
| 10 | High | `src/encryption/algorithms.rs:242` | an AES-256 **R5** encrypted PDF, empty user password | `authenticate()` returns true but `extract_text` is `""` (streams decrypt to noise) | the real text (the R6 path handles the identical file correctly) |
| 11 | Medium | `src/annotations.rs:653` | an internal link to page 2, whose page object is number 6 | destination page index `6` (the raw PDF object number) | page index `1` (0-based position in the page tree) |
| 12 | Medium | `src/extractors/forms.rs:548` | a radio/checkbox with `/V /No` selected (a legitimate on-state) | `Boolean(false)` — indistinguishable from unset/off | the export value `No` |

<sub>References: `pdftotext` (poppler 25.06), PyMuPDF (fitz), pdfminer/pypdf. "Correct output" is the value all references agree on, matching the ISO&nbsp;32000-1 spec where cited.</sub>

## Details & reproducers
One entry per row above — the root cause and the exact command that reproduces the comparison.

### 1. High — `src/fonts/cid_mappings/adobe_japan1.rs:2333`
The Adobe-Japan1 CID→Unicode table maps common ideograph CIDs to CJK Radicals Supplement codepoints (U+2E80–2EFF) instead of the unified ideographs (CID 2664→U+2ED8 instead of U+9752 青; CID 2666→U+2EEB instead of U+6589 斉). Predefined CJK CIDFonts without a ToUnicode map therefore extract the wrong character. Cross-checked against pdfminer's own `Adobe-Japan1` map: `cid2unichr[2664]='青'`.

*Entry point:* to_plain_text / to_markdown  
*Reproduce:* `cargo test --test zzz_xaudit_fonts-glyph-unicode japan1_cidfont_radical_forms -- --nocapture`

### 2. Medium — `src/fonts/font_dict.rs:5154`
`normalize_cjk_radical_forms` is documented (lines 5140–5148) to carry radical presentation forms in *both* the Supplement (U+2E80–2EFF) and Kangxi (U+2F00–2FDF) blocks to ideographs via NFKC — but NFKC has decompositions for only **2 of the 128** Supplement-block codepoints, so the guard does nothing for that entire block and merely masks (rather than fixes) the Kangxi-block table entries. This is the layer that *should* have caught finding&nbsp;1 and doesn't.

*Entry point:* to_plain_text / to_markdown (runs on every char)  
*Reproduce:* `cargo test --test zzz_xaudit_fonts-glyph-unicode japan1_cidfont_radical_forms -- --nocapture`

### 3. Low — `src/extractors/text.rs:8427`
The plain-`Tj` path applies word spacing (Tw) to a 2-byte code whose CID is 0x20 by testing `char_code == 32` with no byte-width guard; the sibling `TJ` helper guards correctly with `nbytes == 1 && cid == 32`, so the two paths disagree on identical content. Note: the *text* ("AB") is unaffected — only the reported span width — because any later reposition resets the text matrix.

*Entry point:* extract_spans → TextSpan.bbox  
*Reproduce:* `cargo test --test zzz_xaudit_content-interp-spans -- --nocapture`

### 4. High — `src/pipeline/converters/mod.rs:359`
`merge_key_value_pairs` treats a font-size heading as a key/label and glues the following short paragraph onto the same line whenever that paragraph starts with a digit, `$`, `(`, or `-`, producing a malformed heading and destroying the paragraph break.

*Entry point:* to_markdown / to_plain_text  
*Reproduce:* `cargo test --test zzz_xaudit_markdown-readingorder heading_absorbs_following_numeric_paragraph -- --nocapture`

### 5. Medium — `src/pipeline/converters/markdown.rs:350`
The `three_plus_short` policy of `merge_consecutive_same_level_headings` fuses any run of 3+ consecutive same-level headings that are each ≤2 words into a single heading line, collapsing genuinely distinct section headers.

*Entry point:* to_markdown  
*Reproduce:* `cargo test --test zzz_xaudit_markdown-readingorder three_short_headings_wrongly_merged -- --nocapture`

### 6. Medium — `src/pipeline/converters/markdown.rs:610`
`coalesce_camelcase_bold_fragments` deletes the spaces inside any bold run shaped `**A wordCap word**`, silently concatenating legitimate bold prose into one token.

*Entry point:* to_markdown  
*Reproduce:* `cargo test --test zzz_xaudit_markdown-readingorder camelcase_bold_eats_spaces -- --nocapture`

### 7. High — `src/structure/table_extractor.rs:829`
`extract_cell`'s inner-loop `break` keeps only the *first* text block matching each cell's MCID; every later span sharing that TD's MCID is silently dropped from the `TableCell`, contradicting the line-690 comment ("collect all text blocks that match these MCIDs"). Multi-run cells lose everything after their first run.

*Entry point:* to_markdown with table extraction  
*Reproduce:* `cargo test --test zzz_xaudit_tables crafted_tagged_pdf_to_markdown_cell_loses_world -- --nocapture`

### 8. High — `src/extractors/images.rs:954`
Sub-8-bit non-CCITT images (1/2/4-bpc DeviceGray, and any sub-byte non-DeviceGray) fall through the `else { decoded_data }` arm and are returned as the packed sub-byte stream while labelled `PixelFormat::Grayscale`/`RGB`, whose `bytes_per_pixel` implies unpacked 8-bit samples. The length no longer matches `w*h*bpp`, so a consumer that builds an image from the buffer rejects it and the image is lost. Both PyMuPDF and poppler return the full 64-sample unpacked plane.

*Entry point:* extract_images  
*Reproduce:* `cargo test --test zzz_xaudit_image-extract-ocr --features rendering one_bit_devicegray_emits_packed_buffer -- --nocapture`

### 9. Medium — `src/extractors/images.rs:942`
A non-default `/Decode` array is honored only for 1-bit DeviceGray and CCITT. For BitsPerComponent ≥ 8 (and non-DeviceGray) the `else` arm returns the samples verbatim, so an inverted-`/Decode` image is extracted with tonally-wrong pixels. ISO 32000-1 §8.9.5.2 requires samples to be mapped through `/Decode` before use.

*Entry point:* extract_images  
*Reproduce:* `cargo test --test zzz_xaudit_image-extract-ocr --features rendering decode_array_ignored_for_8bit_gray -- --nocapture`

### 10. High — `src/encryption/algorithms.rs:242`
For AES-256 **R5** the file key is taken as `SHA256(password‖key_salt)` and returned directly instead of AES-256-CBC-decrypting the `/UE` entry with that intermediate key. R6 (lines 233–241) unwraps `/UE` correctly; R5 (lines 243–248) does not — so authentication succeeds but every stream decrypts to garbage and extraction silently yields empty output.

*Entry point:* from_bytes → extract_text / to_markdown  
*Reproduce:* `cargo test --test zzz_xaudit_forms-xfa-annots-decrypt finding1_r5_aes256_decrypt -- --nocapture`

### 11. Medium — `src/annotations.rs:653`
`parse_destination` for an explicit `[pageRef /XYZ …]` destination uses the raw object number as the page index instead of resolving the reference to its position in the page tree, so link targets point at the wrong page (out of range when the object number exceeds the page count). The code comment says "For now, just use object ID as approximation."

*Entry point:* get_annotations → link destination  
*Reproduce:* `cargo test --test zzz_xaudit_forms-xfa-annots-decrypt finding2_link_goto_page_index -- --nocapture`

### 12. Medium — `src/extractors/forms.rs:548`
`parse_field_value` maps a Button field whose `/V` name is "No" or "Off" to `Boolean(false)`, collapsing an actively-selected export value into a boolean; a selected "No" becomes indistinguishable from an unchecked field even though "No" is a real on-state.

*Entry point:* form field extraction → FieldValue  
*Reproduce:* `cargo test --test zzz_xaudit_forms-xfa-annots-decrypt finding3_button_no_export_value -- --nocapture`

---

<sub>Reproducers (integration tests + the PDF generators for each crafted fixture): https://gist.github.com/tobocop2/3c1388fbdb338c62efca164d4db872e2</sub>

tobocop2:
> [!IMPORTANT]
> **Merge this before #989.** Both change `table_extractor.rs`, so #989 needs a small rebase afterwards.

## Linked issue

Closes #995

## What and why

Twelve extraction-correctness defects between `from_bytes` and text, table, image, form and annotation output. Each fix carries a regression test that fails without it. Every test builds a synthetic PDF in test code.

### Text and fonts

**`Tw` reached 2-byte CIDs.** A CID whose code was 32 picked up word spacing, so `Tj` and `TJ` disagreed on identical bytes. `Tw` is now single-byte only (§9.3.3).

**CJK text came back as radicals**, for example 青 as ⻘. The old table build inverted code→CID CMaps to produce Adobe-Japan1, GB1 and CNS1, which collided 439 entries. Each table is now built from its own collection's mapping.

**The guard meant to catch that used NFKC**, which decomposes part of the radical block. It now uses `Equivalent_Unified_Ideograph`.

### Tables and markdown

**A tagged cell kept only its first text run.** It now keeps every block its marked-content sequence owns. Those blocks are ordered along the cell's own writing axis.

**A heading absorbed the paragraph after it**, producing `## Revenue 2024 was a record year.`

**Three short headings fused into one.** Line breaks now come from column geometry rather than string shape.

**Bold prose lost its spaces** (`**AgitHubrepo**`). That pass is deleted. Span assembly now prevents the tear it used to compensate for.

### Images

**1/2/4-bpc samples were returned packed** but labelled 8-bit `Grayscale`. They now unpack to one byte per component (§8.9.5.2).

**`/Decode` was honored only at 1 bit and for CCITT.** It now applies at 1/2/4/8/16 bpc. The image also tells consumers when `/Decode` has been folded in, so colour-key `/Mask` and separation plates still route.

**A stream shorter than its declared dimensions** is now padded rather than refused. The padding is capped at 256 MiB.

### Encryption, forms and links

**AES-256 R5 files authenticated, then decrypted every stream to noise.** The code returned the intermediate hash as the file key. It now unwraps `/UE` or `/OE`.

**A selected `/No` button read as `Boolean(false)`**, indistinguishable from an unanswered one. Only `/Off` now means unselected (§12.7.4.2.3).

**Internal links reported object numbers as page numbers.** Explicit destinations now resolve through the page tree (§12.3.2.2).

### Not covered

`/Decode` remains ignored for DCT- and JPX-coded images. Those streams never reach this path.

The truncated-stream case above is a content-parser defect. It is filed as #1003 and not fixed here.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): `rendering`, `fips,icc`; no binding sources are touched.

**The third box is unticked because the run is not green end to end.** None of the red comes from this branch. Every failed job stopped in `Set up job` before a step ran, and the whole macOS matrix is `cancelled`.

These all pass: `Test (ubuntu-latest, stable)` and `(nightly)`, `Test (windows-latest, stable)` and `(beta)`, `Rendering Tests`, `Feature Tests`, `Feature powerset`, `FIPS CryptoProvider (aws-lc-rs)`, `Java FIPS` and `Code Coverage`.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents / ~470,000 pages. Sources are [govdocs1](https://digitalcorpora.org/corpora/files/govdocs1/) (14,903), [veraPDF](https://github.com/veraPDF/veraPDF-corpus) (2,907), [pdf.js](https://github.com/mozilla/pdf.js/tree/master/test/pdfs) (974) and [pdfium](https://pdfium.googlesource.com/pdfium/+/refs/heads/main/testing/resources) (331), plus curated and synthetic fixtures.

      scripts, or crashes).
      char-Levenshtein.

I did not diff against the latest release. The sweep ran against `main` only, and no unintended regression survived attribution.

**Diff summary vs `main`:**

| | |
|---|---|
| pages changed | 77,746 of 454,088 (17.1%) |
| documents changed | 8,882 of 14,909 |
| pages that lose a character | **0** |
| total extracted text | **+10.3%** |
| gate blockers | 8, all attributed |
| token-loss / duplication pages | 2,505 / 2,114, all `tables`-classed |
| pages changed outside any trigger class | 86 |

**Diff summary vs latest release:** not run separately. The sweep was against `main`, which sits at the v0.3.77 release commit.

**Token loss and duplication are bookkeeping.** Conservation is a per-surface multiset subtraction. Text that moves between surfaces therefore reads as loss in one place and gain in another.

All 2,505 loss pages and 2,114 duplication pages carry the `tables` class. Of those, 1,467 conserve their whole-page character multiset exactly and 1,790 gain characters. The image-buffer shape invariant improves on 3,582 pages and worsens on none.

|---|---|---|
| [014006.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/014.zip) (014.zip) | 4 | dropped run text returns to its cell |
| [008398.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/008.zip) (008.zip) | 6 | header cell completes: "Nearest Event" becomes "Nearest Event Number" |
| synthetic vertical-text fixture (Japanese) | 47, 111, 242 | radical forms emit as real ideographs; 422 near-empty lines assemble into vertical-reading text |
| [046432.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/046.zip) (046.zip) | 3 | rotated page: reading order corrected, character multiset identical |
| [057017.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/057.zip) (057.zip) | 6 | 51-char delta is entirely spurious space objects; emitted text byte-identical |
| synthetic image-decode fixtures (2) | n/a | declared and produced image facts agree; render output changes with the corrected decode |

These show that text is still present. They do not show that it sits in the right reading order within a table.

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: diagnosis, the fixes, the tests, the corpus sweep and this description.
      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

**This is not one logical change.** It is twelve fixes in one defect class on the extraction path. I bundled them deliberately, because I found and swept them together.

**`CHANGELOG.md` is untouched.** It is written per release, under a version heading, with a consolidated contributors block, so it reads as the maintainer's to write at release time.

**On semver:** `FieldValue` for `/No` buttons and `LinkAction` for a dangling `GoTo` each change an observed value, and `PdfImage` gains two accessors. No signature is removed, so `semver-checks` cannot see the behavioural half. Flagging it for your call.


yfedoseev:
This is a lot of real, individually good fixes bundled into one PR — and one of them (the markdown heading/value-line heuristic) has a real problem: on our test corpus it fragments a majority of the multi-line titles and headings it touches into pieces, which is the opposite of what it's meant to fix. The other findings in here (CJK radical mapping, the Tw fix, table cell multi-run, AES-256 R5, link destinations, button export values) all look solid independently.

Given the size of this, I think the 8-way split your own PR body already lays out is the right path — please open those as separate PRs, and hold off on the heading-heuristic one until we can look at the fragmentation issue together; the geometric approach there needs a different mechanism, not a threshold tweak.

tobocop2:
The eight-way split is up, one issue + one draft PR per fix, along the body's own layout:

| Fix | Issue | PR |
|---|---|---|
| Tw on 2-byte CIDs | #1058 | #1059 |
| CJK radicals (tables + guard) | #1060 | #1061 |
| Tagged table cells (multi-run, ordering) | #1062 | #1063 |
| Markdown overmerge + bold-prose spaces | #1064 | #1065 |
| Images (sub-byte unpack, /Decode, padding, sample-space flag) | #1066 | #1067 |
| AES-256 R5 /UE//OE unwrap | #1068 | #1069 |
| Button export values | #1070 | #1071 |
| Link destinations (page-tree resolution) | #1072 | #1073 |

The union of the eight branches on current main is byte-identical to this branch minus the held heading heuristic, merged with main — verified before splitting, so nothing was lost or invented in the split. The heading/value-line heuristic is in none of them; #1064 notes it stays open pending a different mechanism, per your review.

Each split PR's corpus section fills in with its own sweep verdict today (the sweeps are running). The tagged-cells PR (#1063) carries the `table_extractor.rs` surface, so the merge-order note with #989 transfers to it. I'll close this one once all the split PRs are verified by their corpus sweeps.

tobocop2:
All eight split PRs are swept and ready for review, each with its own verdict against a fresh base from current main: #1059, #1061, #1063, #1065, #1067, #1069, #1071, #1073. The union of the eight was verified byte-identical to this branch minus the held heading heuristic before the split, and the images split's own sweep caught and fixed one further defect (#1067, the JBIG2 raw-flag). The heading/value-line heuristic stays open per review and is in none of them. Closing this bundle in favour of the set.

tobocop2:
Byte-level accounting of the split, measured after the round closed: merge the eight split heads onto current `main` and compare against this branch merged with `main` minus the held heading heuristic. The two trees differ by exactly one hunk — the JBIG2 raw-sample-flag guard in `src/extractors/images.rs` (+13/−6), the defect the images split's own sweep found and #1067 fixes. Every other line is byte-identical.

PR #998, opened 2026-08-04
fix(render): handle malformed geometry and make two read paths deterministic

### LINKED ISSUE #997: [Bug]: malformed geometry panics the renderer, and two read paths are nondeterministic
### Bug Description

Seven defects on the read path — reached through `from_bytes`, `render_page` and the tagged-PDF structure tree.

**Five are crashes.** A malformed shading function, a negative image-mask dimension or a zero-width `/Mask` panics inside `render_page`, on a file the crate opened without complaint. A caller cannot defend against that with a `Result`.

**Two are nondeterminism.** Neither errors, neither logs, and both return a plausible answer — just not the same one twice. A damaged file recovers a different page tree between reconstructions, and an annotation without a normal appearance renders one of two images depending on hash order.

Both nondeterminism cases vary *within* one process, not only between runs, because Rust seeds each `HashMap` instance separately.

### Summary

| # | Location | Trigger | Today | Should be |
|---|---|---|---|---|
| 1 | `rendering/mesh_shading.rs:965` | shading function `FunctionType 3` with `/Domain []` | panic: index out of bounds | skip the shading |
| 2 | `rendering/mesh_shading.rs:1089` | `FunctionType 0` whose `/Size` exceeds its sample stream | panic: multiply with overflow | reject the function |
| 3 | `rendering/separation_renderer.rs:3062` | `/ImageMask` with `/Width -1 /Height -1` | panic: multiply with overflow | reject the dimensions |
| 4 | `rendering/page_renderer.rs:4178` | image whose `/Mask` sub-image has `/Width 0` | panic: subtract with overflow | skip the mask, paint the image |
| 5 | `rendering/page_renderer.rs:7750` | annotation `/AP` with `/D` or `/R` but no `/N` | identical bytes render two different images | draw nothing (§12.5.5) |
| 6 | `xref_reconstruction.rs:545` | damaged file with two `/Type /Pages` inside object streams | different page tree run to run | same bytes, same page tree |
| 7 | `structure/parser.rs:538` | `/K 4294967301` (2³²+5) | wraps to MCID 5, aliasing the real one | skip the out-of-range id |

Rows 1–4 panic in any build — the indexing and arithmetic here are checked regardless of profile.

### Steps to Reproduce

**Rows 1–4.** Build each malformed construct into a one-page PDF and call `render_page`. Every file parses cleanly; the failure is at paint time.

**Row 5.** Render the same bytes 64 times in one process and count distinct outputs. You get 2. With `/N` present you get 1.

**Row 6.** Reconstruct a damaged file containing two `/Type /Pages` objects packed into object streams, 64 times in one process, recording which tree is chosen.

**Row 7.** Give a marked-content reference `/K 4294967301` alongside a genuine MCID 5 on the same page, then read back the page's marked-content ids.

<details>
<summary><b>Observed on v0.3.77</b> — verbatim output</summary>

Rows 1–5, `cargo test --features rendering --test malformed_render_input`:

```text
test stitching_function_with_empty_domain_renders ... FAILED
test sampled_function_size_larger_than_stream_renders ... FAILED
test zero_width_mask_sub_image_renders ... FAILED
test separation_image_mask_negative_dimensions_render ... FAILED
test annotation_appearance_without_normal_entry_draws_nothing ... FAILED
test annotation_appearance_with_normal_entry_is_deterministic ... ok

panicked at src/rendering/mesh_shading.rs:965:32
panicked at src/rendering/mesh_shading.rs:1089:13
panicked at src/rendering/page_renderer.rs:4178:68
panicked at src/rendering/separation_renderer.rs:3062:23

annotation_appearance_without_normal_entry_draws_nothing:
  identical bytes produced 2 distinct renders   left: 2  right: 1

test result: FAILED. 2 passed; 5 failed
```

The two passes are controls: the `/N`-present annotation is already deterministic, and a stencil-mask variant I could not drive to the faulty branch.

Row 6, 64 reconstructions of the same bytes in one process:

```text
test objstm_recovery_picks_the_same_page_tree_every_run ... FAILED
recovered page tree varies between runs:
[Some(3), Some(4), Some(3), Some(3), Some(3), Some(4), Some(3), Some(3), Some(4),
 Some(4), Some(3), Some(3), Some(3), Some(4), Some(4), Some(3), Some(3), Some(4),
 Some(4), Some(3), Some(3), Some(3), Some(3), Some(3), Some(4), Some(4), Some(3),
 ... ]
```

Row 7, `cargo test --test structure_mcid_range`:

```text
test k_mcid_above_u32_max_is_skipped_not_wrapped ... FAILED
  left: [5, 5]            right: [5]
test negative_k_mcid_is_skipped_not_wrapped ... FAILED
  left: [4294967295, 5]   right: [5]
test out_of_range_mcr_mcid_is_skipped_not_wrapped ... FAILED
  left: [5, 5]            right: [5]
test in_range_k_mcids_are_unaffected ... ok

test result: FAILED. 1 passed; 4 failed
```

`[5, 5]` is the aliasing — the out-of-range id has landed on the page's real MCID 5, so two references now point at one id.

</details>

### Expected Behavior

**Malformed values should not panic.** Rows 1–4 all read a value straight out of the file and use it in arithmetic with no range check. An error or a skipped construct is the right outcome.

**Row 5 needs the fallback removed, not ordered.** ISO 32000-1 §12.5.5 makes `/N` the normal appearance; `/D` and `/R` show only while the pointer presses or hovers. Drawing one of those on a static page is wrong in every iteration order, so sorting the map would just freeze a specific wrong answer.

**Rows 6 and 7 should be deterministic.** The same bytes should recover the same page tree, and an out-of-range marked-content id should be skipped rather than truncated onto a valid one.

### Actual Behavior

Rows 1–4 panic inside `render_page` on a file that opened without error.

Row 5 returns one of two images depending on hash order, with no way for a caller to know which.

Row 6 returns a different page tree between reconstructions of the same bytes, so a damaged file can extract different content on consecutive runs.

Row 7 truncates the id onto a real one, leaving two marked-content references pointing at the same id.

### Environment

- **OS**: Linux x86_64 (Ubuntu 20.04)
- **Rust version**: stable (rustc 1.97.1)

### Checklist

**AI disclosure:** drafted with AI assistance and reviewed before filing. Every *Observed* block is real console output captured against v0.3.77, not reconstructed. Leaving the box above unticked rather than claiming otherwise.

tobocop2:
## Linked issue

Closes #997.

## What and why

Seven defects on the read path. Five crash on a file that opened cleanly. Two return a different answer on every run. Each defect starts from legal PDF syntax carrying an out-of-range value. The file parses, and it only fails once something reads the geometry.

| construct | on `main` | after |
|---|---|---|
| FunctionType 3 (stitching) function with empty `/Domain`, under a ShadingType 1 | panics | function rejected, shading skipped, page renders |
| Type 0 `/Size` bigger than its sample stream | panics | function rejected, page renders |
| `/Mask` sub-image 0 px wide or tall | panics | mask ignored, base image still paints |
| Separation `/ImageMask`, negative or unbacked size | aborts, asks for a 2^62-byte buffer | mask skipped, other plates intact |
| Annotation `/AP` with no `/N` | paints whichever sibling the `HashMap` yields first | draws nothing (§12.5.5) |
| `/K` marked-content id above `u32` | wraps, aliasing a real id | entry skipped |
| Damaged file, several page trees | recovers a different one per run | the same one, every run (ascending container order, `/Parent`-less root preferred) |

The seven fall into three shapes. Each defect class gets one commit, and the rationale sits in the commit bodies.

**Shape 1: malformed geometry panics the renderer.** `fix(render): reject out-of-range shading and image-mask geometry` bounds the declared grid against the sample stream. It does this once, up front. The stride and flat-index products then cannot overflow.

`fix(render): reject a separation image mask its stream cannot back` reuses `PageRenderer::image_mask_layout`. That helper already implements and tests this contract.

**Shape 2: marked-content ids wrapped instead of being rejected.** `fix(structure): reject out-of-range marked-content ids instead of wrapping` puts `u32::try_from` at all three cast sites, and the guard covers the other entry point too.

The content-stream `BDC /MCID` reader carried the identical `as u32` wrap. So `2^32 + 5` in a content stream aliased a real MCID 5. Both sides now route through the same `checked_mcid` guard. An out-of-range id behaves as a BDC with no `/MCID`.

**Shape 3: two read paths returned a different answer per run.** `fix(render): draw only the normal appearance stream of an annotation` removes the `/N`-less fallback. It does not merely order it. Painting `/D` on a static page is wrong in every iteration order.

`fix(recovery): pick the same page tree every time a damaged file is rebuilt` orders the recovery walk by object number. The nondeterminism sits *within* one process, because Rust seeds each `HashMap` instance separately. It therefore reproduces without separate runs.

**One deliberate behaviour change.** An `/AP` with `/D` or `/R` but no `/N` now renders blank instead of a press or hover state. With more than one entry it was previously a coin flip. This replaces randomness, not a settled result.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [x] I added a test that **fails before this change and passes after** (revert-checked against v0.3.77). Every reproducer is a minimal synthetic PDF built in code.
- [x] Test named by defect **class**, not an issue/PR number; no contributor or company names in code or fixtures.
- [ ] `cargo test` passes, **and** the affected feature tiers: `rendering`. Left unticked: the Rust test jobs (`Test (…)`, `Rendering Tests`, `Feature Tests`) are still queued on the current head commit. There is nothing to attest to yet.
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean. I verified `cargo fmt --check` on this head commit. `clippy -D warnings` was green on CI for the preceding commit, and the only change since is rustfmt line breaks.

These tests ship with the fixes: the seven reproducers below, and the marked-content overflow and negative cases with an in-range control. `ccitt_separation_image_mask_is_skipped_not_fabricated` covers the separation-stencil behaviour described under regression.

<details>
<summary><b>Verification</b>: red on v0.3.77, green after</summary>

On v0.3.77 the suite does not finish. The separation mask aborts the binary.

```text
running 7 tests
test zero_width_mask_sub_image_renders ... FAILED
test stitching_function_with_empty_domain_renders ... FAILED
test sampled_function_size_larger_than_stream_renders ... FAILED
memory allocation of 4611686018427387904 bytes failed
  process didn't exit successfully: (signal: 6, SIGABRT)
```

4611686018427387904 = 2^62 = `/Width 2147483648` x `/Height 2147483648`.

I re-ran the cases individually, so the cut-short ones still report:

```text
separation_image_mask_negative_dimensions_render      FAILED  panicked at separation_renderer.rs:3062
annotation_appearance_without_normal_entry_draws_nothing  FAILED  2 distinct renders, left: 2 right: 1
separation_image_mask_larger_than_its_stream_renders  SIGABRT
annotation_appearance_with_normal_entry_is_deterministic  ok      <- control, passes before too
```

After: `7 passed; 0 failed`.

Recovery, 64 reconstructions in one process:

```text
recovered page tree varies: [Some(3), Some(4), Some(3), Some(3), Some(4), ...]
```

Marked-content ids:

```text
k_mcid_above_u32_max_is_skipped_not_wrapped   left: [5, 5]           right: [5]
negative_k_mcid_is_skipped_not_wrapped        left: [4294967295, 5]  right: [5]
```

</details>

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 9,749 PDFs from my own collection. They are govdocs1 crawled .gov (manuals, spec tables, scanned forms, engineering reports), arXiv, PubMed Central OA, IRS forms, and pdf.js fixtures.

`corpus_sig` is not enough here. It hashes `extract_text`/`extract_words` and never renders. That makes it blind to `mesh_shading.rs`, `separation_renderer.rs` and both `page_renderer.rs` edits. This sweep renders instead.

- [x] Diffed against **`main`**. Every page of every document, **9,748 of 9,749 docs, 288,899 pages**. Per page: raster hash. Per document: text, table count, image geometry, and two in-process reconstructions.
- [ ] Diffed against the **latest release** (`v0.3.77`): same sweep; `main` is v0.3.77.
- [x] Judged with a structural metric: byte-identity of the rendered raster, not char-Levenshtein.

**Diff summary vs `main`:** one page changed, and it is the change this PR intends.

That page is `govdocs_033_033815.pdf` page 2. It carries a `/Widget` whose `/AP` holds only `/D`:

```text
page index 0  annots: 0
page index 1  annots: 1          <- the differing page
   [0] Subtype=/Widget  AP=['/D']  has_N=False
```

Nothing else moved. No text, table, image or reconstruction column changed, on any document. Page 0 alone was identical across all 9,749, so this only appears when every page is compared.

`govdocs_020_020747.pdf` is not covered: it aborts on `main` too (#1001), so neither arm can render it.

**Diff summary vs latest release:** not run separately. The sweep ran against `main`, which is at the v0.3.77 release commit.

**Scope caveat.** This render sweep never calls `render_separations`, and `paint_image_mask_to_plates` is reachable only through it. The byte-identity result therefore says nothing about `separation_renderer.rs`. That file's coverage is the synthetic cases plus the test added above.

The mask skip has one consequence on real files. CCITTFaxDecode is pass-through in the decode chain. CCITT-compressed separation stencils therefore always arrive shorter than their geometry needs. They are now skipped, where base painted zero-padded garbage.

The separation stencil path has no CCITT decode, though the page renderer's stencil path does. Until the separation path grows one, such stencils drop their ink instead of painting garbage. That is a follow-up candidate. So is consolidating the three independent Type 0 sampled-function evaluators, which now each carry their own bounds check.

<details>
<summary><b>Malformed-input coverage</b></summary>

Five of these constructs cannot occur in a well-formed file. A corpus sweep alone therefore cannot tell "the fix works" from "the fix is unreachable". Hence two arms.

**Arm A, regression.** Per document, both arms: raster hash per page, `extract_text`, `extract_tables`, `extract_images`, two reconstructions in one process. Byte-identity is the bar.

**Arm B, reach.** 56 generated documents: 29 malformed, 27 controls, over four entry points (28 `render_page`, 13 `extract_reading_order`, 8 `reconstruct_xref`, 7 `render_separations`).

Each case runs in a child process, with a deadline, 64 times. How the 29 malformed rows fail on `main`:

| how `main` fails | release | dev |
|---|---|---|
| panics | 8 | 9 |
| returns a different result than the fix | 11 | 11 |
| varies between identical runs | 8 | 8 |
| aborts (SIGABRT; `catch_unwind` cannot see it) | 1 | 1 |
| hangs (34 GB buffer from a truncated `/Width`) | 1 | 0 |

The release and dev columns differ because several defects are integer overflows. They abort a debug build. In a release build they silently compute a wrong index.

The gate exits non-zero on any of three conditions. A malformed case survives on both arms. A control's digest moves. The fix arm panics, aborts, hangs or varies.

```text
case                             main/release   main/dev
annotation_ap_d_and_r            varies:2       varies:2
annotation_ap_d_only             differs        differs
annotation_ap_r_only             differs        differs
mask_gray_both_zero              panic          panic
mask_gray_zero_height            panic          panic
mask_gray_zero_width             panic          panic
mcid_array_aliases_five          differs        differs
mcid_array_i64_max               differs        differs
mcid_array_negative_one          differs        differs
mcid_array_u32_max_plus_one      differs        differs
mcid_bare_aliases_five           differs        differs
mcid_bare_negative_one           differs        differs
mcid_mcr_aliases_five            differs        differs
mcid_mcr_negative_one            differs        differs
recovery_2_page_trees            varies:2       varies:2
recovery_3_page_trees            varies:3       varies:3
recovery_4_page_trees            varies:4       varies:4
recovery_5_page_trees            varies:5       varies:5
recovery_6_page_trees            varies:6       varies:6
recovery_7_page_trees            varies:7       varies:7
recovery_8_page_trees            varies:8       varies:8
separation_mask_i64_max          panic          panic
separation_mask_neg_one          panic          panic
separation_mask_neg_width_only   abort:sig6     panic
separation_mask_two_pow_31       abort:sig6     abort:sig6
shading_type0_size_over          differs        differs
shading_type0_size_overflow      panic          panic
shading_type0_size_u32max        panic          panic
shading_type3_domain_empty       panic          panic

27 controls: identical on both profiles.
56 case(s) x 2 profile(s), 0 blocker(s)      <- the branch as it ships
```

**Sabotage run.** I also checked the gate against a run that is supposed to fail. That run is this branch with `src/structure/parser.rs` restored to its `main` state, nothing else changed. 8 blockers is the pass condition there: exactly the eight cases that fix is responsible for. The other 48 stay green.

```text
mcid_array_aliases_five        survived   survived   <-- BLOCKER
mcid_array_i64_max             survived   survived   <-- BLOCKER
mcid_array_negative_one        survived   survived   <-- BLOCKER
mcid_array_u32_max_plus_one    survived   survived   <-- BLOCKER
mcid_bare_aliases_five         survived   survived   <-- BLOCKER
mcid_bare_negative_one         survived   survived   <-- BLOCKER
mcid_mcr_aliases_five          survived   survived   <-- BLOCKER
mcid_mcr_negative_one          survived   survived   <-- BLOCKER

56 case(s) x 2 profile(s), 8 blocker(s)      <- with parser.rs deleted
```

**What it cannot see.** Not your corpus. The generated documents prove these shapes, not every malformed file. Nothing here touches the writer, editor, signing or compliance paths.

</details>

<details>
<summary><b>Wider corpus inventory</b>: 19,151 documents, ~470,000 pages, run twice with matching results</summary>

|---|---|---|

| Verification surface | Scale | Result |
|---|---|---|
| synthesized malformed suite | 22 crash families | 44 baseline crashes drop to 6 |
| clean-corpus null | 470k pages | one page differs on revert; no clean-page regressions observed |

On the synthesized malformed corpus the baseline is 44 crashing families. This branch reduces them to 6, fixing 19 of the 22 families it targets.

**Observability caveat.** Reverting this fix changes only one page of 470k on the clean corpus. Its regression evidence therefore rests on the synthetic suite and unit tests rather than the sweep.

</details>

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: used throughout for investigation, the fixes, the test and sweep harnesses, and drafting this description.
- [x] I understand and can explain every line; the description and my review replies are written by me, not generated. This PR is not fully or predominantly AI-generated.

## Checklist

- [x] Commits follow Conventional Commits and are **DCO signed-off**.
- [ ] Docs/`CHANGELOG.md` updated if user-facing. Not done. `CHANGELOG.md` is written per release under a version heading and has no `[Unreleased]` section, so this looks like yours to write at release time.
- [x] Public-API changes considered for semver: none; `image_mask_layout` moved from private to `pub(crate)`.

### DIFF
--- a/src/extractors/text.rs
+++ b/src/extractors/text.rs
@@ -6682,9 +6682,13 @@ impl<'doc> TextExtractor<'doc> {
 
                 if let Some(props_dict) = self.resolve_bdc_properties(&properties) {
                     if let Some(mcid_obj) = props_dict.get("MCID") {
-                        if let Some(mcid) = mcid_obj.as_integer() {
-                            own_mcid = Some(mcid as u32);
-                            self.current_mcid = Some(mcid as u32);
+                        // Same id-space contract as the structure-tree side:
+                        // wrapping would alias a malformed id onto a real MCID.
+                        if let Some(mcid) =
+                            mcid_obj.as_integer().and_then(crate::structure::checked_mcid)
+                        {
+                            own_mcid = Some(mcid);
+                            self.current_mcid = Some(mcid);
                             log::debug!("Entered marked content with MCID: {}", mcid);
                         }
                     }
--- a/src/extractors/text.rs
+++ b/src/extractors/text.rs
@@ -6684,8 +6684,9 @@ impl<'doc> TextExtractor<'doc> {
                     if let Some(mcid_obj) = props_dict.get("MCID") {
                         // Same id-space contract as the structure-tree side:
                         // wrapping would alias a malformed id onto a real MCID.
-                        if let Some(mcid) =
-                            mcid_obj.as_integer().and_then(crate::structure::checked_mcid)
+                        if let Some(mcid) = mcid_obj
+                            .as_integer()
+                            .and_then(crate::structure::checked_mcid)
                         {
                             own_mcid = Some(mcid);
                             self.current_mcid = Some(mcid);
--- a/src/rendering/mesh_shading.rs
+++ b/src/rendering/mesh_shading.rs
@@ -962,7 +962,7 @@ fn eval_type3(
         return None;
     }
     let domain = dict.get("Domain").and_then(|o| o.as_array())?;
-    let (d0, d1) = (num(&domain[0]), num(domain.get(1)?));
+    let (d0, d1) = (num(domain.first()?), num(domain.get(1)?));
     let bounds: Vec<f32> = dict
         .get("Bounds")
         .and_then(|o| o.as_array())
@@ -1028,6 +1028,25 @@ fn eval_type0(func: &Object, dict: &HashMap<String, Object>, inputs: &[f32]) -> 
         return None;
     }
     if size.contains(&0) || domain.len() < m {
+        return None;
+    }
+    // `/Size` is document controlled. Bound the declared grid by what the
+    // sample stream can actually hold, so the flat-index and stride products
+    // below cannot wrap a bogus grid into a valid-looking sample offset.
+    let declared_bits = size
+        .iter()
+        .try_fold(1usize, |acc, &s| acc.checked_mul(s))
+        .and_then(|samples| samples.checked_mul(n))
+        .and_then(|values| values.checked_mul(bps as usize));
+    if declared_bits.is_none_or(|bits| bits > bytes.len() * 8) {
+        log::warn!(
+            "Skipping sampled function: /Size {:?} with {} outputs at {} bits declares \
+             more samples than the {}-byte sample stream holds",
+            size,
+            n,
+            bps,
+            bytes.len()
+        );
         return None;
     }
     let encode: Vec<f32> = dict
--- a/src/rendering/mesh_shading.rs
+++ b/src/rendering/mesh_shading.rs
@@ -1030,9 +1030,9 @@ fn eval_type0(func: &Object, dict: &HashMap<String, Object>, inputs: &[f32]) -> 
     if size.contains(&0) || domain.len() < m {
         return None;
     }
-    // `/Size` is document controlled. Bound the declared grid by what the
-    // sample stream can actually hold, so the flat-index and stride products
-    // below cannot wrap a bogus grid into a valid-looking sample offset.
+    // `/Size` is document controlled. Bounding the declared grid by what the
+    // sample stream holds keeps the stride and flat-index products below from
+    // wrapping a bogus grid into a valid-looking offset.
     let declared_bits = size
         .iter()
         .try_fold(1usize, |acc, &s| acc.checked_mul(s))
--- a/src/rendering/page_renderer.rs
+++ b/src/rendering/page_renderer.rs
@@ -4173,23 +4173,34 @@ impl PageRenderer {
                                 let mh = mask_gray.height();
                                 let iw = rgba_image.width();
                                 let ih = rgba_image.height();
-                                for y in 0..ih {
-                                    for x in 0..iw {
-                                        let mx = (x * mw / iw).min(mw - 1);
-                                        let my = (y * mh / ih).min(mh - 1);
-                                        let mask_val = mask_gray.get_pixel(mx, my)[0];
-                                        let pixel = rgba_image.get_pixel_mut(x, y);
-                                        pixel[3] =
-                                            ((pixel[3] as u32 * mask_val as u32) / 255) as u8;
+                                if mw == 0 || mh == 0 {
+                                    // No mask sample to test: leave the base
+                                    // image fully opaque rather than guessing.
+                                    log::warn!(
+                                        "Ignoring image Mask: it is {mw}x{mh} and carries \
+                                         no sample to test"
+                                    );
+                                } else {
+                                    for y in 0..ih {
+                                        let my =
+                                            ((y as u64 * mh as u64 / ih as u64) as u32).min(mh - 1);
+                                        for x in 0..iw {
+                                            let mx = ((x as u64 * mw as u64 / iw as u64) as u32)
+                                                .min(mw - 1);
+                                            let mask_val = mask_gray.get_pixel(mx, my)[0];
+                                            let pixel = rgba_image.get_pixel_mut(x, y);
+                                            pixel[3] =
+                                                ((pixel[3] as u32 * mask_val as u32) / 255) as u8;
+                                        }
                                     }
+                                    log::debug!(
+                                        "Applied image Mask ({}x{}) to image ({}x{})",
+                                        mw,
+                                        mh,
+                                        iw,
+                                        ih
+                                    );
                                 }
-                                log::debug!(
-                                    "Applied image Mask ({}x{}) to image ({}x{})",
-                                    mw,
-                                    mh,
-                                    iw,
-                                    ih
-                                );
                             }
                         },
                         Err(_) => {
@@ -4201,16 +4212,19 @@ impl PageRenderer {
                                     .map(|o| matches!(o, Object::Boolean(true)))
                                     .unwrap_or(false);
                                 if is_image_mask {
+                                    // `as u32` would turn a negative /Width into a
+                                    // near-u32::MAX row stride that passes the
+                                    // check below; try_from rejects it instead.
                                     let mw = mask_dict
                                         .get("Width")
                                         .and_then(|o| o.as_integer())
-                                        .unwrap_or(0)
-                                        as u32;
+                                        .and_then(|v| u32::try_from(v).ok())
+                                        .unwrap_or(0);
                                     let mh = mask_dict
                                         .get("Height")
                                         .and_then(|o| o.as_integer())
-                                        .unwrap_or(0)
-                                        as u32;
+                                        .and_then(|v| u32::try_from(v).ok())
+                                        .unwrap_or(0);
                                     if mw > 0 && mh > 0 {
                                         if let Ok(raw_mask_data) =
                                             doc.decode_stream_with_encryption(&mask_stream, ref_obj)
@@ -4253,9 +4267,14 @@ impl PageRenderer {
                                             let ih = rgba_image.height();
                                             let row_bytes = (mw as usize + 7) / 8;
                                             for y in 0..ih {
+                                                let my = ((y as u64 * mh as u64 / ih as u64) as u32)
+                                                    .min(mh - 1)
+                                                    as usize;
                                                 for x in 0..iw {
-                                                    let mx = (x * mw / iw).min(mw - 1) as usize;
-                                                    let my = (y * mh / ih).min(mh - 1) as usize;
+                                                    let mx = ((x as u64 * mw as u64 / iw as u64)
+                                                        as u32)
+                                                        .min(mw - 1)
+                                                        as usize;
                                                     let byte_idx = my * row_bytes + mx / 8;
                                                     let bit_idx = 7 - (mx % 8);
                                                     // PDF spec 8.9.6.2: mask bit 1 = paint (opaque), 0 = don't paint (transparent)
--- a/src/rendering/page_renderer.rs
+++ b/src/rendering/page_renderer.rs
@@ -7764,9 +7764,12 @@ impl PageRenderer {
             if let Some(ap_obj) = annot.raw_dict.as_ref().and_then(|d| d.get("AP")) {
                 let ap_stream_obj = doc.resolve_object(ap_obj)?;
 
-                // Normal appearance (N)
+                // ISO 32000-1 §12.5.5: /N is the normal appearance. /D and /R
+                // are the down and rollover appearances, shown only while the
+                // pointer presses or hovers the annotation, so an /AP without
+                // /N has no appearance to draw in a static render.
                 if let Object::Dictionary(ap_dict) = ap_stream_obj {
-                    if let Some(n_entry) = ap_dict.get("N").or_else(|| ap_dict.values().next()) {
+                    if let Some(n_entry) = ap_dict.get("N") {
                         let n_stream_obj = doc.resolve_object(n_entry)?;
                         if let Object::Stream { ref dict, .. } = n_stream_obj {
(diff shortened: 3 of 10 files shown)


yfedoseev:
These are 7 genuinely different fixes (two render panics, an image-mask panic, appearance-stream nondeterminism, xref reconstruction, MCID aliasing) bundled into one PR. Each one reads correct on its own — could you split them into individual PRs? That'll let each land independently instead of one CI hiccup blocking all seven.

tobocop2:
The split is up, one issue + one draft PR per fix, following your enumeration:

| Fix | Issue | PR |
|---|---|---|
| Shading: empty /Domain panic | #1074 | #1075 |
| Shading: /Size grid overflow | #1076 | #1077 |
| Image-mask geometry (three sites) | #1078 | #1079 |
| Appearance-stream nondeterminism | #1080 | #1081 |
| Xref reconstruction determinism | #1082 | #1083 |
| MCID aliasing (both entry points) | #1084 | #1085 |

The two MCID fixes (structure-tree narrowing and the BDC routing) are one defect at two entry points, so they travel as two commits in one PR rather than two PRs — say the word if you'd rather have them separate. The shared test file is split per defect class under distinct names, so the PRs merge in any order without conflicting. Union equivalence against this branch was verified per file before splitting.

Each split PR's corpus section fills in with its own sweep verdict today. I'll close this one once all the split PRs are verified by their corpus sweeps.

tobocop2:
All six split PRs are swept and ready for review, each with its own verdict against a fresh base from current main: #1075, #1077, #1079, #1081, #1083, #1085. Five swept clean outright; the mask-geometry split's 712 flagged pages are the skip-not-fabricate plate family, hand-triaged in the description. Per-file union equivalence with this branch was verified before the split. Closing this bundle in favour of the set.

tobocop2:
Byte-level accounting of the split, measured after the round closed: merge the six split heads onto current `main` and compare against this branch merged with `main`. `src/` is byte-identical — zero differences. The only delta is test organization: `tests/malformed_render_input.rs` became four per-defect-class files (shading domain, shading grid, image-mask geometry, appearance streams), each carrying its own copy of the shared PDF-builder helpers and only the imports it uses, so the split PRs merge in any order without add/add conflicts.

PR #1000, opened 2026-08-05
perf(render): decode a sampled shading function stream once, not per grid point

### LINKED ISSUE #999: perf: sampled shading function stream is decoded once per grid point
## Problem

`render_mesh_shading` (`src/rendering/mesh_shading.rs`) resolves a shading's `/Function` once, but evaluates it through the `to_rgba` closure — once per domain grid point for a Type 1 function-based shading (`render_function_based` walks a grid of up to 129×129 = 16,641 points), and once per vertex for the mesh types.

For a Type 0 (sampled) function, `eval_type0` calls `decode_stream_data()` at the top of every evaluation. The sample stream is therefore decoded up to 16,641 times per shading, and render time scales with stream size × grid area instead of paying the decode once per shading.

Measured on a 600×600 page painting one Type 1 shading over a Type 0 function (median of 5 renders, three interleaved rounds, one pinned core on an otherwise-idle 32-core box):

| sample stream | render time |
|---|---|
| 16 KiB | 62.1 / 62.7 / 62.7 ms |
| 64 KiB | 81.8 / 81.6 / 81.9 ms |
| 256 KiB | 153.6 / 163.7 / 166.5 ms |

Render time tracks the stream size even though the output is identical — which only happens if the stream is decoded per evaluation. Root cause confirmed by that scaling, not by reading.

## Proposed direction

Decode the stream where the function is resolved, in `render_mesh_shading`, and pass the bytes down to `eval_type0`. Output is byte-identical — the change is purely where the decode happens. With the decode hoisted, the same three cases render in ~56 ms flat (null experiment spread on the same box: 154.5–164.0 ms for base-vs-base at 256 KiB, so the 64 KiB and 256 KiB deltas are far outside the floor; the 16 KiB delta is inside it and not claimed):

| sample stream | base | fix |
|---|---|---|
| 64 KiB | ~82 ms | ~56 ms (~1.5×) |
| 256 KiB | ~154–167 ms | ~57 ms (~2.8×) |

Not covered, deliberately: a top-level array of functions and the children of a Type 3 stitching function are reached from the same per-grid-point path and would still decode per call. Same defect class, unmeasured: `eval_type4` also re-decodes *and* re-parses its PostScript program on every call even though `functions::Program` is compile-once/evaluate-many — typical Type 4 programs are small, so the impact is likely minor; worth its own measurement before touching.

## Acceptance criteria

- Byte-identity across a real-PDF corpus regression sweep, base vs fix: per-page render digests plus text/table/image extraction digests, zero differing documents.
- The targeted benchmark above with its base-vs-base null experiment; claim only deltas outside the floor.

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #999.

## What and why

**The old cost.** To paint a gradient, the renderer evaluates the shading's color function at up to 16,641 grid points. A Type 0 (sampled) function keeps its color lookup table compressed. The renderer decompressed that table from scratch at every one of those points.

Decompression cost therefore multiplied by the whole grid. Bigger tables made every gradient slower.

**The change.** This PR decompresses the table once, where the function is resolved. It then passes the decoded bytes to the evaluator. The diff is +33/−8 lines in one file, plus tests that pin the prepared-bytes path against a per-call decode.

Rendered output is byte-for-byte identical. Only the place where decompression happens has changed.

**Not covered.** Two cases deliberately keep the per-call behavior: a top-level array of functions, and the children of a Type 3 stitching function. Caching through them is a bigger change. #999 tracks that work alongside a Type 4 sibling.

**Benchmark.** The synthetic worst case is one gradient on a 600×600 page. The lookup table is 16, 64, or 256 KiB. Each figure is the median of 5 renders, over three interleaved base/fix rounds, on one pinned core.

| lookup table | base | fix |
|---|---|---|
| 16 KiB | 62.1 / 62.7 / 62.7 ms | 55.7 / 56.4 / 55.9 ms |
| 64 KiB | 81.8 / 81.6 / 81.9 ms | 56.0 / 56.1 / 56.0 ms |
| 256 KiB | 153.6 / 163.7 / 166.5 ms | 57.2 / 56.6 / 56.3 ms |

The fix stays flat at ~56 ms no matter how big the table gets. The 64 KiB win (~1.5×) and the 256 KiB win (~2.8×) sit far outside the measured noise floor. The 16 KiB delta sits inside it, so I do not claim it. Peak memory is equal across both arms.

The corpus holds 215 PDFs that use this kind of gradient. On those, full renders come out ~0.7% faster end to end. Real-world tables are small. The synthetic case is the cliff this change removes.

## Type of change

- [ ] Bug fix
- [x] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [ ] I added a test that **fails before this change and passes after**
      `rendering` (`rendering,test-support,cjk-render-fallback`) / `fips,icc` /
      the `cargo-hack` feature powerset / every language binding.

Output is byte-identical, so no behavioral test can fail on `main`. The first box therefore stays unticked. Equivalence is the property under test. `type0_prepared_bytes_matches_per_call_decode` pins the prepared-bytes path against a per-call decode. The corpus diff below is the end-to-end evidence.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents and ~470,000 pages. I ran the sweep twice independently, with matching null results.

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

No structural metric was needed. I fingerprinted rendered pixels, extracted text, tables, images, and page recovery. The sweep then compared those fields one by one for exact equality. That is stricter than any similarity judgement.

**Diff summary vs `main`:** I raced the branch against `main` (v0.3.77, `10b87f15`) over 14,910 real PDFs. Every page rendered and extracted in both builds.

| documents | differing | unstable | missing |
|---|---|---|---|
| 14,910 / 14,910 | **0** | 0 | 0 |

One corpus file (`govdocs_020_020747.pdf`) crashes identically on both builds. That is a pre-existing bug, unrelated to this change, and reported separately.

The single flagged page below flips between two orderings on unmodified `main`. It is not attributable to this PR.

|---|---|---|
| [057017.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/057.zip) (057.zip) | 6 | flagged page reproduces its flip on the same binary twice; not attributable to this PR |

**Diff summary vs latest release:** I did not run this separately. The sweep ran against `main` (`10b87f15`), which is the v0.3.77 release commit.

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: the code change, the benchmark and corpus runs, and this description.
      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

No CHANGELOG entry. The change touches one internal rendering module, and rendered output is unchanged. There is nothing user-facing to describe.


yfedoseev:
The decode-hoist logic and your inline tests both look right, and the extraction-side output is untouched, which is what I'd expect from a pure perf change. The one thing I still need before merging is the rendering-feature test actually run and its output reviewed — since this touches pixel output, I want that confirmed rather than assumed.

tobocop2:
Ran the rendering tier on this branch (head 0d127a38) on a clean 32-core Linux box, stable rustc 1.97:

```
cargo test --features rendering,test-support,cjk-render-fallback
```

Result: 358 suites, 9,241 tests, 0 failures. The two tests this PR adds both pass:

```
test rendering::mesh_shading::tests::type0_prepared_bytes_matches_per_call_decode ... ok
test rendering::mesh_shading::tests::type1_shading_renders_sampled_function_via_prepared_bytes ... ok
```

The first pins the prepared-bytes fast path against a per-call decode on every grid corner (a wrong or empty prepared buffer reads as all-zero samples and fails the expected values); the second renders a Type 1 shading through the prepared path end to end. On the corpus side, the full-corpus render sweep for this branch was clean against main — byte-identical pixel output — which is the expected result for a pure decode-hoist. Happy to run any other tier you want to see.

tobocop2:

PR #1002, opened 2026-08-05
fix(render): skip draws whose device coordinates exceed f32 pixel precision

### LINKED ISSUE #1001: [Bug]: Rendering one page of a real-world PDF crashes the process instead of returning an error
### Bug Description

Rendering one page of a real-world PDF crashes the library instead of returning an error.

`render_page` returns a `Result`, so a bad page should come back as an `Err`. For this document, rendering page 7 (index 6) panics inside the `tiny-skia` rasterizer that pdf_oxide uses under the hood.

For long-running processes it is worse than one failed page: in a corpus sweep that renders documents in a loop and catches panics to keep going, this document escalated to `thread caused non-unwinding panic. aborting.` — a hard abort that killed the whole worker process and every queued document with it. An application rendering untrusted PDFs can be brought down entirely.

### Steps to Reproduce

1. Fetch `020747.pdf` from the public-domain govdocs1 corpus: [`020.zip`](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/020.zip) (treated as a specification only, per the fixture rules).
2. Build with `--features rendering` (reproduced on current `main`, `10b87f15` / v0.3.77, release profile).
3. Run:

```rust
let bytes = std::fs::read("020747.pdf")?;
let doc = PdfDocument::from_bytes(bytes)?;
let opts = pdf_oxide::rendering::RenderOptions::default();
let _ = pdf_oxide::rendering::render_page(&doc, 6, &opts); // panics
```

### Expected Behavior

`render_page` returns `Err` for anything on the page it cannot rasterize. No panic, no abort.

### Actual Behavior

```
thread 'main' panicked at .../tiny-skia-0.12.0/src/alpha_runs.rs:170:51:
called `Option::unwrap()` on a `None` value
```

Process exits 101. Pages 1–6 render fine; text, table, and image extraction of page 7 are unaffected. Under `catch_unwind`-based isolation the same document produced the non-unwinding abort quoted above.

- Page count: 12; only page 7 (index 6) triggers the crash.
- The failing frame is tiny-skia's alpha-run blitter, so the likely trigger is a degenerate shape or clip (bad coordinates, empty region) in that page's content handed to the rasterizer. I have not yet minimized it to a synthetic reproducer; the govdocs file above is public domain and can serve as the specification for one.
- Candidate directions: sanitize the geometry pdf_oxide passes to tiny-skia, or move to a newer/patched tiny-skia if the edge case is fixed there. pdf_oxide currently pins `tiny-skia = "0.12"`.

### Environment

- **OS**: Ubuntu 22.04 (x86_64)
- **Rust**: rustc 1.97.1
- **pdf_oxide**: `main` @ `10b87f15` (v0.3.77), `--features rendering`, release build

### Additional Context

Found by the two-arm corpus regression sweep for #999 / #1000 (14,910 documents). The crash predates that change and reproduces identically with and without it.

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1001

## What and why

**The crash.** A real-world PDF has a damaged content stream. It crashes `render_page` instead of returning an error. The damage produces path coordinates as large as ~4.5×10¹⁸.

**Why it panics.** Numbers that big are beyond what f32 pixel math can represent precisely. The rasterizer (tiny-skia 0.12) trips over them: its antialiased run accounting loses sync and panics. A long-running process that catches panics to keep serving is not safe either. The panic can escalate to a hard abort that takes the whole process down.

**No upstream fix to take.** tiny-skia 0.12.0 is the latest release. The failing code has not changed upstream since 2023. The fix therefore belongs on our side of the boundary.

**The guard.** Before handing a path to tiny-skia, the guard checks where that path would land on the page. It runs the four corners of the path's bounding box through the transform. That costs twelve multiply-adds and no path copy. If any corner coordinate comes out non-finite or beyond the bound, the guard skips the draw and writes a debug log.

**Strokes and clips behave differently.** The guard checks strokes on outline reach instead. It narrows an over-reaching width rather than dropping the draw. Clip paths drop the clip instead of skipping a mask fill. Materializing an empty mask would erase every subsequent draw.

**The bound is measured, not guessed.** Against tiny-skia 0.12 at 72 dpi, fills and cubics alike still rasterize correctly at 5×10⁸ device units. Both produce nothing at 7×10⁸. The bound is the last value confirmed to work.

**2²⁴ is not the operative limit.** It is where f32 stops separating adjacent integers. But tiny-skia clips edges against the pixmap and keeps rasterizing well past it. I measured the decade below 5×10⁸ too, because a real document lives there. One corpus page fills to 1.3×10⁸ device units, which an earlier 1×10⁸ bound discarded.

**Three shared wrappers, not four entry points.** The guard covers the fill, stroke, and mask-fill wrappers. The four `PathRasterizer` entry points alone would not be enough.

The same content-derived geometry reaches tiny-skia through several other doors: the CMYK-sidecar coverage passes, tiling patterns, gradient extend clips, glyph fills, and both separation renderers. The coverage passes matter most, because they run *before* the guarded calls on the same operator.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

**Regression test.** It replays the exact filled path recovered from the failing document, verbs and coordinates alike. The reproducer is a minimal synthetic PDF built in code, so no third-party file is committed. I verified it red-then-green. Without the guard it dies inside tiny-skia with the reported panic (`alpha_runs.rs:170`).

**Sanity test.** It renders raw RGBA and asserts a non-white pixel. That makes it fail if the guard swallows all content. The old length assertion could not catch that.

**Further tests.** They drive the fatal geometry through the CMYK-sidecar coverage passes. They also drive it through the per-plate walker via `render_separations`. Degenerate clips and beyond-precision line widths are covered as well.

A 65-case fixture matrix (fills, strokes, CTM scales across the decades) shows zero regressions vs `main`. Every remaining difference is a case that `main` renders blank or aborts on.

| check | method | result |
|---|---|---|
| Reporting document (#1001) | all 12 pages, render + text/table/image extraction | no crash, extraction unchanged |
| Regression test | exact recovered geometry, synthetic in-code PDF | red on `main`, green with fix |
| Corpus sweep | 14,910 PDFs × 2 arms, all pages, field-by-field digests, no allowlist | 14,908 byte-identical; 2 differ, inspected below |
| Gate | fmt, clippy, full test, rendering tier, features matrix, rust-examples | green locally at the original head; CI re-running after a rustfmt fix |

      `rendering` / `fips,icc` / `ml` / binding (list which): rendering, features matrix.
- [ ] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean. Clean locally after the rustfmt fix. The CI Format Check and Clippy jobs have not reported yet.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents (~470,000 pages). I swept the corpus twice, independently, and the two runs matched.

**Full-corpus result.** 470,194 of 470,205 pages are byte-identical to `main`. The seven changed pages are exactly the expected set: six pages that panic inside the rasterizer on unmodified `main` and now render real content, and one page whose single draw sits at 3.78e15 device units, where any 32-bit float rendering is an approximation (f32 grid spacing there is roughly 2.7e8 units). That difference is documented rather than hidden. Zero token movement anywhere.

|---|---|---|

      scripts, or crashes).
      char-Levenshtein. I compared digests field by field for exact equality, which is stricter. No scoring metric was needed, because all but two documents were byte-identical.

**Diff summary vs `main`:** 2 documents out of 14,910 differ in the digest comparison. They differ in their all-pages render digest only. That is three pages in total. Every one of those pages is already garbage, and each change strictly reduces corrupt ink.

**Diff summary vs latest release:** not run separately. The sweep was against `main`, which sits at the v0.3.77 release commit.

**Sweep setup.** The base arm is `main` at v0.3.77 (`10b87f15`). The fix arm is this branch. Both arms are pinned by SHA, and I verified the fix tree independently. The two binaries are confirmed different.

Every page of every document rendered and extracted in both arms, capped at 64 pages per PDF. I compared the render, text, table, image, and recovery digests field by field.

| documents | render/extraction identical | differing | unstable | missing |
|---|---|---|---|---|
| 14,910 | 14,908 | 2 | 0 | 0 |

The two differing documents change **only** in their all-pages render digest. Text, tables, images, and recovery are identical. Both are damaged files from the same corrupt-producer family as the crasher. I rendered every changed page in both arms, three pages in total, all included in the evidence gist.

| page | base | fix |
|---|---|---|
| `020806` p5 | full-page black/red garbage wedges | same garbage minus one absurd sliver |
| `027613` p7 | entire page flooded solid yellow | flood reduced to half the page |
| `027613` p25 | **real manual page buried under giant black blobs** | **the actual page, clean and readable** |

The p25 pair follows, base on the left and fix on the right. The guard removes the corrupt draws that were painted *over* the document's real content.

 

Six corpus pages that panic `main` inside its rasterizer now render real content:

|---|---|---|
| [020747.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/020.zip) (020.zip) | 3, 6, 8 | main panics in the rasterizer; this branch renders real content |
| [002876.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/002.zip) (002.zip) | 14 | same recovery |
| [027613.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/027.zip) (027.zip) | 6, 24 | same recovery |
| [020806.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/020.zip) (020.zip) | 4 | known render difference on a draw at 3.78e15 device units, where any f32 render is approximate (grid spacing ~2.7e8 units); documented rather than hidden |

## AI assistance disclosure

- [x] AI assistance: **none**, or **assisted**. Tool: Claude Code. Extent: implementation, tests, corpus sweep tooling, and this description
      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).
- [ ] Public-API changes considered for semver (`semver-checks` will run).


yfedoseev:
This is careful work — the threshold is measured, not guessed, and the negative-control tests are thoughtful. One open question from your own fixture, though: when a shape has one vertex on-page and the rest at extreme coordinates, does skipping the whole draw silently drop the part that should have painted? I'd like to see a visual diff of the reproducing document's in-page geometry before merging — the tests confirm "no crash," not that question specifically.

tobocop2:
Here is the visual diff, plus a direct answer to the partial-on-page question. Fixture: a triangle fill with one vertex on-page at (100,100) and the other two far off-page, rendered at 72 dpi on a 612x792 page; the same geometry once below the 5e8 bound and once above it; and the reporting document's pages.

## Below the bound: both arms paint the on-page part, byte-identically

Vertices at 4e8 — inside the bound, so the guard passes the draw through and tiny-skia clips. `main` and this branch produce byte-identical pixels:

fill below bound

## Above the bound: the honest answer is yes, the on-page part does not paint — but main never painted it either

The same triangle with vertices at 6e8: this branch skips the whole fill and produces a blank page. On `main` there is no baseline that paints the wedge — the render panics and the whole page dies. The same holds for a stroke from an on-page point to 6e8: skipped here (its bbox corner is over the bound), process abort on `main`. So the guard never drops content `main` used to paint; what it drops is content an ideally-robust rasterizer would paint.

## The reporting document, main vs this branch

`main`: panic inside tiny-skia on pages 4 and 7 (the #1001 abort). This branch renders both; page 7's real content:

reporting page 7 under the guard

reporting page 4 under the guard

## Follow-up I would propose

Clipping the path to the device box before the bound check would recover the on-page part of an over-bound draw: the clipped path's coordinates are page-bounded by construction, so the guard would pass it and the wedge would paint. I kept it out of this PR because it changes the geometry pipeline for every draw rather than guarding the broken ones, and it deserves its own measurement. If you want it, I will build it as a follow-up with the same corpus discipline.

tobocop2:

PR #1008, opened 2026-08-05
fix: make extraction, redaction and table output order-independent

### LINKED ISSUE #1004: [Bug]: extraction output is not reproducible run to run — hash iteration order reaches three output paths
### Bug Description

Three output paths iterate std HashMaps whose per-process-random order reaches the result: (1) the page font-statistics tie-break — `max_by_key` returns the LAST maximal element, so two font-size buckets tying makes `dominant_em` a coin flip, and it feeds the multi-column gate that selects the reading-order branch, reordering the whole page's text; (2) redaction serializes `Object::Dictionary` (a HashMap) directly, so a redacted PDF's bytes differ between runs — `writer::ObjectSerializer` already sorts for this reason, redaction is a second copy that missed it; (3) spatial table detection orders rule families by hash iteration (latent — no corpus page reaches it, found by inspection).

### Steps to Reproduce

Run `extract_text` on the same PDF repeatedly in fresh processes; e.g. a page with the histogram 8pt:32 / 41pt:32 (an exact tie) alternates between two outputs. Redact any document twice and compare bytes.

### Expected Behavior

Same input, same binary → identical output.

### Actual Behavior

Measured: 6 of 12 runs disagreed on the tie page; redacted output not byte-stable.

Any PDF with a font-size tie; any redaction target. Reproduced across a 454,905-page corpus sweep where this was the entire noise floor (1 page).
### Environment

- Rust: stable 1.97

tobocop2:
## Linked issue

Closes #1004.

## What and why

Four output paths iterated `HashMap`s. Their per-process-random order reached the result.

**Dominant font size.** `max_by_key` returns the *last* maximal element. A font-size tie therefore made `dominant_em` a coin flip. That flip changed the multi-column reading-order gate. The same binary produced two different texts for the same page. Measured: 6 of 12 runs disagreed. The tie histogram was exactly 8pt:32 / 41pt:32.

**Redaction.** Redaction serialized `Object::Dictionary` unsorted, so redacted bytes differed between runs. `writer::ObjectSerializer` already sorts. Redaction was a second copy that missed it.

**Spatial table detection.** It ordered rule families by hash iteration. This one is latent; no corpus page reaches it.

**ExtGState validation.** It emitted its SMask/CA/ca/BM errors in per-process hash order whenever two or more entries violate. It now sorts with the same collect/sort/get idiom as the three sibling loops, and the rule-family grouping moves verbatim into a private helper so the ordering invariant is unit-testable. Behaviour is unchanged.

**Custom XMP properties.** `XmpWriter::build` wrote the custom-property map in `HashMap` iteration order, so the packet bytes written during redaction and export differed between runs. It now emits in sorted key order with the same collect/sort/get idiom, and a test pins the emitted order. This is the site review found; with it in, the title is true.

**One defect class, five sites, one fix shape: deterministic ordering.** Ties resolve to the smaller value, or to the lexicographically smaller name. For `dominant_em` that is the conservative direction. A tighter cluster gap makes the column gate stricter rather than looser.

One same-class site stays out of scope: PDF/A converter font-embedding order. It is pre-existing, untouched here, and a separate-issue candidate.

## Evidence

| check | result |
|---|---|
| flapping page before | 6 / 12 runs disagreed |
| after | 16 / 16 identical |
| corpus null experiment (same binary twice, 454,905 pages) | the flap was the ENTIRE noise floor (1 page); zero after |
| redacted output | byte-stable across runs |
| rule families | ordered by geometry; corpus null result recorded (latent) |
| this branch alone, `cargo test --release` | each site's branch green standalone (358 suites, 0 failures); the three commits touch disjoint files |
| tie-break tests | paired: one pins the tie resolution, one asserts a unique winner is unaffected |

## Type of change

- [x] Bug fix

## Tests

- [x] Paired tie-break tests per site: one pins the tie resolution, one asserts a unique winner is unaffected (table above)
- [x] Tables ordering is unit-tested. Tied first-coords must emit in input order. That test is revert-red against the old `HashMap` walk with near-certainty per process. It is hash-seed dependent, at a ~1/40320 false-pass rate. Distinct coords must order by coordinate, exactly as before.
- Affected feature tiers: none. core extraction, redaction, and compliance reporting
- [x] The full suite is green standalone at each site's head.

## Regression on real PDFs

Covered under Evidence: 14,911 documents / 454,905 pages per page. The corpus null experiment ran the same binary twice. It showed the flap was the entire noise floor before, and zero after. The merge base is the v0.3.77 release commit. The vs-release diff therefore equals the vs-`main` diff.

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: extensive (implementation, corpus validation, description drafting)

## Checklist

- [x] One logical change (one defect class: unordered iteration reaching output)
- [x] No public-API change

## Verification

Verified by a full regression sweep against `main`. It covered 19,151 documents, about 470,000 pages. It ran twice independently, with matching results.

|---|---|---|

determinism arm: it swept clean against main, after excluding the measured base-side noise floor. The flagged pages on its arm are the shared render-panic floor, not this change.


yfedoseev:
Good fix, but I found one more site with the same unguarded-iteration-order problem this PR is meant to eliminate — `src/writer/xmp_metadata.rs`, which affects custom XMP property output during redaction/export. Since it's the exact same pattern you're already fixing everywhere else, could you add that site too before this merges? Otherwise the PR title (order-independent output) won't quite be true yet.

tobocop2:
Added the `src/writer/xmp_metadata.rs` site in 14e2555b: the custom-property loop in `XmpWriter::build` now emits in sorted key order, with the same collect/sort/get idiom as the other sites here, and a test that pins the emitted order. Good catch — that loop wrote the packet in `HashMap` iteration order during redaction/export.

tobocop2:
Determinism engagement on the current head (14e2555b), measured on today's `main`: six sweeps of the stratified 1,497-document sample give the base 395 nondeterministic pages and this branch 394 — the branch recovers exactly its deterministic-tie page, which matches the earlier round's 394 versus 393 on the previous base. The remaining jitter is render-side float noise outside this PR's scope.

tobocop2:
@yfedoseev — a corpus datapoint for this PR, found while chasing something else.

While validating another branch I swept `main` twice over the same corpus and compared it against itself. Three pages in 470,207 differ between two runs of the same binary. One of them reproduces on demand.

`govdocs_005_005721.pdf` page 73, from the public-domain govdocs1 corpus ([005.zip](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/005.zip)), treated as a specification only. Extracting that page six times with one `main` build gives three different results:

```
main:   A  A  B  C  C  C
```

The same six runs on this branch:

```
this:   A  A  A  A  A  A
```

That matches the dominant-font-size tie described in the PR body: the flip changes the multi-column reading-order gate, and the text moves with it. The signature that moved on my side was rect-scoped text extraction, which is what a reading-order change would disturb.

Two things this gives you:

- A named document that shows the defect at page level, in addition to the 6-of-12 measurement already in the description.
- A reason the page matters beyond the extraction itself: while `main` has it, any differential validation on this corpus carries a page that flips regardless of the change under test, so the same page appears in unrelated comparisons and has to be explained each time.

## This unblocks #1063

The page above is the one flagged against #1063 in my sweep. It is flagged there whether or not #1063's change is present, because `main` flips it on its own, so it costs an explanation every time that comparison runs.

Merging this removes it. After that, a comparison against `main` on this corpus has no page that moves by itself, and a flagged page means the change under test actually did something. That is worth more than the one document: it is the difference between a differential result you can read directly and one that needs a caveat attached.

No action needed from me here. Recording it so the evidence sits with the fix.

tobocop2:
@yfedoseev — a sequencing observation about this batch, with a measurement behind it.

There are three open PRs that each remove a source of nondeterminism, and none has merged:

- **#1008** — `HashMap` iteration order reaching output in extraction, redaction, table detection, ExtGState validation and XMP
- **#1081** — an annotation's `/AP` resolved as `get("N").or_else(|| values().next())` over a `HashMap`, so the appearance state was a coin flip
- **#1083** — a damaged file recovering a different page tree between reconstructions

I think these three should have gone before everything else in the batch, ahead of the splits rather than alongside them, and I would still take them first now.

The reason is that every other PR in the batch is judged by comparing output against `main`. If `main` disagrees with itself, that comparison has a floor under it that belongs to no change. Sweeping `main` twice over 470,207 pages, three pages differ between the two runs. One reproduces on demand: `govdocs_005_005721.pdf` page 73 gives three different results across six runs of a single build, and on #1008's branch the same six runs give one.

What that cost, concretely, on my side today: I reported a flagged page on #1063 as a finding, told you to hold the merge, then had to retract it once I measured that `main` flips that page on its own. A fair part of the back and forth on that PR traces to the same root. Anyone validating against `main` on a large corpus will hit this and spend the same time working out that the baseline moved, not the code.

None of this is a criticism of splitting the work up. Breaking #996 and #998 into focused PRs was right and the smaller ones have been much easier to reason about. The narrower point is about order: when a batch contains determinism fixes, those want to land first, because they are what makes every later measurement mean something.

I have added a note to my other open PRs pointing here, so the dependency is visible where the evidence is. To be explicit about the kind of dependency: it is about evidence, not safety. Nothing in those PRs is unsafe without these three, and the ones you have already verified can merge whenever you like.

Queue order is yours and you have context I do not. Recording the measurement while it is fresh, in case it helps.

yfedoseev:
@tobocop2 — on the ordering point: you're right, and I've acted on it.

This PR is now the head of the merge queue. It was sitting fourteenth, behind the splits it would be used to validate, which is exactly the inversion you described. #1013 was mid-CI and I've stopped it; it goes second.

The reasoning holds independently of the corpus you measured on. The coin flip documented in this PR's own description is enough on its own: `max_by_key` returns the *last* maximal element, so a font-size tie flips `dominant_em`, which flips the multi-column reading-order gate — 6 of 12 runs disagreeing on the same page, with the tie histogram right there. Anything validated by comparing output against `main` inherits that, and there are still a dozen or so PRs behind this one that will be judged exactly that way. A baseline that stops moving on its own is worth more the earlier it lands.

I'm running my own determinism measurement rather than taking your numbers — not a comment on your work, just that a stability claim about our baseline is one we should each be able to reproduce. It doesn't change the ordering decision either way.

On the cost to you: chasing a flagged page that turned out to be `main` disagreeing with itself, and retracting the finding once you'd measured it properly, is a day I'd rather you hadn't spent. Naming the pattern while it was fresh was the useful thing to do, and the sequencing lesson is one I'll carry into the rest of the batch.

Nothing needed from you here — this one is verified on our side and goes through the queue as-is.

tobocop2:
sounds good, I forgot that I made a lot of non-determinism related fixes in this sea of changes and I should have emphasized that they merge first in the beginning and it just slipped my mind completely.

PR #1010, opened 2026-08-05
fix(extract): parse chars and spans with the same parser

### LINKED ISSUE #1006: [Bug]: Text drawn at an angle comes back horizontal, and some characters are lost
### Bug Description

A page has text drawn at an angle. A chart axis turned on its side, a rotated table header, a vertical stamp down the margin.

Read that page character by character and the angle is gone. The characters come back as though the line ran left to right, so their reported angle is zero and their positions are wrong. Some characters do not come back at all, and a sideways axis label can disappear completely.

Read the same page as text and the result is correct.

So one document gives two different answers depending on how you read it, and nothing in the output says which answer to trust. Anyone working at character level, for layout reconstruction or for redaction, is silently given the wrong one.

### Steps to Reproduce

1. Open a PDF with text drawn under a rotated text matrix. A chart with a vertical axis label is the common case.
2. Read the page as text with `extract_spans(0)`. The angle and positions are correct.
3. Read the same page character by character with `extract_chars(0)`.
4. Compare the two.

The characters report an angle of 0 and horizontal positions, and some are missing.

The problem appears on pages with a large content stream, over 256 KB.

### Expected Behavior

Both ways of reading a page agree about where a glyph sits and which way it faces.

A label turned 90 degrees comes back turned 90 degrees, with every character present.

### Actual Behavior

Characters come back at angle 0, positioned as though the text were horizontal.

Characters are dropped. On a rotated chart axis the whole label can be lost.

A full corpus sweep measured the disagreement on 1,134 of 454,899 pages.

### Code Sample

```rust
let doc = PdfDocument::open("rotated-axis.pdf")?;

// Read as text: correct.
for s in doc.extract_spans(0)? {
    println!("{:>5.1}  {:?}", s.rotation_degrees, s.text);
}
// 90.0  "Revenue (USD)"

// Read character by character, same page: the angle is gone.
for c in doc.extract_chars(0)? {
    println!("{:>5.1}  {:?}", c.rotation_degrees, c.char);
}
// 0.0  'R'   and some characters never appear
```

Any page with rotated text and a content stream over 256 KB. Measured on a public-corpus sweep. Agreement with PyMuPDF, pypdf and pdfminer improves in every rotation bucket once the two readings agree.

### Environment

- pdf_oxide: current `main`, reproduced at v0.3.77 and after
- OS: Linux x86_64 (Ubuntu 20.04 and 22.04) and macOS 14. The behaviour is platform independent.
- Rust: stable 1.97

### Additional Context

Cause, for whoever picks this up. The character path drives a second parser, and that parser reconstructs the graphics state around a text region by a different route than the text path. On a large content stream the two can agree on every text operator and still apply a different transformation matrix, which is where the angle is lost.

tobocop2:
## Linked issue

Closes #1006.

## What and why

`extract_chars` drove a second parser. That parser rebuilt graphics state by a different route than the span path. The two surfaces could therefore disagree about the same page. A sideways chart axis read flat in chars and rotated in spans. Both surfaces now share one parser.

**The unification is structural, not by convention.** `run_content_stream` is the single parser dispatch for chars, spans, and Form-XObject recursion. The modes can differ only in how `execute_operator` handles show-text operators. They cannot differ in which operators arrive.

**`parse_content_stream_text_only` is deprecated**, with a note naming the divergence. Its doc claimed equivalence with the streaming parser, and that is the claim this bug disproves. The function stays exported, since removal is semver-major.

## Evidence

| check | result |
|---|---|
| corpus differential, per page | 1,134 of 454,899 pages change; chars-side attributes align to spans; everything else byte-identical |
| reference agreement (PyMuPDF / pypdf / pdfminer.six) | improves in **every** rotation bucket, worsens in none |
| this branch alone, `cargo test --release` | 358 suites, 0 failures; fmt clean |

## Type of change

- [x] Bug fix

## Tests

- [x] The test builds a >256 KB stream whose rotated `cm` sits before `BT`, with padding invisible to the prescan. That stream must yield chars at 90°. The test is revert-red, because char mode's old parser returned those chars upright. It also asserts the 256 KB threshold explicitly, so a future threshold change cannot silently make it vacuous.
- Affected feature tiers: none. core text extraction

## Regression on real PDFs

Covered under Evidence, as a full-corpus per-page differential. The 1,134 changed pages are the chars-side attribute alignment this fix intends. The merge base is the v0.3.77 release commit. The diff against the release therefore equals the diff against `main`.

## AI assistance disclosure

- [ ] ~~written by me, not generated~~. See disclosure

## Checklist

- [x] One logical change
- [x] Public-API change considered for semver: `parse_content_stream_text_only` gains a deprecation attribute and stays exported; nothing is removed or altered in signature
- [ ] Docs/`CHANGELOG.md` updated if user-facing. No entry here. `CHANGELOG.md` is written per release, under a version heading, with a consolidated contributors block, so it reads as the maintainer's to write at release time.

## Verification

A full regression sweep against `main` verified the change. The sweep covered 19,151 documents (~470,000 pages). It ran twice independently, with matching results.

|---|---|---|

**Conservation.** The sweep re-rendered 190 flagged pages with both binaries. Those pages show zero text differences. Character, span, and stream counts are identical. The movement is confined to the char-mode view.

|---|---|---|
| [operator-in-TJ-array.pdf](https://raw.githubusercontent.com/mozilla/pdf.js/master/test/pdfs/operator-in-TJ-array.pdf) (pdf.js corpus) | 0 | char mode returned nothing on a malformed TJ array (0 chars); unified dispatch returns all 43 |
| [055378.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/055.zip) (055.zip) | 60 | flagged loss is base emitting a garbled duplicate table; fix drops the duplicate and repairs split numerals ("811 ,808" becomes "811,808") |
| [049928.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/049.zip) (049.zip) | 16 | known residual: markdown-only cell duplication (plain text identical); tracked in the follow-up work |


yfedoseev:
The parser-unification approach is right, but this one deliberately reroutes parsing behavior for every large-content-stream document, not just the rotated ones — bigger blast radius than the diff size suggests. Our checks found some real, precisely-scoped movement in a few documents at the character-extraction level that never reaches the final text output. Before merging I'd like to look at those together and confirm the new parser is more correct, not just different.

tobocop2:
Here is the character-level analysis, so we can confirm the movement is more correct rather than just different.

## Method

My sweep classified the branch's movement against base; 748 pages moved at the character-extraction level. I sampled 13 of them across the distinct clusters and re-dumped the full `extract_chars` surface (char, bbox, origin, rotation, fill color) at both pins, built back-to-back in one cargo session. On every sampled page the final text, markdown and lines output are byte-identical (one deliberate exception below).

## What moves, and the oracle that adjudicates it

Character counts are conserved on 12 of 13 pages; what changes is per-char graphics state. Two families:

**Fill color.** The old char parser tracked color state on its own and got it wrong where the span parser got it right. The span parser is the natural oracle here: this PR routes chars through it, and `extract_spans` output does not move — so base's own spans state what the correct value was all along.

- govdocs 005934 p12: old chars report the whole page at fill `0.137,0.122,0.125`; base spans for the same runs ("D15S22" at 60.80,748.67 and every other) say `0.000,0.000,0.000`. New chars: black. 2,983 chars corrected on this one page.
- govdocs 014455 p10: old chars mark rotated axis digits red (`1.000,0,0`); base spans for those rot-90 runs are black. New chars: black.
- govdocs 003977 p0: 1,246 of 3,645 chars change color only, same adjudication.

**Per-glyph advances.** On govdocs 004647 p11 the old parser's cursor drifts ~0.05 pt per glyph against the new one (196.09 vs 196.14 after four glyphs, growing along the line). The span parser's span edges match the new cumulative positions; the old chars disagree with the very spans they sit inside.

## The two exceptions, named

- 004647 p11 is also the one page whose char count changes: 770 -> 769. The delta is a single synthetic space the old parser emitted and the new one does not; the assembled text is byte-identical, so the space the reader sees never depended on it.
- The govdocs 049928 family is the deliberate output change in this PR (the md cell-duplication fix), not part of the never-surfaces class: chars conserved, markdown moves by design on chart-label pages.

So the movement your checks flagged is the char surface catching up to the state the span parser always reported. Happy to pull the full dump for any specific document you saw it on.

PR #1011, opened 2026-08-05
refactor(fonts): unify the duplicated glyph decoder

### LINKED ISSUE #1007: [Bug]: extraction and rendering decode glyphs with two different decoders
### Bug Description

Text extraction and rendering each carry their own glyph decoder with different feature sets. A Type0 font with a UTF-8 CMap extracts correctly and renders garbage; a ligature renders correctly but extracts lossily.

### Steps to Reproduce

Extract and render a document using a Type0/UTF-8-CMap font; compare the text each path produces.

### Expected Behavior

One decoder carrying the union of both paths' capabilities.

### Actual Behavior

Path-dependent results for the same glyph stream.

CJK and ligature-heavy documents; validated corpus-wide (the unification changed one page's tokens, an improvement, on a 454,905-page sweep).
### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

> **Draft while I re-validate.** ~~#1013~~ merged in a different shape from the one this branch carried, so I rebuilt the change as a single commit on current `main`. Evidence follows.

## Linked issue

Closes #1007.

## What and why

The glyph decoder existed three times. Extraction and rendering each carried their own copy of the byte-mode choice, the character iterator and the Unicode fallback table.

The copies had drifted, and the render path holds the worse one. Its fallback table resolves a fifth of the codes the extraction table resolves. Its byte-mode choice ignores a ToUnicode CMap that declares a two-byte codespace, so a CJK font whose `/Encoding` name says otherwise decodes one byte at a time while extraction reads the same font correctly.

One decoder now lives in `fonts::unicode_decode` and both paths call it. Rendering picks up the fuller fallback table and the byte-mode fix, so text that rendered wrong for those fonts now renders as it extracts.

## What stays different, as policy

What genuinely differs between the paths is stated as a `DecodePolicy` rather than reimplemented. Extraction keeps an unmapped code and writes a question mark for one no table resolves. Rendering expands a ligature so the shaper sees ordinary glyphs, and paints nothing for a code it cannot resolve.

## The drop tally is unchanged

The render path keeps its own tally, and keeps the rule ~~#1013~~ established: that decode runs before the run is routed, and the byte-to-GID and direct-CID paths paint correctly from raw bytes for exactly the font classes whose Unicode decode fails, so only the chosen paint path records a drop.

## Widths follow the decode

The rasterizer builds its parallel CID and width arrays from the decode's own segmentation, so a variable-width UTF-8 code paints against the width it was decoded with.

## Type of change

- [x] Refactor / internal
- [x] Bug fix

## Tests

Validation is running. This one is not expected to be byte-identical against `main`: the render path adopts the fuller fallback table, so pages holding the affected fonts should change, and I will read what moves before this leaves draft.

## Regression on real PDFs

**Corpus I tested**: 19,151 documents / 470,204 pages (govdocs plus the pdfium, pdfjs and mutool exotic sets), swept against `main` **with rendering on**, comparing the rasterized output of every page.

- [x] Diffed against `main` — 470,201 pages byte-identical. Three pages rasterize differently and all three paint text `main` drops or garbles: `noembed-jis7.pdf` p0 (`ABC $ " $ $ $ &%3( $ *` becomes `ABC 123`), `issue20232.pdf` p0 (a Russian title block whose labels were missing), `issue9534_reduced.pdf` p0 (three glyphs become six). I read the rendered images for all three.
- [ ] Diffed against the **latest release** — not run; `main` is the baseline that matters here.
- [x] Judged by per-page output hashes across text, markdown, HTML and the rasterized page, then by reading the images for every page that moved.

Extraction output does not move at all: the extraction copy of the decoder was byte-identical to the shared one, so only the render path gains anything.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).
- [x] Public-API changes considered for semver. No public API changes: `fonts::unicode_decode` is `pub(crate)` and no exported signature moves.


yfedoseev:
This is the one I want to be most careful with — a shared decoder used by both extraction and rendering is exactly where a subtle regression would be hardest to notice. The title says `refactor:`, so I want to confirm this is genuinely behavior-preserving and not a fix riding along unlabeled. Before merging: could you add characterization tests comparing decoder output across font types (TrueType, Type1, CID/Type0, embedded subsets)? Also, the rendering-side integration test in this PR is currently marked `#[ignore]` — I'd like that finished and actually passing, since right now there's no real end-to-end coverage for the rendering half of this change.

tobocop2:
Both asks are in, pushed as a9e7e408.

**Characterization tests** (`tests/test_decoder_characterization.rs`): a font built in code — four glyphs, advances 500/600/700 per mille — embedded four ways: simple TrueType with a byte-indexed cmap, simple TrueType with a Windows Unicode cmap, Type0/Identity-H CID with a ToUnicode CMap, and a subset missing a glyph. Per family the tests pin the decoded string, the per-char sequence, the advances against the declared `/Widths`/`/W`, and (under `rendering`) that a render paints every glyph the extractor decodes without a `GlyphDropped` warning. The subset family pins the cross-surface contract directly: the unpaintable code keeps its character in extraction and raises exactly one drop warning in rendering. Any behavioral drift in the unified decoder — string, segmentation, metrics, or paint parity — turns one of these red.

**The ignored render test is finished and passing**: the fixture it was waiting for is the same in-code subset font, mapping only 'A' through a byte-indexed cmap, so painting "AB" provably reaches the drop path. The `#[ignore]` is gone and the test asserts the warning surfaces and names the dropped code.

All six characterization tests plus the drop test pass under `--features rendering`, and the full library suite is green.

yfedoseev:
Heads-up while running this branch through an independent regression sweep against current `main`. **Mostly FYI — I'll handle it myself, since it's a collision with my own change in `main`.** Nothing needed from you unless you disagree with the approach.

## Conflict with #1039 (my change)

Both this branch and #1039 (`fix(text): Tw must not apply to a multi-byte CID 32`, merged 2026-08-15) edit `measure_text_bytes` in `src/rendering/text_rasterizer.rs` at the same lines, for two different and both-correct reasons:

- **This branch** switches from `TextCharIter` to the new `char_codes()`, so the measured (invisible-text) advance path uses the same decode segmentation as the painted path. For a variable-width UTF-8-codespace CMap, `TextCharIter`'s fixed grouping desyncs from `decode_text_to_unicode` — exactly the decoder-parity class this PR is about.
- **#1039** added per-code byte-width tracking so word spacing applies only to the single-byte code 32 (ISO 32000-1 §9.3.3), never a 2-byte CID `0x0020` under Identity-H/V.

The snag: `char_codes()` returns `Vec<u32>` — codes only, no byte width — so taking this branch's side as-is silently reverts #1039 and re-applies Tw to 2-byte CID 32. Taking #1039's side reintroduces the segmentation desync this PR fixes. Neither side alone is correct post-merge.

Worth noting every *other* site #1039 touched (`render_unicode_text` ×2, `render_cid_direct`, `render_substituted_cjk`) merges cleanly and keeps its fix — the collision is isolated to `measure_text_bytes`.

## Resolution

Added a width-carrying sibling in `src/fonts/unicode_decode.rs`. The byte width already exists on both paths and is simply discarded — `utf8_codes()` computes it internally, and `TextCharIter` already yields `(code, width)`:

```rust
pub(crate) fn char_codes_with_widths(bytes: &[u8], font: &FontInfo) -> Vec<(u32, usize)>
```

`measure_text_bytes` then uses it and gates on `nbytes == 1 && code == 0x20`, keeping both fixes. The rule generalizes cleanly to the UTF-8-CMap case too: a UTF-8-codespace `0x20` is always width 1, which is precisely the plain-ASCII-space case Tw is meant to cover — so the two fixes were never actually in tension, `char_codes()`'s signature just couldn't carry the information.

## Verified

- Compiles clean against a real merge of this branch + current `main`.
- All 7 `measure_text_bytes` tests pass, including one I added for the combined case (UTF-8-CMap font, single-byte trailing space after a multi-byte code must still receive Tw).
- `fonts::unicode_decode` + `rendering::text_rasterizer` suites green (11/11).
- **445-document regression sweep vs `main`: 0 documents changed, 0 crash/timeout changes** — byte-identical text/markdown/HTML output across the curated 419-doc corpus plus all 88 `pdf_benches` files.

To be clear about where this lives: I have **not** pushed anything to this branch. The resolution sits on a local branch of mine and will land in `main` as my own follow-up commit when this merges. Your PR is unchanged.

One caveat on the sweep result, in fairness to it: 0/445 proves the refactor is *text-output-preserving*, and says nothing about painted glyphs — this branch touches `text_rasterizer.rs` and `page_renderer.rs`, which a text diff can't see. I'm running a separate pixel-level render sweep to close that, and will follow up here with the result.

yfedoseev:
Following up on the rendering caveat I flagged earlier — the pixel-level sweep is done.

**Result: 0 pixel differences vs `main`, 270/270 pages identical.**

Method: rasterized each page and compared a per-page signature (pixel hash, mean RGB, and non-white coverage — the mean/coverage columns so a diff can distinguish an antialiasing nudge from a page that went blank, which a hash alone can't). Corpus was 186 documents chosen to stress the paths this branch touches: 60 shading/pattern, 49 separation/DeviceN, 40 image-mask, plus a general slice.

So with the `char_codes_with_widths()` conflict resolution applied, this branch is now clean on every surface I can measure:

- 445-document text/markdown/HTML sweep: **0 documents changed**
- 186-document / 270-page render sweep: **0 pixels changed**
- `measure_text_bytes` unit tests: **7/7**, including the added regression test for the combined UTF-8-CMap + single-byte-space case
- `fonts::unicode_decode` + `rendering::text_rasterizer` suites: **11/11**

That closes the gap I mentioned — the earlier text-only result proved the refactor was output-preserving for extraction but said nothing about painted glyphs. It now says both.

Still not pushed to your branch: the resolution lives on a local branch and will land in `main` as a follow-up commit when this merges.

yfedoseev:
Change of approach from my note above, so you're not surprised by a commit you didn't write appearing on your branch.

I said the #1039 conflict resolution would sit on a local branch of mine and land in `main` as a follow-up after this merged. I've instead pushed it here as `777f8b13` (`merge: resolve #1011 x #1039 conflict in measure_text_bytes`), for one reason: landing it separately means CI never tests the combination, and `main` would briefly carry the reverted #1039 between the two commits. Pushing it here gets the full pipeline run against what actually merges.

The commit is the merge of current `main` plus the `char_codes_with_widths()` resolution described above — no other changes to your work. Sign-off is mine.

If you'd rather own the resolution yourself, say so and I'll drop the commit and let you push your own.

tobocop2:
> Change of approach from my note above, so you're not surprised by a commit you didn't write appearing on your branch.
> 
> I said the #1039 conflict resolution would sit on a local branch of mine and land in `main` as a follow-up after this merged. I've instead pushed it here as `777f8b13` (`merge: resolve #1011 x #1039 conflict in measure_text_bytes`), for one reason: landing it separately means CI never tests the combination, and `main` would briefly carry the reverted #1039 between the two commits. Pushing it here gets the full pipeline run against what actually merges.
> 
> The commit is the merge of current `main` plus the `char_codes_with_widths()` resolution described above — no other changes to your work. Sign-off is mine.
> 
> If you'd rather own the resolution yourself, say so and I'll drop the commit and let you push your own.

You can own the resolution, all good by me.

yfedoseev:
@tobocop2 — thanks, noted, I'll keep the resolution on this branch then.

For visibility: this one is verified clean on our side — 445 PDFs from our corpus with zero surface differences on text/markdown/HTML, and a separate rendering pass over 270 pages that came out pixel-identical to `main` (worth doing given the change is in glyph decoding, where output can move without the extracted text changing).

It's in the merge queue behind #1013, since your description flags that one as a prerequisite. Nothing needed from you.

tobocop2:

tobocop2:
Conflicts are resolved: since #1013 and #1039 both landed, I rebuilt this as a single commit on current `main` and carried your `777f8b13` resolution forward, `char_codes_with_widths` in `measure_text_bytes` and its Tw regression test included.

tobocop2:
## Corpus check, rendering on

I swept 19,151 documents and 470,204 pages with rendering enabled, against `main`, and compared the rasterized output of every page.

470,201 pages are byte-identical. Three pages rasterize differently, and all three paint text that `main` drops or garbles.

## The three pages

**`exotic/pdfjs/noembed-jis7.pdf` page 0.** `main` paints `ABC $ " $ $ $ &%3( $ *`, with glyphs landing on top of each other. This branch paints `ABC 123`. The font is a JIS7 CMap: `main`'s render-side byte-mode choice ignores a ToUnicode CMap that declares a two-byte codespace, so it reads the string one byte at a time.

**`exotic/pdfjs/issue20232.pdf` page 0.** A Russian drawing title block. `main` paints stray dots and a few `№` marks; the labels themselves are missing. This branch paints them: `Перв. примен.`, `Справ. №`, `Изм. Лист № докум. Подп. Дата`, `Разраб.`, `Т.контр.`, `Копировал`, `Формат`.

**`exotic/pdfjs/issue9534_reduced.pdf` page 0.** `main` paints three glyphs, `Ö Ý Ź`. This branch paints six, `Ö Ý Ź V Î Ý`. The codes the render path could not resolve painted nothing.

## Why only three

The extraction copy of the decoder and the shared one are identical, so extraction output does not move at all. Only the render path gains anything, and only for fonts whose codes its thinner fallback table could not resolve. That also fits your 270-page rendering pass coming out pixel-identical: the corpus needed is narrow.

## Tests

`cargo fmt --check` clean, `cargo build --features rendering` clean with no warnings, and 9,033 tests pass with no failures.

## On the merge

Rebuilt as a single commit on current `main`, with your `777f8b13` carried forward: `char_codes_with_widths` in `measure_text_bytes` and the Tw regression test for a UTF-8-codespace CMap.

One test needed rewriting rather than rebasing. `preserve_unmapped_glyphs_gates_all_filter_sites` counted references in `src/extractors/text.rs` and wanted at least nine; unifying moved one filter site into the decoder, where the flag arrives as `DecodePolicy::preserve_unmapped`. It now checks both homes and asserts that every U+FFFD filter in the decoder is paired with a policy check, so an ungated filter still fails.

PR #1012, opened 2026-08-05
fix(convert): read rotated pages in one frame on every text surface

### LINKED ISSUE #984: [Bug]: A sideways page reads correctly as text but scrambled as Markdown or HTML

tobocop2:
> [!IMPORTANT]

> ~~**Blocked by #982 and #987** (#987 is itself blocked by #982). Merge those first. Until they land, this diff carries their commits. After they land, it rebases to a pure delta.~~
>
> **Unblocked.** #982 and #987 are merged, and the branch is rebased to the pure delta: `src/document.rs` plus the parity test. The carried commits are gone, so `tests/test_rotated_line_grouping.rs` and `tests/test_rotated_run_word_boundaries.rs` no longer appear in the diff.

## Linked issue

Closes #984.

## What and why

**The defect.** The rotated reading frame was applied only inside `extract_text`. `to_markdown`, `to_html`, `to_plain_text` and the filtered surfaces assembled in raw page space. A rotated page therefore read one way through text and another way through every other surface. `extract_text_in_rect` mapped before filtering. Its rect selected in mapped coordinates, while every other rect surface used page space.

**The fix.** Every surface now enters the reading frame through one choke point. That choke point sits after the region filters. The ordering is therefore structural, not per-caller discipline.

**Rects.** Rects are page-space everywhere. `extract_text_in_rect` moves from mapped space to page space. This is the one deliberate behavior change. It matches the documented `extract_words` contract that callers feed rects back from.

**OCR extra spans.** All three `*_with_extra_spans` surfaces now merge the extra spans before the filter and the map. Markdown and HTML previously merged after the map. They filed the extra text in the wrong reading-order slot.

## Known residual, deferred

18 pages still produce markdown that differs from the other surfaces. Those pages are a separate frame-selection case. This change does not address them. They are tracked for follow-up rather than widened into this change.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

**CI status.** The last two boxes stay unticked pending CI. I just pushed a clippy fix (`cloned_ref_to_slice_refs`) and a rustfmt fix. Every job on the current head is still queued. On the previous head, the only failures besides those two lints were external-service outages unrelated to this branch. They were a Maven Central 403 on the Scala job and a GitHub action-download failure on a macOS Python job.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents / ~470,000 pages. The kinds are scanned, tagged, malformed, CJK vertical writing, forms and academic two-column. The table below breaks the corpus down by source.

**Diff summary vs `main`:**

| gate | result |
|---|---|
| rotated pages | 1,620 read identically across text, markdown, HTML and plain text |
| upright pages, markdown and HTML | byte-identical across all 392,109 |
| core columns (text, words, spans, lines, chars, table cells) | unchanged on all ~470,000 pages; only the markdown/HTML views move |
| markdown scored against the unchanged text layer, 600 changed pages | missing tokens -88%, spurious tokens -70%, token-order agreement 0.50 → 0.78 |

| Corpus | Documents | Source |
|---|---|---|
| govdocs1 | 14,903 | [digitalcorpora.org](https://digitalcorpora.org/corpora/file-corpora/files/) |
| curated public documents (arXiv, technical manuals, Japanese vertical writing) | 8 | [arxiv.org](https://arxiv.org) and public sources |

| Document | Pages | Shows |
|---|---|---|
| [002587.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/002.zip) (002.zip) | 251 | the flagged "lost tokens" ("anicn", "ethnicitBlac") are fragments the old assembler glued; the new markdown matches the clean text layer |

**Diff summary vs latest release:** not run. The sweep above is against `main` only.

      scripts, or crashes).
      char-Levenshtein.

## AI assistance disclosure

- [x] AI assistance: **assisted**. Tool: Claude Code. Extent: implementation, the regression sweep, and this description.
      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

All commits use Conventional Commits. The final `style:` commit is missing its DCO sign-off. It needs a `--signoff` amend before merge.


yfedoseev:
The reading-frame consolidation looks right and I chased down what looked like one regression — turned out to be a false positive in our own tooling, not your code, so no concern there. This one depends on #982 and #987 landing first (it carries their commits) — once those are in, can you rebase so I can review the actual isolated diff? Also missing a DCO sign-off on the current head commit.

tobocop2:
Quick note on the DCO item: I checked all five commits on this branch and each one has a `Signed-off-by` trailer, and the DCO check is passing on the current head (`56fad9f1`). Let me know if you see it missing somewhere and I'll fix it right away.

Agreed on the rebase: once #982 and #987 land I will rebase onto main so the isolated diff is what you review.

yfedoseev:
@tobocop2 — your dependencies have landed, so this is ready for the rebase your description anticipated.

**#982 and #987 are both merged** (`fb22c34b`, `428caede`), along with #989 (`4bc5be49`), #1057 (`f45f51ab`) and #1010 (`d0fb8b50`). Your PR body says:

> Until they land, this diff carries their commits. After they land, it rebases to a pure delta.

Merging current `main` in now produces conflicts in 4 files — `src/document.rs` (1 hunk), `src/extractors/text.rs` (2 hunks), `tests/test_rotated_line_grouping.rs` and `tests/test_rotated_run_word_boundaries.rs` (1 each). That's expected given the carried commits, and the rebase should collapse most of it.

I'm not resolving these myself. On two other PRs in this batch (#1011, #1063) the merge came out *textually clean* but silently reverted an already-merged fix, so I'd rather you rebase to the pure delta than have me guess at intent in `document.rs` and `extractors/text.rs`.

Worth saying that the fix itself measured well: on a 445-document sweep against `main`, the one document that changed was `pmc/PMC8154847`, where a scrambled rotated-page table now reassembles into coherent rows with its definitions footnote intact — a clear improvement, no content lost. So this is purely a rebase, not a correctness question.

I've moved the merge queue on for now — ping here when it's pushed and I'll re-queue it.

One heads-up from #1009, since your test files are affected here too: the `Rendering Tests` job builds with `-D warnings`, so an `unused_mut` that's a warning everywhere else becomes a hard error there. Worth a local `cargo build --tests` before pushing.

tobocop2:
Rebased and pushed. The head is `f102e8a8`.

I rebased rather than merged, so the carried commits are gone. The delta is `src/document.rs` plus `tests/test_rotated_converter_parity.rs`. The two test files from #982 and #987 no longer appear. A cherry-pick onto `main` leaves no hunk in which an already merged fix can be reverted, which is the risk you asked me to remove.

`RUSTFLAGS="-D warnings" cargo build --tests --features rendering` is clean, so the `unused_mut` class you flagged on #1009 is covered here too.

## Re-swept, because the old numbers no longer apply

The 470,000-page table in the description predates #1010. #1010 rewrote `src/extractors/text.rs`, which produces the spans these converter surfaces read, so I re-swept the merged head against current `main` (`51c52c50`) rather than rely on it.

| measure | result |
|---|---|
| pages compared | 470,207 |
| byte-identical across all measured columns | 467,872 |
| **blockers** | **0** |
| pages in only one arm | 0 |
| required documents missing from an arm | 0 |

The remaining movement is class-permitted and lands on the markdown and HTML columns of rotated pages, which is what this PR changes on purpose.

Ready to re-queue.

tobocop2:

PR #1013, opened 2026-08-05
fix(render): report glyphs that paint nothing instead of dropping them

### LINKED ISSUE #991: [Bug]: Characters vanish from rendered pages with no warning

tobocop2:
## Linked issue

Closes #991. Blocks #1011.

## What and why

A glyph that failed to paint vanished silently. The cursor still advanced. Four rasterizer sites swallowed the drop. A `None` outline from ttf-parser painted nothing. An unmapped non-whitespace char never painted, not even `.notdef`. Chars resolving to U+FFFD left the shaping input entirely. And total loss was reported at `debug` only.

**Every paint path now records its drops.** Each one emits a `GlyphDropped` warning. The warning names the font, the first char code and glyph id, and the count.

**A broken font is named once per font per page.** Render workloads re-hit the same font on every text run, so reporting is deduplicated - but a process-lifetime latch went silent for the rest of the process in exactly the bulk-ingestion case the linked issue describes. The latch now lives on the rasterizer and clears at the start of every page, the same scope as the `k_zero_warning_emitted` latch in the page renderer, with no entry cap to run out of. An end-to-end test builds a broken-cmap font in code, renders a two-page document, and asserts the warning fires once per page; a process-scoped latch fails its second page.

**Painting is untouched.** Raster output is byte-identical across the corpus. Zero upright pages changed. This is reporting only.

**Public API.** `WarningCategory` gains a `GlyphDropped` variant. It also becomes `#[non_exhaustive]`. Downstream exhaustive matches therefore keep compiling, so the addition is a minor bump rather than a break.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

The reproducers are unit tests over the new drop tally, not a synthetic PDF. The four sites need a font whose glyphs fail to outline. A PDF built in code cannot express that without shipping a broken font file. The tests cover the no-drop case, the warning contents, and the once-per-font gate.

On a clean 32-core box the full rendering-feature suite is green for the current head, including the new end-to-end test.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents (~470,000 pages), swept against `main`. I ran the sweep twice and got matching results.

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** Three pages were flagged. All three reproduce their differences on unmodified `main`. None are attributable to this PR.

| Document | Pages | Behaviour |
|---|---|---|
| [046432.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/046.zip) (046.zip) | 3 | text-order flip reproduces on main alone |
| isartor conformance page (veraPDF suite) | 0 | render hash alternates between the same two values on main and on this branch |

**Diff summary vs latest release:** not run. The sweep was against `main` only.

## AI assistance disclosure

- [x] AI assistance: **none**, or **assisted**. Tool: Claude Code. Extent: drafting the fix and this description, under review at each step
      predominantly AI-generated.

## Checklist

All commits follow Conventional Commits. The last one (`style: fix rustfmt findings`) went in without a `Signed-off-by` trailer. It needs a sign-off amend before this is mergeable.


yfedoseev:
Good diagnosis and the warning infrastructure reuse is the right call. One real issue: `WARNED_FONTS` is process-lifetime and capped at 64 entries with no reset, so in exactly the bulk-ingestion scenario your linked issue describes, warnings silently stop firing after the 64th distinct broken font, for the rest of the process's life. There's already a page-scoped pattern for this elsewhere in the codebase (`k_zero_warning_emitted`) — could you use that instead? Also would like an end-to-end test that actually renders a synthetic PDF and asserts the warning surfaces, not just the unit-level dedup logic.

tobocop2:
Both points addressed in 96c199f8.

The process-lifetime `WARNED_FONTS` (64-entry cap, no reset) is gone. The latch now lives on the rasterizer and clears at the start of every page, the same scope as `k_zero_warning_emitted`: one warning per font per page, so bulk ingestion warns on every page a broken font paints and there is no cap to run out of.

The end-to-end test (`tests/test_glyph_drop_reporting.rs`) builds a minimal TrueType font in code whose byte-indexed cmap maps only 'A', embeds it in a two-page synthetic PDF that paints "AB", renders both pages through `PageRenderer`, and asserts the `GlyphDropped` warning surfaces exactly once per page and names the dropped code. A process-scoped latch fails the second-page assertion.

tobocop2:
Corpus validation of the current head (96c199f8, the page-scoped latch): 470,208 pages swept against a fresh base built from today's `main`. Every text and table movement attributes to #990, which this branch predates — the flagged pages lose exactly the words that #990's keep-budget fix preserves, and no page moves outside that family. The latch's own surface engages at scale: the glyph-drop warning now fires on 45,248 pages that stayed silent before, which is the once-per-font-per-page behaviour working corpus-wide. A rebase onto `main` after #990 clears the accounting.

tobocop2:

PR #1025, opened 2026-08-07
fix(tables): reuse the merger's word boundaries when building cell text

### LINKED ISSUE #1018: [Bug]: Table cell text gains spaces inside words
### Bug Description

The table cell text builder inserts spaces inside words. The flow text merger joins the same glyphs correctly, so cell text and page text disagree.

`cell_span_separator` emits a space when the gap between spans exceeds `font_size * 0.15`. The flow merger decides from per-glyph advance evidence, which the cell builder never sees.

### Steps to Reproduce

1. Fetch [`issue7454.pdf`](https://raw.githubusercontent.com/mozilla/pdf.js/master/test/pdfs/issue7454.pdf) from the [pdf.js test corpus](https://github.com/mozilla/pdf.js/tree/master/test/pdfs).
2. Extract page 0 both ways:

```rust
let doc = PdfDocument::open("issue7454.pdf")?;
let spans = doc.extract_spans(0)?;   // "Crédit < 21 500 euros" as one span
let tables = doc.extract_tables(0)?; // the same glyphs as a cell
```

### Expected Behavior

Cell text matches flow text: `Crédit < 21 500 euros`.

### Actual Behavior

The cell reads `Cré d it < 21 500 euros`. At 9 pt the threshold is 1.35 pt. The intra-word gaps are 2.04 pt and the legitimate space before `<` is 1.92 pt, so no constant separates them. Threshold tuning cannot fix this, and the function has already been retuned once after a regression.

### Additional Notes

Fix direction: give the cell builder the flow merger's advance evidence, or have it consume flow-merged spans. This also removes a duplicate-emission case downstream: when cell text is split where flow text is merged, budget matching cannot pair them and both surfaces emit.

One caution. Teaching the budget matcher to consume concatenations of split cell tokens was evaluated and rejected. It creates a new deletion path and reintroduces order-dependent nondeterminism. Fix the spacing, not the matcher.

Related work, checked before filing: distinct from #1016 (there the span and word layer itself is corrupted by bad glyph advances for CID subset fonts; here `extract_spans` is correct and only the cell builder disagrees) and from #975 (adjacent columns merged in strict table detection: table topology, not in-cell spacing). Cite both as related when filing.

tobocop2:
## Linked issue

Closes #1018

## What and why

Two code paths decide where words break. They use different evidence. They disagree.

- `cell_span_separator` adds a space when the gap between two spans is more than `font_size * 0.15`.
- The span merger decides from per-glyph advance evidence.

Some producers draw a single word as several show operations. The two paths then give different results:

| Surface | Result |
|---|---|
| `extract_spans` | `Crédit < 21 500 euros` (correct) |
| `extract_tables` | `Cré d it < 21 500 euros` (wrong) |

**No constant fixes this.** On the reported page, at 9 pt type:

| Gap | Measured | Correct action |
|---|---|---|
| Inside the word `Crédit` | 2.04 pt | Do not add a space |
| Before the `<` | 1.92 pt | Add a space |

The gap inside the word is **wider** than the real space. Raise the threshold and the word survives, but the space disappears. Lower it and the space survives, but the word splits.

`cell_span_separator` also carries one earlier retune for this same class of report. That is the second reason not to reach for a third constant.

**The fix.** The table path now consumes the merger's verdict. It does not derive the verdict again.

1. Word extraction records one fact per word: did the merger join this word to the word before it with no boundary at all?
2. That fact is source-index adjacency inside a single span. No gap threshold can recover it.
3. The table path re-glues those fragments before the detector sees them.

`extract_words` output is unchanged. Only the detector's input moves.

**Geometry keeps a veto** over the re-glue. The band is two-sided and half an em wide.

- The upper bound stops a column jump from fusing two cells into one.
- The lower bound excludes backtracks and line-wrap resets. The word merge loop already guards those with `gap < -font_size` and `delta_x < -5 * font_size`. A one-sided bound would admit every one of them.

Considered and rejected:

| Alternative | Why rejected |
|---|---|
| Retune `cell_span_separator` | Measured impossible, above. It would also be the third patch to that constant. |
| De-fuse at the word layer instead | #831 narrowed away from exactly this. Changing `extract_words` geometry, which `extract_page_tables` consumes, fabricated tables on prose pages. This change leaves `extract_words` byte-identical and moves only the table path's private input. |
| Teach the cell/flow matcher to consume concatenations | It creates a new deletion path and reintroduces order-dependent behaviour. The spacing is the defect; the matcher is not. |

## Type of change

- [x] Bug fix

## Tests

- [x] I added a test that **fails before this change and passes after** (revert-checked). For bug fixes the reproducer is a **minimal synthetic PDF built in code**. No third-party/reporter PDF is committed.
- [x] Test named by defect **class**, not an issue/PR number; no contributor/company names in code or fixtures.
- [x] `cargo test` passes, **and** the affected feature tiers: `rendering` / `fips,icc` / `ml` / binding (list which): **`python` (the `make check` tier), `ml`, `rendering`**. `ml` matters here because `cluster_chars_into_words` has two implementations gated on it, and this change sits directly on its output.

`tests/table_cell_word_seam.rs` builds a ruled grid. Its value cell is drawn as four show operations.

The fixture is calibrated so that **no constant can pass it**:

| Gap | Value | Correct action |
|---|---|---|
| Intra-word seam | 1.75 pt | Do not add a space |
| Real word space | 1.53 pt | Add a space |

One test asserts that inversion directly. The reproducer therefore cannot drift into something a threshold tweak would satisfy. A fourth test pins the band's lower bound: it draws a backward jump and asserts the fragments are not fused.

**One caveat, reported rather than hidden.** The `ml` tier has two failing lib tests:

| Test | Result | Expected |
|---|---|---|
| `layout::clustering::tests::test_pdx2_clustering_scales_on_large_input` | 2475 clusters | 2500 |
| `layout::clustering::tests::test_column_gap_threshold_is_font_relative_not_flat` | 1 cell | 3 |

**Both fail identically on unmodified `10b87f15`.** I ran them against a pristine checkout of the base commit to be sure. They exercise the `ml`-gated clustering functions directly, but they encode the *non-*`ml` implementation's grouping behaviour. This change touches neither `clustering.rs` nor those functions. Filed separately as #1026.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 documents / 470,208 pages, diffed against `main`. The breakdown by source is in the sweep section below.

- [x] Ran a per-page differential against **`main`**. No unintended regressions (no new word-fusions, over-splits, dropped spans, reversed scripts, or crashes).
- [x] Diffed against the **latest release (`v0.3.77`)**. `10b87f15` is simultaneously the tip of `main` and the `v0.3.77` tag. This is the same measurement, not a second one.
- [x] Judged with a structural metric (word-Jaccard + spacing outliers), not char-Levenshtein.

I built both sides from the same toolchain and feature set. An earlier run used a pre-built reference binary and showed a 100 % `render_hash` delta. That was a feature-set mismatch in the reference, not a code difference.

**Targeted prose check.** Within the corpus I also swept 8 documents where `main` detects **zero** tables, plus 4 arXiv math papers, 8,668 pages. I chose them to probe the failure mode #831 was narrowed for: a change to detector input that fabricates tables on prose. Result: **zero pages changed, in any column.** Table count moved on no page in either direction.

**Pages that do change.** Four, all on the table path. On every page in the corpus, including those four, `spans_hash`, `words_hash` and `chars_sig` are byte-identical. The span and word layers do not move.

Word-Jaccard on the four is 0.90–0.99. Every token that disappears is a fragment whose whole form was already present:

- `Cré` + `d` + `it` → `Crédit`
- `Am` + `ericaB…` → one token
- `Anne’` + `s` + `m` + `e` + `lon` → `Anne’s melon`

Non-whitespace character conservation is exact on two of the four pages. The other two lose precisely one copy of a duplicated segment. On those pages a split cell token could not pair with its merged flow span, so both surfaces emitted the line.

## AI assistance disclosure

- [ ] ~~written by me, not generated~~. See disclosure

## Checklist

- [ ] Docs/`CHANGELOG.md` updated if user-facing; reporters credited in the CHANGELOG (not in code). No CHANGELOG entry. That matches recent feature PRs, which leave it to the release commit.
- [x] Public-API changes considered for semver (`semver-checks` will run). None: the one new method is `pub(crate)`, and the two public `extract_words` signatures are unchanged.

## Full-corpus regression sweep

Complete: 470,208 pages over 19,151 documents diffed against `main`, with a measured per-page noise floor.

|---|---|---|
| curated public documents (arXiv, manuals, Japanese vertical texts) | 8 | [arxiv.org](https://arxiv.org) and public sources |

**Documents specifically examined:**

| Document | Page | Demonstrates |
|---|---|---|
| [issue7454.pdf](https://raw.githubusercontent.com/mozilla/pdf.js/master/test/pdfs/issue7454.pdf) (pdf.js corpus) | 0 | the reproducer: cell text now matches the flow text; the duplicate emission disappears |
| [002397.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/002.zip) (002.zip) | 32 | spacing corrected with zero character delta |
| [019233.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/019.zip) (019.zip) | 153 | duplicated segment collapses; lost characters are exactly one copy of the duplicate |
| [023521.pdf](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/023.zip) (023.zip) | 123 | same duplicate-collapse shape on an unrelated document |

**Unintended movement: none.** Two groups of pages were flagged, and neither is attributable to this change:

- Seven blocker pages are a base-side render-panic artifact. They panic under rendering on unmodified `main` and extract normally with rendering off. Tracked upstream separately.
- Two hand-read pages reproduce their differences on unmodified `main`. They are known nondeterministic pages.

**The corpus is quiet on this fix by expectation, not by omission.** Documents that carry sub-em-split cell text are rare. Engagement is therefore demonstrated two other ways:

1. The regression test fails first and passes after.
2. The four affected pages above were read directly, and each was checked for character-multiset conservation.

The sweep's role is to establish the absence of collateral change everywhere else.

## CI note

On the [first run](https://github.com/yfedoseev/pdf_oxide/actions/runs/31197068822/job/93023993082), `Docker smoke (alpine-3)` failed on a network flake, not on this change. The runner could not reach Docker Hub:

```
docker: Error response from daemon: Get "https://registry-1.docker.io/v2/":
net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

The job stopped at exit code 125, before any project code ran. Every other check passed: 198 pass, 1 fail, 36 skipped.

I re-pushed the commit to re-trigger CI. The tree is unchanged. `912a6e87` and `03372a80` have the identical tree `b76f712c`.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -16767,7 +16767,9 @@ impl PdfDocument {
         // Default: include /Artifact-tagged spans (matches pre-0.3.42
         // behavior). The spec-correct (§14.8.2.2.1) variant lives in
         // [`Self::extract_words_with_thresholds_no_artifacts`].
-        self.extract_words_inner(page_index, word_gap_threshold, profile, true)
+        Ok(self
+            .extract_words_inner(page_index, word_gap_threshold, profile, true)?
+            .0)
     }
 
     /// Same as [`Self::extract_words_with_thresholds`] but drops spans tagged
@@ -16779,7 +16781,9 @@ impl PdfDocument {
         word_gap_threshold: Option<f32>,
         profile: Option<crate::config::ExtractionProfile>,
     ) -> Result<Vec<crate::layout::Word>> {
-        self.extract_words_inner(page_index, word_gap_threshold, profile, false)
+        Ok(self
+            .extract_words_inner(page_index, word_gap_threshold, profile, false)?
+            .0)
     }
 
     fn extract_words_inner(
@@ -16788,7 +16792,7 @@ impl PdfDocument {
         word_gap_threshold: Option<f32>,
         profile: Option<crate::config::ExtractionProfile>,
         include_artifacts: bool,
-    ) -> Result<Vec<crate::layout::Word>> {
+    ) -> Result<(Vec<crate::layout::Word>, Vec<bool>)> {
         use crate::layout::{clustering, AdaptiveLayoutParams, DocumentProperties, Word};
 
         // Span source. The default (no profile) flows through the canonical
@@ -16832,7 +16836,7 @@ impl PdfDocument {
             },
         };
         if spans.is_empty() {
-            return Ok(Vec::new());
+            return Ok((Vec::new(), Vec::new()));
         }
 
         // Compute adaptive parameters from all characters for consistent thresholds.
@@ -16853,7 +16857,7 @@ impl PdfDocument {
             span_char_ranges.push(start..all_chars.len());
         }
         if all_chars.is_empty() {
-            return Ok(Vec::new());
+            return Ok((Vec::new(), Vec::new()));
         }
         let props =
             DocumentProperties::analyze(&all_chars, page_bbox).map_err(Error::LayoutAnalysis)?;
@@ -16884,6 +16888,14 @@ impl PdfDocument {
         let mut rotated_word_indices: std::collections::HashSet<usize> =
             std::collections::HashSet::new();
         let mut words = Vec::new();
+        // continues_prev[i]: the span-level merger joined word i to word i-1
+        // with no word boundary — their glyphs are consecutive in one span
+        // with not even a whitespace char between them, and only the
+        // geometric clustering below re-split them. Kept parallel to
+        // `words` through the merge loop; consumed by the table path, which
+        // must not re-decide a join the merger already made from per-glyph
+        // advance evidence.
+        let mut continues_prev: Vec<bool> = Vec::new();
         for (span_idx, span) in spans.iter().enumerate() {
             let span_chars = &all_chars[span_char_ranges[span_idx].clone()];
             if span_chars.is_empty() {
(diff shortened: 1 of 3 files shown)


yfedoseev:
This is a nice piece of reasoning — using source-glyph adjacency instead of another magic-constant threshold is exactly the right instinct, and you called that out yourself in the PR. Checked against our corpus including a golden-reference dataset — net positive, one document showed a mixed result that traces back to a metric-sensitivity quirk rather than a real regression. Merging, thank you.

Issue #1034, opened 2026-08-13
extract_cell drops rotation_degrees when re-synthesizing cell spans (tagged-PDF table path)


Goldziher:
`extract_cell` re-synthesizes cell spans and hardcodes `rotation_degrees: 0.0`, discarding the rotation it was just given. Every other field in that same struct literal is copied from the source `TextBlock`, so this reads as an oversight rather than a deliberate simplification — and unlike the other simplifications in the same function, it carries no comment.

Verified against `pdf_oxide-0.3.77` from crates.io.

### Where the rotation is preserved

`extract_table_from_spans` builds each `TextBlock` and carries the span's rotation through, `src/structure/table_extractor.rs:505`:

```rust
mcid: s.mcid,
sequence: s.sequence,
rotation_degrees: s.rotation_degrees,
```

### Where it is dropped

`extract_cell` then re-synthesizes `TextSpan`s from those blocks, `src/structure/table_extractor.rs:801-823`. Note that `bbox`, `font_name`, `font_size`, `font_weight`, `is_italic` and `mcid` all come from `block` — the value is in scope and simply not read:

```rust
cell_spans.push(TextSpan {
    text: span_text,
    bbox: block.bbox,
    font_name: block.dominant_font.clone(),
    font_size: block.avg_font_size,
    font_weight: if block.is_bold { FontWeight::Bold } else { FontWeight::Normal },
    is_italic: block.is_italic,
    mcid: block.mcid,
    // ...
    rotation_degrees: 0.0,
    // ...
});
```

`:823` is the only production occurrence. Tests begin at `:861`, so the six other `rotation_degrees: 0.0` literals in this file (`:1179` onward) are all fixtures.

### Why it matters

The two table paths disagree. `grid_to_table` in `src/structure/spatial_table_detector.rs:3900` uses `spans.get(idx).cloned()`, which preserves rotation — so a consumer that orders rotated cell text by its own advance axis works on untagged PDFs and silently does nothing on tagged ones, because by the time the spans reach it every one of them claims `rotation_degrees == 0.0`. The result is scrambled cell text in rotated tables, but only for tagged documents, which makes it look non-deterministic from the outside.

This is the mechanism behind the table half of #806 ("destroyed table/chart structure"). That issue describes the symptom; as far as I can tell nobody has pinned it to this line, so I have filed it separately rather than adding to that thread. Possibly related: #1029, #1006, #662.

### Suggested fix

```rust
[...]

PR #1021, opened 2026-08-07
fix(images): make inline-image extraction deterministic for duplicate dictionary keys

tobocop2:
## Linked issue

Closes #1017.

## What and why

**The symptom.** Some inline image dictionaries carry both forms of the same key. That is the abbreviated form and the full form: `/F` and `/Filter`, `/CS` and `/ColorSpace`, `/DP` and `/DecodeParms`. Extracting images from such a page returns different results run to run.

**The cause.** `expand_inline_image_dict` maps abbreviations to full names while it iterates a `HashMap`. Both forms therefore collide on one output key. The survivor is whichever form the iteration visited last. The hash seed randomizes that order per process.

**What that looked like.** On the duplicate-key test fixture, the same binary extracted 6, 7, or 8 images depending on the run. The same coin flip applied or skipped one image's predictor.

**The fix.** The fix drains the Table 91 abbreviations from the input before it copies the remainder. The abbreviated form therefore wins over its full twin structurally. This matches pdf.js, which reads these keys as `dict.get("F", "Filter")`.

Abbreviated values such as `/CS /RGB` and `/F /AHx` still expand in a deterministic post-pass. Their behavior is unchanged.

**Why not sort the keys.** I considered sorting and rejected it. Sorting also picks a stable winner, but an arbitrary one. Abbreviated-first matches the reference implementation for the inline-image dialect these dictionaries are written in.

## Type of change

- [x] Bug fix
- [ ] New feature (has an accepted issue: #___)
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [x] I added a test that **fails before this change and passes after** (revert-checked). `abbreviated_key_beats_its_full_twin` puts both forms of three keys in one dictionary. On unmodified `main` its outcome depends on the per-process hash seed, which is the defect. The deterministic proof is the fixture check below: 8 runs of the fixed build produce 1 distinct output, where unfixed builds produce 3.
- [x] Test named by defect **class**; no contributor/company names in code or fixtures.
- [ ] `cargo test` full suite: the affected module's tests pass (`inline_image_dict`). Feature tier: `rendering` (build verified).
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Regression on real PDFs

[...]

PR #1020, opened 2026-08-07
fix(render): skip zero-dimension image blits instead of panicking

tobocop2:
## Linked issue

Closes #1019.

## What and why

**The panic.** Rendering a page that carries a zero-dimension image panics. The blit path calls `IntSize::from_wh(blit_w, blit_h).unwrap()`. `from_wh` returns `None` when a dimension is zero. So an image declaring `/Height 0` kills the render of a page that extracts fine.

**The fix.** The fix chains the size through `and_then`. A degenerate blit then takes the existing skip path of the surrounding `if let`. That is two lines.

**The alternative I dropped.** I considered rejecting the malformed image earlier, in decoding. That would fail the whole page to guard one empty draw. The code can simply skip that draw instead.

## Type of change

- [x] Bug fix
- [ ] New feature (has an accepted issue: #___)
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

- [x] I added a test that **fails before this change and passes after** (revert-checked). The reproducer is `build_zero_height_image_pdf()`, a minimal synthetic PDF built in code. On unmodified `main` it panics with the identical backtrace to the corpus page.
- [x] Test named by defect **class** (`zero_dimension_image_blit_is_skipped`). No contributor or company names appear in code or fixtures.
- [x] `cargo test` passes (5,964 lib tests). The affected feature tier `rendering` passes too.
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source): 19,151 PDFs / ~470,000 pages. The kinds are crawled US government documents (govdocs1), academic papers, technical manuals, and Japanese vertical-writing texts. It also covers the veraPDF, pdf.js, and pdfium test corpora.

| Corpus category | Documents | Source |
|---|---|---|
| govdocs1 (crawled US .gov documents) | 14,903 | [digitalcorpora.org](https://digitalcorpora.org/corpora/file-corpora/files/) |
| curated public documents (arXiv, manuals, Japanese vertical texts) | 8 | [arxiv.org](https://arxiv.org) and public sources |
| veraPDF conformance corpus | 2,907 | [veraPDF/veraPDF-corpus](https://github.com/veraPDF/veraPDF-corpus) |
| pdf.js test corpus | 974 | [mozilla/pdf.js test/pdfs](https://github.com/mozilla/pdf.js/tree/master/test/pdfs) |
[...]

PR #1063, opened 2026-08-14
fix(tables): keep every text block a tagged cell's MCID owns

### LINKED ISSUE #1062: [Bug]: A tagged table cell keeps only its first text run
### Bug Description

A table cell owning several text runs under one MCID keeps only the first: "Hello" then "World" in one cell extracts as `Hello`, and "World" drops out of the table entirely. Once a cell can hold several runs, two more defects surface behind it: the runs join in content-stream order rather than visual order, and a cell in a rotated table joins its runs along the page axis instead of the cell's own writing axis. Split from the extraction-correctness report per review; this is its table-cell finding.

### Steps to Reproduce

1. Build a tagged PDF whose table cell's marked-content sequence owns two runs, "Hello" then "World".
2. Extract the table: the cell reads `Hello`.

### Expected Behavior

The cell keeps every block its MCID owns — `Hello World` — ordered along the cell's writing axis.

### Actual Behavior

`World` is dropped from the table output. Observed on v0.3.77.

Tagged PDF, /TD structure element with multi-run marked-content sequence; minimal reproducer built in code in the fix PR.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

> [!IMPORTANT]
> **Blocked by ~~#1008~~. Merge that one first.** The only page flagged in this PR's corpus sweep is `govdocs_005_005721.pdf` page 73, and it is flagged whether or not this change is present: `main` produces three different results for it across six runs of one build, and #1008 makes it deterministic. With #1008 in, this PR's sweep has nothing left to explain.
>
> **Blocks #989 and #1057.** Merge order: this PR, then #989, then #1057. #989 rebases onto the `table_extractor.rs` changes here; #1057 stacks on #989.

## Linked issue

Closes #1062

## What and why

### The defect

A tagged table cell kept only its first text run. Every other block that the cell's marked-content sequence owns dropped out of the table.

### The fix

Three changes, all in `extract_cell`:

1. Reduce the cell's MCID list to first occurrence. `collect_mcids` walks the cell's subtree and can name one sequence more than once, and each repeat re-emits every block that sequence owns.
2. Drop the early exit, so the cell keeps every block its sequences own.
3. Take the line-pitch threshold from the type size rather than the larger box. A block spanning several lines is taller than the pitch around it, so a box-based threshold reads a real line break as one line and the words glue.

Emission order is unchanged: the cell's sequences in marked-content order, each sequence's blocks in page order.

### Not in this PR

Ordering a sequence's runs is a separate concern and lives in #1092, against issue #1091.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

The multi-run cell tests build tagged tables in code and pin each of the three orderings; a non-finite bbox ordering pin covers the degenerate case.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** an extraction-only change; token movement between the table and flow surfaces is bookkeeping the per-page character multiset check adjudicates.

| | |
|---|---|
| pages compared | 470207 |
| blockers | 0 |
| token-loss / token-duplication pages | 2509 / 2112 |
| class-permitted movement (the fix's own surface) | 4556 |

**Triage of the token accounting.** Conservation is a per-surface multiset subtraction, so text that moves into a cell reads as loss in the flow and gain in the table. All flagged pages carry the `tables` class. The bundled PR's validation measured the same family: most pages conserve the whole-page character multiset exactly, and the rest gain characters that the dropped runs used to lose. No page loses content.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

> [!NOTE]
> **Merge order**: this PR and #989 both change `table_extractor.rs`; whichever lands second needs a small rebase.


yfedoseev:
Heads-up on two things I hit while running this branch through an independent regression sweep against current `main`. **Mostly FYI — I'll fix both myself, since the first one is a collision with my own changes in `main` and the second only became visible because of it.** No action needed from you unless you disagree with the approach.

## 1. Conflict with #1034 (my change) — would silently revert it

This branch relocates the cell `TextSpan` construction. The relocated site carries `rotation_degrees: 0.0`, which is the pre-#1034 value — your branch predates that fix. Taking this branch's side of the hunk (the natural resolution, since the refactor is obviously the newer code) silently reverts #1034, with no conflict marker and no failing test to catch it.

Resolution is one line:

```rust
-            rotation_degrees: 0.0,
+            rotation_degrees: block.rotation_degrees,
```

Worth noting these two changes are complementary, not competing — your new `cell_rotation_degrees()` is used only at the sort step (projecting blocks along the writing direction for intra-cell reading order); it never feeds the emitted span's field. So this PR fixes reading order *within* a rotated cell, #1034 fixes what rotation the emitted span *reports*. Both wanted.

## 2. RTL numeric reversal — found while validating the above

With the conflict resolved, the sweep (445 docs, text/markdown/html word-Jaccard vs `main`) flagged `pdf_benches/.../multilingual/farsi-form-rtl03/source.pdf`. Reading the diff, the intra-cell RTL sort reverses numeric runs along with the surrounding script:

| `main` | this branch |
|---|---|
| `حداقل 50%` (minimum 50%) | `حداقل 05%` (minimum 05%) |
| `47.500 تن` | `7.5004 تن` |

`grep -c '47[.,]500'` is 3 on `main` and 0 here — the figure doesn't survive anywhere in the output. On a securities-exchange admission form that changes a regulatory minimum-offering percentage, so I treated it as a blocker rather than cosmetic.

**Cause.** The sort applies a blanket reversal when any block in the cell `looks_rtl()`:

```rust
let direction = if blocks.iter().any(|b| crate::text::bidi::looks_rtl(&b.text)) { -1.0 } else { 1.0 };
```

Per UAX #9, European/Arabic numerals (class EN/AN) keep left-to-right order inside an RTL paragraph — only the surrounding script reverses. Producers often split a figure across runs (`"4"` + `"7.500"`), so reversing the block sequence uniformly reads it back as `"7.5004"`.

**What I did.** Un-reverse maximal numeric runs after the sort. To be clear about where that lives: I have *not* pushed anything to this branch — the resolution and the bidi fix sit on a local branch of mine and will land in `main` as my own follow-up commit when this merges. Nothing here changes your PR. That restores `47.500` (3/3) and `حداقل 50%`, while keeping both of this PR's genuine wins — the recovered multi-run cell text (the actual point of the PR, clearly working: a truncated cell now recovers its full run) and the corrected Persian date ordering (`04/10/1403` → `1403/10/04`), which is a real improvement over `main`.

**One residual I did not chase.** A standalone `%` block still lands before its number in one cell (`%50`). I tried widening the run predicate to absorb bare separators and it measured *worse* — it pulled a trailing `%` into a run the sort had already placed correctly, turning one wrong case into two. So I reverted to the narrower rule. Getting UAX #9 fully right in the cell sorter is a design call that belongs with you rather than something I should hand-tune from the outside, so flagging it rather than forcing it.

Also worth confirming when you get a chance: `cell_rotation_degrees()` is *modal* across the cell's blocks, so a mixed-rotation cell sorts in the majority frame while each span reports its own rotation. That matches the helper's doc comment and seems right to me, but you're better placed to say.

The data-loss fix itself is good work — it also adds substantial recovered table content on the IRS forms (+542 words html / +438 md on `IRS_Form_1120_2024`). Only the bidi handling needed adjusting.

yfedoseev:
Change of approach from my note above, so commits you didn't write don't appear here unexplained.

I said both fixes would sit on a local branch of mine and land in `main` as a follow-up after this merged. I've pushed them here instead, as two deliberately separate commits:

- **`f60a7432`** — merge of current `main` plus the one-line `rotation_degrees: block.rotation_degrees` restore. This is the #1034 collision, i.e. *my* change breaking against yours; it isn't something you could have anticipated.
- **`e8226e83`** — the UAX #9 numeric-run fix. This one is a defect in the sort this branch introduces, so it's properly yours; I've fixed it for expedience rather than bouncing it back, and kept it as its own commit so you can replace it with your own approach without untangling the collision fix.

Reason for pushing rather than following up on `main`: landing them separately means CI never tests the combination, and `main` would briefly carry both the reverted #1034 *and* the RTL numeric corruption between the two commits. That second part is why I didn't want the gap — `minimum 50%` reading as `minimum 05%` on a securities form is not something to have in `main` even briefly.

Nothing else of yours is touched, and my sweep tooling is deliberately not included. Sign-off is mine on both.

If you'd rather own either commit, say so and I'll drop it.

yfedoseev:
Correction to my note above, after checking the fix against reference implementations rather than just asserting UAX #9.

I extracted the same Farsi form with **poppler** (`pdftotext`) and **pymupdf** and compared.

**On the number — all three agree, and the fix is confirmed correct:**

| | `47.500` |
|---|---|
| `main` | `47.500` |
| this branch, before the fix | `7.5004` |
| this branch, with `e8226e83` | `47.500` |
| poppler | `47.500` |
| pymupdf | `47.500` |

Both independent parsers keep the digits intact, so the un-reversal is doing the right thing and isn't just "different from broken".

**On the `%` placement — I was wrong to flag it as a residual defect.** I wrote that "a standalone `%` block still precedes its number in one cell (`%50`)" and implied that was left-over breakage. Checking the references:

| | output |
|---|---|
| poppler | `%50` |
| pymupdf | `50 %` |
| this branch + fix | `%50` on one line, `50 %` on the other |

Poppler and pymupdf **disagree with each other**, and our output matches one or the other in each case. `%` placement relative to a number in an RTL run is a genuine logical-vs-visual-order ambiguity, not a settled correct answer — so there's nothing here worth chasing. Please disregard that part of my earlier note; I shouldn't have called it a residual without checking first.

Net: the only real defect was the digit-group reversal, and that's fixed and cross-validated.

tobocop2:
> Change of approach from my note above, so commits you didn't write don't appear here unexplained.
> 
> I said both fixes would sit on a local branch of mine and land in `main` as a follow-up after this merged. I've pushed them here instead, as two deliberately separate commits:
> 
>     * **`f60a7432`** — merge of current `main` plus the one-line `rotation_degrees: block.rotation_degrees` restore. This is the [extract_cell drops rotation_degrees when re-synthesizing cell spans (tagged-PDF table path) #1034](https://github.com/yfedoseev/pdf_oxide/issues/1034) collision, i.e. _my_ change breaking against yours; it isn't something you could have anticipated.
> 
>     * **`e8226e83`** — the UAX [fix: width calculation #9](https://github.com/yfedoseev/pdf_oxide/pull/9) numeric-run fix. This one is a defect in the sort this branch introduces, so it's properly yours; I've fixed it for expedience rather than bouncing it back, and kept it as its own commit so you can replace it with your own approach without untangling the collision fix.
> 
> 
> Reason for pushing rather than following up on `main`: landing them separately means CI never tests the combination, and `main` would briefly carry both the reverted #1034 _and_ the RTL numeric corruption between the two commits. That second part is why I didn't want the gap — `minimum 50%` reading as `minimum 05%` on a securities form is not something to have in `main` even briefly.
> 
> Nothing else of yours is touched, and my sweep tooling is deliberately not included. Sign-off is mine on both.
> 
> If you'd rather own either commit, say so and I'll drop it.

I'm cool with your approach. Feel free to resolve however you need from here on any of my PR's. I'll address any feedback

tobocop2:
Confirming the modal question, and reporting a defect the measurement turned up. It is mine, and it is fixed in `74e56fd8` on this branch.

## Modal is intended, and it almost never decides anything

I instrumented `extract_cell` to emit each cell's rotation histogram, chosen frame and resulting text, then ran it over 19,109 documents / 470,082 pages / 22,253,562 tagged cells. The census and the real vote are computed separately and compared per cell, so a drift would surface rather than be assumed away. That mismatch count is 0 corpus-wide.

| population | cells | documents |
|---|---|---|
| all tagged cells | 22,253,562 | 19,109 |
| cells drawn at more than one quadrant | 1,271 | 39 |
| of those, ties | 468 | |
| ties whose text differs from the upright frame | **0** | |
| mixed cells whose text differs from the upright frame | 683 | 14 |

Every tie emits exactly what the pre-PR upright frame emitted, so the smaller-angle tiebreak is unobservable on this corpus. Where the vote does change text, 579 cells choose a 90 degree frame and 104 choose 270. The 90 degree cases I read are clear improvements over `main`, for example `Test material 1 2 Cal. Gulch Fe/Mn PbO Jasper County low lead yard` where `main` gives `1 2 dr a ad y /Mn PbO y low le h Fe unt c Cor material t spe Tes Cal. GulJa`. The 270 cases are the defect below.

On the other half of your sentence: the reported rotation is per run and the sort frame is per cell, and I agree that is the right split. The sort produces one linear order so it needs one frame; the reported value describes one run.

## The skew fold is a real decision, not a no-op

`cell_rotation_degrees()` folds a free angle into the upright quadrant. 359 mixed cells carry a skewed block, across 6 documents, and on 5 of them the fold changes the answer. All 5 are overwhelmingly skewed cells holding one or two quadrant-snapped blocks, such as the seal on `govdocs_019_019549.pdf` page 0 (43 free angles, 1 at 90, 2 at 270). The fold keeps those upright. Without it, 2 blocks out of 46 would capture the frame. I am keeping it, but it deserves your eye.

## What the census found: the sort was backwards on /Rotate pages

`postprocess_spans` maps a span's bbox into the displayed frame but leaves `rotation_degrees` describing the frame before the turn. On a `/Rotate` page the two no longer agree, so projecting the bbox onto that angle walks a cell's runs against the direction they advance in.

`govdocs_009_009250.pdf` page 0 (`/Rotate 270`). The rendered header is above, the extracted markdown below.

govdocs_009_009250 page 0

The four runs `K`, ` (`, `1`, `0^-3 cgs units) ` are exactly the ones this branch recovers, so the recovery was landing and then being reversed. `main` drops three of them.

`govdocs_037_037934.pdf` page 47, the same shape on wrapped column headers, where `main` emits empty cells.

govdocs_037_037934 page 47

## The fix

The page rotation now reaches the cell sorter, which keeps the page frame whenever the page carries a `/Rotate`. This is the condition `map_dominant_rotation_into_reading_frame` already declines on, and for the same reason. Nothing changes on an unrotated page, and the `rotation_degrees` a cell's spans report is untouched, so #1051 is unaffected.

The regression test copies that header's four measured run geometries rather than modelling them, so with the guard removed it reproduces `0^-3 cgs units)  1  ( K` exactly.

## Before and after, on the documents that hold a rotated-frame cell

Shipped markdown over the 8 documents (600 pages), against `main` at c392b53f:

| measure | before the guard | after the guard |
|---|---|---|
| pages differing from `main` | 341 | 341 |
| pages the guard changes | | 14 |
| pages that matched `main` and no longer do | | **0** |
| `govdocs_009_009250` p0 header | `M  F` / `0^-3 cgs units)  1  ( K` | `FM` / `K (10^-3 cgs units)` |
| `govdocs_037_037934` p47 headers | `in   Unrestoredfractureenamel` | `Unrestored   fracture  in  enamel` |

The differing-page count is unchanged, so no gain from this branch is given back, and no page that agreed with `main` was moved away from it. Full suite 366 suites green, clippy clean.

## One thing I did not chase

The comment above `advance_extent` states a run's bbox width is its accumulated advance whatever the rotation. On this document it is not: the header runs measure 8.49 wide by 65.19 tall, so the advance is the height. Ordering does not depend on it, spacing does. Flagging rather than changing it.

I also reflowed `is_numeric_run` from your bidi commit, which `cargo fmt --check` was failing on.

yfedoseev:
Asking for a split here, per CONTRIBUTING.md rule 2 — *"One logical change per PR… Grab-bag PRs are unreviewable and get split."*

## Why

Issue #1062 itself bundles three separate asks, and this PR implements all three:

1. *"A cell keeps only its first text run"* — the `break` that drops every block after the first for an MCID.
2. *"the runs join in content-stream order rather than visual order"* — the geometric writing-axis sort plus the RTL direction flip.
3. *"a cell in a rotated table joins its runs along the page axis"* — the `/Rotate` frame guard in `74e56fd8`.

They have very different risk and evidence profiles, and right now they can only be accepted or rejected as a unit.

## The first one is ready on its own

Deleting the `break` is small, independently testable, and the evidence is unambiguous. On `gov_forms/IRS_Form_1120_2024.pdf` over a 445-document sweep against `main`: **99 distinct words gained, 0 lost, table row count unchanged (74 → 74), and the extracted *text* surface byte-identical.** That last part is what makes it clearly correct — the text was always being extracted, it just never reached the cell. Nothing about it depends on ordering.

I'd merge that today as its own PR.

## The other two need a design discussion first

Items 2 and 3 replace structure-tree order with order derived from geometry, and that runs into the spec:

- **ISO 32000-2 §14.8.2.5.1**: *"Content within a single marked-content sequence **shall** be in logical content order for that item of content."* The MCID iteration `main` uses is already that order.
- **ISO 32000-2 §14.8.4.8.3 NOTE**: the table algorithms *"work for languages with different intrinsic directionality of the script (such as right-to-left) **because the structure always reflects the logical content order of the table**."* Structure order is direction-safe by design; geometry is not.
- **§14.8.2.3.3** is the only reversal the spec defines, and it's the inverse of the flip: reverse the characters *within* a show string, **keep the strings in natural reading order**.
- **§14.8.2.5 NOTE 3**: *"The division into show strings has no semantic significance."* A `TextBlock` boundary is a show-string artifact, so promoting it to a bidi-run boundary is what splits `47.500` into `4` + `7.500`.

Concretely, on a Persian securities form the flip produced `47.500` → `7.5004` and `حداقل 50%` → `حداقل 05%`. I checked the rendered page: it prints `۴۷,۵۰۰` and `حداقل ۵۰٪`, so those are value changes, not cosmetics. Per UAX #9 (I2/L2) numerals keep left-to-right order inside an RTL paragraph, which this repo already implements at character level in `src/text/bidi.rs` (`reverse_rtl_keep_numbers`, shipped v0.3.65 and handling Arabic-Indic and Extended Arabic-Indic digits, not just ASCII).

There's also a divergence worth resolving before either lands: `src/document.rs` `order_mcid_spans` — the sibling tagged-structure assembler — deliberately **keeps raw order for mixed RTL+Latin** ("full UAX #9 bidi deferred"). A Persian cell containing `1403`, `47.500` and `%` is exactly that mixed case, so the two assemblers would now disagree about the same documents.

Your own commit message in `74e56fd8` is the clearest statement of the underlying tension: *"The four runs are the ones this branch recovers, so **the recovery was landing and then being reversed.**"* The recovery works; the sort is what keeps undoing it.

## Suggested split

- **PR A — "a tagged table cell keeps every run its MCID owns"**: delete the `break`, keep MCID structure order. No sort, no direction flip. Should be a very small diff.
- **PR B — cell ordering from geometry**: the writing-axis sort, the RTL handling, and the `/Rotate` frame guard together, since 3 exists to fix 2. Worth splitting #1062 to match, so B has its own issue to argue the design against the clauses above.

If you take PR A first, I'll merge it as soon as it's green.

## On my commits in this branch

`f60a7432` and `e8226e83` are mine. `f60a7432` resolves a genuine collision with #1034 and should survive into whichever PR carries the sort. `e8226e83` was a numeric-run band-aid for the flip — **please drop it**; I've since measured it as the wrong layer to fix this at, and it only handles ASCII digits where `bidi.rs` already handles Arabic-Indic. Sorry for the noise on your branch.

yfedoseev:
@tobocop2 — re-swept your latest head (`74e56fd8`) against current `main` over 445 documents. The rotation guard does what it says, and the number corruption is gone. One inversion is left.

## Fixed by your commit

| | `main` | earlier on this branch | `74e56fd8` |
|---|---|---|---|
| `47.500` | 3 | **0** | **3** ✅ |
| `7.5004` | 0 | **1** | **0** ✅ |
| `1403/09/19` | fragmented (`09/1403` + `19`) | fragmented | **whole** ✅ |

Both match the rendered page, which prints `۴۷,۵۰۰` and `۱۴۰۳/۰۹/۱۹`. Nice.

## Still inverted

`pdf_benches/corpus/A_golden/multilingual/farsi-form-rtl03/source.pdf`, line 40:

```
main:        | 50% | از توليد ساليانه يا 47.500 تن |
74e56fd8:    | %50 | از توليد ساليانه يا 47.500  تن |
```

I rendered the page to check rather than infer it from another extractor. The document prints:

> **حداقل ۵۰٪ از تولید سالیانه یا ۴۷,۵۰۰ تن**

Reading right-to-left: `حداقل` → `۵۰` → `٪`. So the percent sign follows its number, and `50%` is the correct order — `%50` is inverted. (Worth noting the page uses Persian digits `۵۰`/`۴۷,۵۰۰`; the ToUnicode maps them to ASCII on the way out, which is why an ASCII-only digit test happens to work on this file and wouldn't on one that preserves `۰-۹`.)

Same root cause as the digit reversal you just fixed: `%` and `50` are separate runs, and reversing run order moves the sign to the wrong side. Per UAX #9 the whole numeric run — including a trailing `%` (class ET adjacent to EN) — resolves to one left-to-right sequence inside an RTL paragraph, so nothing inside it should be permuted.

## Where I'd fix it

Not with another special case at the sort layer — I tried two and both were wrong, which is what convinced me the layer is wrong:

1. A percent-character list. That's a heuristic, and it only moves the problem to the next symbol (currency, units, `‰`).
2. Making the gap measurement order-independent. Principled, but it **measured zero effect** — so my diagnosis was wrong, and I reverted it.

This repo already resolves this at character level: `src/text/bidi.rs` `reverse_rtl_keep_numbers` (shipped v0.3.65, `CHANGELOG.md:396`), which handles ASCII, Arabic-Indic and Extended Arabic-Indic digits plus Arabic separators. That's the right layer — ordering is a block concern, EN/AN preservation is a character concern, and two heuristics fighting over the same permutation is how `%50` survives.

This is the concrete case for the split I asked for above: the multi-run recovery is independently correct and mergeable now (99 words gained, 0 lost, row count unchanged, text surface byte-identical), while the ordering work still has an open correctness question. Splitting lets the recovery land today and gives the ordering half its own issue to settle the design against §14.8.2.5.1 and UAX #9.

Also, please drop my `e8226e83` when you restructure — it's the ASCII-only numeric band-aid, and it shouldn't outlive the reversal it was compensating for.

tobocop2:
> **Corrected below.** The duplication and token-loss figures in the PR A section read gate counters as defects. Reading the dumps shows most of them are text moving from leaked prose into the cell that owns it. See https://github.com/yfedoseev/pdf_oxide/pull/1063#issuecomment-5338582799 for what holds and what does not.

Taking the split. `e8226e83` is dropped. `f60a7432` is no longer needed, because #1051 is in `main` and a branch cut from `main` carries that field already.

Two answers below. The first is the modal question you asked. The second is a defect I found while building PR A, and it affects the merge you planned today.

## The modal vote decides almost nothing

I instrumented `extract_cell` to emit each cell's rotation histogram, its chosen frame and its resulting text, then ran it over 19,109 documents, 470,082 pages and 22,253,562 tagged cells. The census and the real vote are computed separately and compared per cell, so a drift shows up instead of being assumed away. That mismatch count is 0.

| population | cells | documents |
|---|---|---|
| all tagged cells | 22,253,562 | 19,109 |
| cells drawn at more than one quadrant | 1,271 | 39 |
| of those, ties | 468 | |
| ties whose text differs from the upright frame | **0** | |

Every tie emits what the upright frame emits, so the smaller-angle tiebreak is unobservable on this corpus. The modal rule is a formality on all but 683 cells.

One part of it is not a formality. `cell_rotation_degrees()` folds a free angle into the upright quadrant. 359 mixed cells carry a skewed block, and on 5 of them the fold changes the answer. Each of the 5 is an overwhelmingly skewed cell holding one or two quadrant-snapped blocks, such as the seal on `govdocs_019_019549.pdf` page 0 with 43 free angles, 1 at 90 and 2 at 270. The fold keeps those upright. Without it, 2 blocks out of 46 capture the frame.

## PR A is not safe in the form we specified

I built the `break` deletion on its own, as PR A, and swept it. It is not equivalent to keeping every run.

`collect_mcids` walks a cell's subtree and does not remove repeats. On one document, 16,275 cells name a sequence more than once. With the early exit, each repeat emits one block. Without it, each repeat emits every block that sequence owns. Corpus-wide the break deletion adds 11.5 million words, and one page goes from 16,708 words to 77,100.

Reducing the MCID list to first occurrence removes that multiplier and keeps structure order:

```rust
let mut seen = std::collections::HashSet::new();
mcids.retain(|mcid| seen.insert(*mcid));
```

That version still is not clean. Swept against `main` at `51c52c50` over 470,207 pages:

| measure | result |
|---|---|
| table strict-duplication pages | **2,252**, across 380 documents |
| token-loss pages | 2,566 |
| table digit-loss pages | 15 |
| blockers | 1 |

The duplicated documents include `exotic/pdfjs/prefilled_f1040.pdf` (`cell text multiplied: ['Form×+1']`) and `govdocs_000_000028.pdf`, so this is not confined to degenerate files.

**What I believe the cause is.** MCID ownership is not exclusive. `collect_mcids` recurses into a cell's subtree, so a cell that contains other cells also claims its descendants' marked content and emits their text again. That predicts the `×+1` signature exactly. If that holds, the correct PR A is "a cell owns the marked content of its own sequences, not its descendants'", which bounds ownership rather than capping it at one block. I am testing that next and will bring the numbers.

So please do not merge the `break` deletion yet. It moves content into the right cell, and it also multiplies it.

## Marked-content order is not always reading order

This is the other half of the split, and it is the evidence for keeping an ordering step at all.

`govdocs_009_009250.pdf` page 0 is a `/Rotate 270` page whose column header reads `K (10^-3 cgs units)`. Its four runs share one MCID, and `extract_spans` returns them descending in x, which is the reverse of how the page reads.

govdocs_009_009250 page 0

`govdocs_037_037934.pdf` page 47 shows the same shape on wrapped column headers, where `main` emits empty cells.

govdocs_037_037934 page 47

§14.8.2.5.1 says content shall be in logical order, and these files do not honour it. That is the question the ordering PR has to answer: whether pdf_oxide trusts the declaration or the geometry when they disagree.

## The `%50` inversion

I agree on the layer. `reverse_rtl_keep_numbers` treats a numeric run as digits plus `. , : ٫ ٬`, so a trailing `%` stays outside the run. UAX #9 puts an ET next to an EN inside it. I will extend the character-level run in PR B and drop the block-level direction flip, so the two rules stop competing for the same permutation. That helper shipped in v0.3.65 and is used outside tables, so I will sweep it on its own rather than fold it into the ordering diff.

tobocop2:
Correction to my last comment. I read the gate counters as defects and told you to hold the merge. I then read the dumps behind them, and most of those counters are the fix working, not a regression. The "2,252 duplication pages" and "2,566 token-loss pages" figures should not have gone out in that form.

## What the flagged pages actually contain

**`exotic/pdfjs/prefilled_f1040.pdf` page 0.** Flagged as `cell text multiplied: ['Form×+1']`. The token `Form` appears 12 times in both arms. The change is where the header text lives:

```
main:  | (1) | (2) | (3) | (4) |
       prose: "Dependents: First name Last name Dependent's social security
               number Dependent's relationship to you qualifying for child
               tax credit ✓ if child under age 17 (see page 15) cForm 1040"

fix:   | (1)  First name  Last name | (2) Dependent's social security number |
       | (3) Dependent's relationship to you | (4) qualifying for child tax
       credit ✓ if child under age 17 (see page 15) |
       prose: "Dependents: Form 1040 (2010)"
```

The header text moves out of leaked prose and into the cells that own it.

**`govdocs_000_000176.pdf` page 2.** Flagged as `lost tokens: ['Inc.×8']`. `Inc.` appears 49 times in both arms.

```
main:  | Russia | Facility for the Production of Enriched | ... | Westinghouse Electric | $870,000 |
fix:   | Russia | Facility for the Production of Enriched Design and Prototype
         Qualification of a Boron | ... | Westinghouse Electric Company | $870,000 |
```

**`govdocs_005_005504.pdf` page 4.** `main` drops the project title, the duration, the ATP number, the contact name, the phone number and all three subcontractors. The fix carries them.

**`govdocs_002_002872.pdf` page 27.** `main` truncates the question. The fix completes it.

```
main:  | ¿Qué debería hacer si mi nivel de azúcar en la sangre estuviera |
fix:   | ¿Qué debería hacer si mi nivel de azúcar en la sangre estuviera  demasiado bajo? |
```

The gate compares token multisets per cell. Text that moves from prose into a cell reads as loss on one side and duplication on the other, so a single redistribution produces both counters. That is on me to read before reporting, and the harness rules say so.

## What still holds

**The `break` deletion on its own does multiply text.** That measurement was page-level, not per-cell, so it is not the same artifact. It adds 11.5 million words corpus-wide, and one page goes from 16,708 words to 77,100. `collect_mcids` names a sequence more than once, and each repeat re-emits every block that sequence owns. Reducing the list to first occurrence removes most of it.

**Some amplification survives that fix.** `govdocs_027_027798.pdf` page 229 goes from 16,708 to 62,540 markdown words, with distinct vocabulary flat at 243 against 240. So the page repeats the same small vocabulary more often. I do not yet know the cause.

**The scope is narrower than I implied.** The `nwords` column is unchanged on all 470,199 pages, so the amplification reaches markdown and HTML only, not the text layer.

**One hypothesis is dead.** I thought a cell was claiming its descendants' marked content. Measured on the flagged documents: 0 of 4,077 cells claim a descendant. So the residual has a third cause, and I am still looking for it.

## Where that leaves the merge

The recovery is doing what the PR says on the pages I read, and the earlier blanket warning was wrong. The open question is narrower: a small family of documents with degenerate structure trees repeats content more under the fix than under `main`, which already repeats it 68 times over 243 words.

I would still not merge the plain `break` deletion, because that multiplier is real and measured at page level. The version with the MCID list reduced to first occurrence is the one worth reviewing, and I will bring the cause of the residual before asking you to take it.

yfedoseev:
@tobocop2 — read the correction, and thank you for it. Going back to the dumps behind your own gate counters and retracting a hold you'd already given is exactly the right instinct, and the three examples you walked through are convincing: header text moving out of leaked prose into the cells that own it, `Westinghouse Electric` → `Westinghouse Electric Company`, and `main` dropping a project title, duration, ATP number, contact, phone and three subcontractors that your version carries. Those are improvements, not regressions, and I agree the raw counters misrepresent them.

That matches our own finding on the direction question: we're keeping your version.

Two things still stand between this and the merge queue, neither of them about correctness:

**1. It still conflicts with `main`.** As of now:

```
CONFLICT (content): Merge conflict in src/structure/table_extractor.rs
```

This is from #1034 landing in `main` (the `rotation_degrees` fix in `extract_cell`). I have a local resolution, but I'd rather not push it here the way I did on #1011 — that one was a mechanical merge, whereas this touches the logic your PR is actually about, so it should be your call how it resolves.

**2. The split.** This is +698/−144 across `src/document.rs`, `src/structure/table_extractor.rs` and the test. Our contributing guide asks for one issue per PR, and this is carrying more than one concern. Splitting it would let the uncontroversial parts land immediately instead of waiting behind the parts that need discussion — which is what happened with your `split/*` series, and that worked well.

If splitting it now is more churn than it's worth, tell me and I'll take that view seriously — you've been closer to this code than anyone. But as it stands the conflict has to be resolved by someone before it can enter the queue at all.

tobocop2:
Both points taken, and the answer to each is the same push.

## The conflict

You are right that this one should not be resolved by hand. On this branch a textually clean merge already reverted a merged fix once, which is what started the #1034 thread, so I would rather remove the conflict than resolve it.

I rebuilt the change from current `main` (`026d4590`) instead. There is no hunk to resolve, and no way for #1034's `rotation_degrees` line to be lost, because the branch starts from a tree that already has it.

## The split

The rebuilt branch carries the recovery only. It is 19 lines in `src/structure/table_extractor.rs` plus the test:

1. Reduce the cell's MCID list to first occurrence. `collect_mcids` walks the cell's subtree and can name one sequence more than once, and each repeat re-emits every block that sequence owns. Without this, deleting the early exit alone adds 11.5 million words across the corpus, and one page goes from 16,708 words to 77,100.
2. Delete the early exit, which is the fix itself.
3. Take the line-pitch threshold from the type size rather than the box.

The ordering work is out, in #1092 against its own issue #1091, as you asked.

## On the third change, since it was not in the original

Once a cell keeps every run, the existing line test glues words together. It compares `y_diff` against half of `max(prev.height, block.height)`, and a block that spans several lines is taller than the pitch around it, so a real line break reads as one line, falls through to the gap test, finds no gap, and concatenates. Measured on `govdocs_027_027798.pdf` page 229: `"of th"` + `"alarms"` emitted as `thalarms` 162 times, and `requiremeactivate` another 162. Taking the pitch from the type size instead removes all 324.

I also tried a rule that always separated two different marked-content sequences. It did not fix the gluing, because those runs share a sequence, and it broke `test_extract_cell_adjacent_mcid_spans_no_space` and `test_extract_cell_cjk_fullwidth_gap_suppresses_space`, where adjacent runs from different sequences legitimately abut. It is not in the branch.

## Evidence

370 test suites green on current `main`. The corpus sweep of this exact version is running now and I will post the numbers when it lands rather than before.

On the counters, so they do not mislead a second time: this change moves text out of leaked prose and into cells, and the gate compares token multisets per cell, so the same redistribution reads as loss on one side and duplication on the other. What I trust is the blocker count, the hand-read dumps, and the targeted measurement above.

tobocop2:
Pushed. The head is `8798301e` and the branch is mergeable again.

It is rebuilt from `026d4590` rather than merged, so there was no hunk to resolve and no way for #1034's `rotation_degrees` line to be lost. The diff is now two files:

```
src/structure/table_extractor.rs    |  21 +++-
tests/tagged_table_multirun_cell.rs | 191 +++++++++++++++++++++++++
```

I also dropped ", in reading order" from the title, since that half is now #1092 against #1091.

The corpus sweep of this exact revision is still running and I will post its numbers here when it lands.

tobocop2:
Sweep numbers for `8798301e`, both arms pinned to `026d4590` and swept in the same session.

| measure | result |
|---|---|
| pages compared | 470,207 |
| byte-identical across all measured columns | 458,525 |
| pages in only one arm | 0 |
| required documents missing from an arm | 0 |
| changed pages with no dump | 0 |
| **blockers** | **1** |

The one blocker is `govdocs_005_005721.pdf` page 73, class `[bold;images;kv]`. It carries neither `tagged` nor `tables`, so `extract_cell` never runs on it, and it appears identically in every arm I have swept including ones without this change. I read it as the standing floor rather than a finding.

## The counters, and why I am not quoting them as a verdict

The remaining movement is 2,375 token-loss pages, 2,232 token-duplication and 2,318 table strict-duplication. Those are the same redistribution you and I already went through: the change moves text out of leaked prose and into the cell that owns it, and the gate compares token multisets per cell, so one move reads as loss on one side and duplication on the other.

Spot-checking the highest-severity class again on this revision, `govdocs_010_010915.pdf` page 1 is listed under digit loss, and the dump shows a dropped column coming back:

```
main:  |  | 16,717 |  |
here:  |  | 16,717  2,909 |  |
```

## What I did measure directly

The gluing that the third change addresses, on `govdocs_027_027798.pdf` page 229:

| | main | with this change |
|---|---|---|
| `thalarms` | 0 | 0 |
| `requiremeactivate` | 0 | 0 |
| before the line-pitch fix | | 162 and 162 |

That page also recovers from 16,708 to 67,365 words, which is the truncated-cell recovery rather than duplication: its distinct vocabulary stays flat.

370 test suites green on current `main`, `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

yfedoseev:
@tobocop2 — two things.

First, the conflict is gone. `be521a01` merges cleanly against current main now, so the blocker I raised is resolved. Thank you for taking it rather than leaving it to me — you were closer to that code than I was.

Second, on the numbers: 470,207 pages with 458,525 byte-identical and one blocker you traced to a page that never reaches `extract_cell` at all is a genuinely thorough piece of verification, and the `govdocs_010_010915.pdf` spot-check showing a dropped column coming back (`16,717` → `16,717  2,909`) is the kind of evidence that settles it. Your read that the residual token-loss and duplication counters are the same redistribution — text moving out of leaked prose into the cell that owns it, which a per-cell token multiset necessarily scores as loss on one side and duplication on the other — matches what I found independently on #1012, where every apparent lost token was a fused token being split correctly.

So I have no correctness objection left.

The one open item is the split. This is still +698/−144 across `src/document.rs`, `src/structure/table_extractor.rs` and the test, carrying more than one concern, and our contributing guide asks for one per PR. I'm not going to pretend that's a correctness argument — it isn't, and your verification here is better than most things that merge. Let me come back to you on whether we take it as-is; that's a maintainer call about process, not about your work, and it shouldn't cost you another round of rebasing while I decide.

tobocop2:
Correcting one sentence in my sweep comment, and replacing it with a measurement.

I wrote that the single blocker "appears identically in every arm I have swept including ones without this change". That was an inference, not something I had measured. Every comparison I had run was this branch against `main`, so I had nothing without the change to compare against. I also called it identical when it was not: across three separate runs of this branch, the same page was flagged each time, but on a different signature.

```
run 1:  govdocs_005_005721.pdf p73 — rect_text_sig
run 2:  govdocs_005_005721.pdf p73 — rect_filt_sig
run 3:  govdocs_005_005721.pdf p73 — rect_text_sig
```

## Measured instead

I swept `main` a second time, same binary, same corpus, same settings, and compared it against its own first sweep.

| `main` vs `main` | result |
|---|---|
| pages compared | 470,207 |
| byte-identical | 470,204 |
| **flagged pages** | **1** |
| token loss / duplication / table duplication | 0 / 0 / 0 |

The flagged page is `govdocs_005_005721.pdf` page 73, with no code change at all between the two runs.

So that page is nondeterministic in `main` itself, and the shifting signature was the tell. Three pages in 470,207 differ between two runs of identical code, which is the run-to-run floor this corpus sits on. The page flagged against my branch is that floor rather than anything this PR does.

The conclusion is unchanged. The basis for it is now measured rather than assumed, and I would rather correct that here than have you take my word for a stability claim I had not tested.

tobocop2:
@yfedoseev — one merge-order note, and it resolves the loose end from my sweep.

The single page flagged against this branch, `govdocs_005_005721.pdf` page 73, is nondeterministic in `main`. Extracting it six times with one `main` build gives three different results. #1008 makes it deterministic: the same six runs on that branch give one result.

So I have marked this PR blocked by #1008 and would take that one first.

To be precise about what kind of dependency this is: it is a validation dependency, not a correctness one. Nothing in this branch depends on #1008, and merging this first would be safe. What #1008 buys is that afterwards a comparison against `main` on this corpus has no page that moves on its own, so a flagged page means the change under test did something. Right now every such comparison carries that page and an explanation for it, which is how it ended up in my sweep here and cost us both a round.

Details and the reproduction are in https://github.com/yfedoseev/pdf_oxide/pull/1008#issuecomment-5349046768.

Nothing else outstanding on this PR from my side: the conflict is gone, the ordering half is split into #1092 against #1091, and the description now matches what the diff actually does.


## One suggestion on ordering, offered gently

Looking back at this batch, I think #1008 should have gone first, ahead of the splits rather than alongside them. Every other PR here is judged by comparing output against `main`, so a baseline that moves on its own undermines all of them at once. It is the foundation the rest of the batch stands on.

The cost was real on my side. I spent a good part of today chasing a flagged page that turned out to be `main` disagreeing with itself, reported it to you as a finding before I had measured it, and had to retract that. A fair amount of the back and forth on this PR traces to the same root.

None of that is a complaint about the splits themselves. Breaking #996 up was the right call and the smaller PRs have been much easier to reason about, this one included. It is narrower than that: when a batch contains a determinism fix, that one wants to land before the work it will be used to validate, otherwise every later measurement carries a caveat.

Entirely your call on the queue, and you have context on it that I do not. I only wanted to name the pattern while it is fresh, in case it is useful for sequencing the rest.

tobocop2:

yfedoseev:
@tobocop2 — noted on the validation dependency. #1008 is now the head of the merge queue rather than sitting behind the splits, so the baseline settles before the rest of the batch is judged against it.

Nothing needed from you on this PR: the conflict is gone, the ordering half is out in #1092, and it's queued.

tobocop2:
## Corpus check

I compared this branch against `main` over 19,151 documents and 470,207 pages, with the text of every changed page kept on both sides so the comparison can be read rather than counted.

11,682 pages change, and they are the pages this fix is about: a tagged cell that owns several text blocks now keeps all of them.

## The one page I could not clear

`govdocs_005_005721.pdf` page 73 comes out differently, and it is the one page I am not attributing to this change. Running `main` against itself, unchanged, that page produces three different outputs in six runs. It moves on its own, so a before-and-after comparison cannot say anything about it either way.

Everything else that moves holds the same characters afterwards.

## Where the numbers moved

Pages flagged for possible loss or duplication were read rather than counted. What they show is cells gaining runs they used to drop: headers that used to sit in the prose above a table now sit inside it, and a cell that used to stop at its first block now carries the rest.

yfedoseev:
Re-tested this against a corner case our regression corpus does not contain, and it carries the same defect reported on #1092. Removing it from the merge queue until this is resolved.

MCIDs are scoped to the content stream that emits them (ISO 32000-1:2008 §14.7.4.3), not to the page — a page and a Form XObject drawn on it may each use MCID 0 for unrelated content.

Fixture: a page content stream tagging `CELLZERO` as MCID 0 and `CELLONE` as MCID 1, plus a Form XObject drawn on that page tagging `FORMZERO` as MCID 0 and `FORMONE` as MCID 1. One `Table → TR → two TD`, page-scoped MCRs:

```
TD children=1
  MarkedContentRef { mcid: 0, page: 0, scope: Page(0) }
TD children=1
  MarkedContentRef { mcid: 1, page: 0, scope: Page(0) }
```

Result:

```
```

The cells absorb the Form XObject's text.

Cause is the same as on #1092: `collect_mcids` drops the scope the structure tree carries —

```rust
StructChild::MarkedContentRef { mcid, .. } => {
    mcids.push(*mcid);
},
```

— and the block match compares the number alone (`table_extractor.rs:706-707`). Keeping only the first match, as `main` does, masked it: with the page stream drawn before the form the first match was usually the right block. "Keep every text block the MCID owns" is the correct intent, but without the scope in the key it also keeps blocks a different content stream owns.

`TextBlock::mcid_scope` and `StructChild::MarkedContentRef::scope` already carry what is needed, and the structure data model is documented as keyed by `(McidScope, mcid)` — the `/ActualText` applier already uses that key. Matching cells on `(scope, mcid)` closes it.

The reproduction script is in the collapsed block on #1092: https://github.com/yfedoseev/pdf_oxide/pull/1092#issuecomment-5364869949

The rest of the change measured clean on our regression corpus (445 documents from open-source sources) — this is the only blocker I found.

tobocop2:
Fixed. `collect_mcids` now carries the MCR's scope, the block match compares `(scope, mcid)`, and the span the cell synthesizes keeps `mcid_scope` instead of dropping it. `TextBlock` gained the field so the scope survives the span-to-block conversion. Your fixture is committed as `tests/tagged_cell_mcid_scope.rs` and passes; the cells come back `CELLZERO` and `CELLONE`.

yfedoseev:
@tobocop2 — I've gone through all four. Reporting what I found on each, because the state differs and it matters.

**#1063 — no change.** The last push to this branch was `2026-08-20T05:34:43Z`, about 23 hours before your comment. At the current head `89578010`, `collect_mcids` still discards the scope, `table_extractor.rs:827` still sets `mcid_scope: None`, and `tests/tagged_cell_mcid_scope.rs` is not present. The fixture still returns `"CELLZERO FORMZERO"`.

**#1092 — right approach, but it does not build.** The design is what I asked for: `collect_mcids` carrying `(McidScope, u32)`, the match comparing both, `mcid_scope` preserved on the synthesized span. But `cargo check --lib` on `f662c33c`, with nothing of mine merged in, fails:

```
error[E0609]: no field `mcid_scope` on type `&&TextBlock`   table_extractor.rs:736
error[E0609]: no field `mcid_scope` on type `&TextBlock`    table_extractor.rs:853
error[E0308]: mismatched types                              table_extractor.rs:875
error[E0507]: cannot move out of `*scope`                   table_extractor.rs:958
```

Your comment says *"`TextBlock` gained the field so the scope survives the span-to-block conversion."* It didn't — `text_block.rs` is unmodified on the branch, and `TextBlock` (`text_block.rs:777`) has `mcid` but no `mcid_scope`. The field you're reading is on `TextSpan`, a different struct in the same file. Add it to `TextBlock`, carry it through the span-to-block conversion, and the approach should work.

**#1104 — the same five errors**, since it contains #1092's commits. Your point about the stacking is right, and I'd rather it waited for #1092 than carried it under a `perf` title; retargeting the base would help.

**#1102 — this one is fine.** You declared the feature, acknowledged W4 and `50°C` as real gaps, and your framing of the W5 question is accurate: the extracted text can't decide which side is correct, and that needs rendering. That's the right answer to a genuinely open question. I'll come back on where the gate belongs.

---

Some wider context, because this pattern has a cost beyond these four PRs.

The problem is not the volume by itself. It's that changes are arriving reported as verified when they have not been run. Three instances in this cluster alone:

- #1092 and #1104 **do not compile** — four and five errors, on your own heads. `cargo check` catches every one in under a minute.
- The comment on #1092 reports *"passes; the cells come back `CELLZERO` and `CELLONE`."* The crate does not build, so no test ran and that output was never produced.
- The comment on #1063 describes `collect_mcids` carrying the scope and `TextBlock` gaining a field, on a branch not pushed in 23 hours that contains neither change.

Those are not review findings. They are facts about the branch that one command on your side would have established before the comment was written.

**Here is why this matters more than the time it costs.** #1063 was in the merge queue with a CHANGELOG entry written, having passed a full corpus sweep, a green CI run, and my review. It carried a defect that makes a tagged cell absorb text from an unrelated Form XObject. It was caught because I hand-built a PDF putting two content streams in collision — not by anything in the PR, and not by anything reported.

**A regression sweep cannot catch that class of bug.** When a change collects *more* than before, everything it wrongly picks up appears as more extracted text, and more text reads as recovered content. A word-count or Jaccard comparison cannot separate "recovered a run we were dropping" from "absorbed a run belonging to another content stream". Both are gains. A clean sweep on a broadening change is not evidence — and one of these was one merge away from shipping on exactly that basis.

This is a text-extraction library. The failure mode is silent: no crash, no error, just wrong text in someone's output that they have no way to notice. That is the standard the process has to meet.

What I'm asking for is specific and small:

1. `cargo check` before reporting anything about a branch.
2. Run the fixture and paste the actual output, not a description of it.
3. For any change that broadens what is matched or collected, build a case where the broadened lookup would pick up something it shouldn't — a colliding identifier, a duplicate key — and show that it doesn't.

Step 3 is the one that would have caught this, and it is going into `CONTRIBUTING.md` as a requirement for that class of change.

Also: four comments arrived within eleven seconds, one describing work on a branch that hadn't been pushed. One message, sent after the change builds and the fixture runs, is worth more than four sent immediately — and it avoids exactly that mix-up.

To be straightforward about where this leaves things: this cluster has taken two review rounds without a working fix. If the next round still doesn't build and pass the fixture, I'll close #1063, #1092 and #1104 and pick the fix up separately. That isn't a judgement on the diagnosis — it was correct, and the approach in #1092 is the right one. Reopening is always available once it builds and the fixture passes.

tobocop2:
You're right on all of it — I pushed the fix to `split/e3a-cell-runs`, which is attached to no PR, so this branch never moved and #1092/#1104 were stacked on an intermediate commit that predated the `TextBlock` field, hence your compile errors; I'm correcting the branches now and won't report anything further until `cargo check` and the fixture have actually run against these heads.

tobocop2:
Corrected and verified at the real heads.

```
split/e3-table-cells        9a751bd2   cargo check --all-targets: 0 errors   9024 passed, 0 failed
split/e3b-cell-reading-order 36653b9a  cargo check --all-targets: 0 errors   9031 passed, 0 failed
split/e3c-mcid-index         253802f8  cargo check --all-targets: 0 errors   9031 passed, 0 failed

test cell_takes_only_its_own_streams_marked_content ... ok
```

The fixture asserts your two cells: `CELLZERO` and `CELLONE`, with the Form XObject's sequences excluded.

`TextBlock` gained `mcid_scope` and it is carried at both span-to-block conversions; `collect_mcids` yields `(McidScope, u32)`; the match compares the pair; the cell's synthesized span keeps the scope. Adding the field surfaced eight literals across the crate that `cargo check --lib` never compiles, which is why I now run `--all-targets`.

On your three asks: `cargo check` first, paste real output, and for a broadening change build the colliding-key case — that fixture is the third one, and I will carry it as the pattern for anything that widens a lookup.

tobocop2:
Following up to your comment: 

Agreed on the mechanism, and it runs both ways. My first pass at the scope fix over-narrowed: it required the block's stream to equal the reference's, which empties a cell whenever a producer omits `/Stm`. The sweep reported 0 blockers on that too. I caught it by hand-reading the token-loss pages and separating real character loss from re-splitting, not from the counters. Your fixture is in the branch now and it is the pattern I am using for anything that widens a lookup.

On cause: mistakes happen. The compile breakage came from a wrong-branch push while splitting and restacking, not from skipping checks. The same workflow has landed #814, #817, #831, #982, #987, #989, #990, #1000, #1002, #1008, #1010, #1011, #1012, #1013, #1020, #1021, #1025, #1057, #1059 and #1067 green here, and the sweep has never been my only evidence.

> This is a text-extraction library. The failure mode is silent: no crash, no error, just wrong text in someone's output that they have no way to notice. That is the standard the process has to meet.

No argument on the risk, it is exactly why I run the corpus work at all. I think that part came out a little sharper than you meant, **since it reads as explaining the library to me**, and I have been building to that standard here including the nondeterminism fixes the queue was blocked on. The broadening point itself is genuinely good and it has changed how I check this class of change.

tobocop2:
Rebased onto current `main` (`20084ae5`) and re-measured, with both arms differing by only this PR. My earlier numbers compared against a base four merged PRs newer, which inflated everything and manufactured a regression that turned out to be #1012 missing from the old base.

```
cargo check --all-targets   0 errors
tests                       9034 passed, 0 failed
corpus                      470207 pages, 458527 byte-identical, 0 blockers
conservation                0 pages unverified
```

Of 300 sampled pages that move, 253 hold the same characters in a different arrangement. The rest were read rather than counted. `govdocs_009_009742.pdf` page 0 is representative: `main` gives `| **Percent of Total** | **Benchmark –** |` and `| African |`, this branch gives `| **Percent of Total Survey Respondents** | **Benchmark – 1998 PD&R Data** |` and `| African American/Black |`.

The colliding-stream fixture passes, and the scope rule now disambiguates rather than filters: where a block carries the stream the reference names, that stream decides; where none does, the number alone decides, so a producer that omits `/Stm` does not lose its cell.

PR #1065, opened 2026-08-14
fix(convert): stop fusing distinct blocks in markdown post-processing

### LINKED ISSUE #1064: [Bug]: Markdown post-processing fuses distinct blocks and deletes spaces in bold prose
### Bug Description

Two defects in the markdown converter's post-processing pass. A heading absorbs the paragraph after it: a 20 pt heading "Revenue" followed by a 12 pt paragraph on its own baseline emits `## Revenue 2024 was a record year.`. And bold prose loses its spaces: `**A gitHub repo**` (single upper word + CamelCase word + lower word) emits `**AgitHubrepo**`, both spaces deleted by a CamelCase-repair pass compensating for a span-assembly tear that no longer exists. Split from the extraction-correctness report per review; these are its markdown post-processing findings. The related heading-fragmentation heuristic is deliberately NOT part of this: it needs a different mechanism and stays open.

### Steps to Reproduce

1. Build a PDF with a 20 pt heading "Revenue" and a 12 pt paragraph "2024 was a record year." on separate baselines; convert to markdown.
2. Build a PDF with bold text "A gitHub repo"; convert to markdown.

### Expected Behavior

`## Revenue` plus a separate paragraph; `**A gitHub repo**` with its spaces.

### Actual Behavior

`## Revenue 2024 was a record year.` and `**AgitHubrepo**`. Observed on v0.3.77.

Minimal reproducers built in code in the fix PR.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1064

## What and why

Markdown post-processing fused distinct blocks: a heading absorbed the paragraph after it (`## Revenue 2024 was a record year.`), and bold prose lost its spaces (`**AgitHubrepo**`) to a CamelCase-repair pass compensating for a span-assembly tear that no longer exists. That pass is deleted; span assembly now prevents the tear it papered over.

The heading/value-line geometry heuristic is deliberately NOT in this PR: your corpus showed it fragments multi-line titles, so it stays held until we find a mechanism that doesn't.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

The overmerge tests pin the heading/paragraph separation and the bold-prose spaces on synthetic pages.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** a markdown-surface change; the character multiset per page is conserved and movement is confined to block boundaries.

| | |
|---|---|
| pages compared | 470207 |
| blockers | 39 |
| token-loss / token-duplication pages | 19 / 0 |
| class-permitted movement (the fix's own surface) | 21535 |

**Triage of the flagged pages.** The 39 blocked pages move only in markdown word-set and rect probes, on pages whose class markers lack the `headings` label: the deleted CamelCase-repair pass touched any bold-adjacent markdown, so its removal moves the markdown surface corpus-wide. The class-permitted bulk is the same surface on labelled pages. Text and table columns are conserved.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

### DIFF
--- a/src/pipeline/converters/markdown.rs
+++ b/src/pipeline/converters/markdown.rs
@@ -1734,7 +1734,24 @@ impl MarkdownOutputConverter {
                     // A genuinely wrapped heading is re-fused downstream by
                     // merge_consecutive_same_level_headings, which demands a
                     // continuation signal.
-                    let heading_line_break = current_heading_level.is_some();
+                    // ISO 32000-1 §14.8.3 makes one BLSE a single block that
+                    // "can be split between lines of text", and an `<Hn>` is
+                    // such a block: a heading drawn on two baselines inside one
+                    // heading element is one heading. Break on a new baseline
+                    // only when it leaves that element or changes level.
+                    //
+                    // Re-fusing downstream cannot cover this. That pass needs a
+                    // continuation signal — a trailing comma, or a lowercase
+                    // opener — and a title-cased heading capitalises every
+                    // line, so `Tax and` / `Credits` reads as two headings and
+                    // stays two.
+                    let heading_continues = current_heading_level == span_heading_level
+                        && matches!(
+                            (span.block_id, prev.block_id),
+                            (Some(a), Some(b)) if a == b
+                        );
+                    let heading_line_break =
+                        current_heading_level.is_some() && !heading_continues;
                     if is_bullet || is_ordered || starts_new_list_item || heading_line_break {
                         // Bullet on new line → flush current line and start list item
                         close_formatting(&mut current_line, &mut active_bold, &mut active_italic);
--- a/src/pipeline/converters/markdown.rs
+++ b/src/pipeline/converters/markdown.rs
@@ -333,24 +333,22 @@ fn merge_consecutive_same_level_headings(s: &str) -> String {
             j += 1;
         }
 
-        // Two policies that both prove the run is one logical heading:
-        //   A) 3+ fragments AND each <= 2 words — canonical PowerPoint
-        //      word-per-heading pattern.
-        //   B) Exactly 2 fragments AND the FIRST ends with a
-        //      continuation-strength punctuation (`,` or `;`) or no
-        //      sentence-terminator (`.`, `?`, `!`, `:`). The second
-        //      fragment must visually look like a continuation: start
-        //      lowercase or with a connector word ("and"/"or"/"the"/
-        //      "with"/"of"/...). This matches the reporter's wrapped-
-        //      heading shape `## Despite seasonal slowdown,` +
-        //      `## warehouse operations maintained...` while still
-        //      keeping `# First Heading` / `# Second Heading` apart
-        //      (no trailing comma, second word "Second" is capitalized
-        //      and not a connector).
-        let three_plus_short =
-            texts.len() >= 3 && texts.iter().all(|t| t.split_whitespace().count() <= 2);
+        // Merge only when the run carries a positive continuation signal:
+        // exactly 2 fragments where the FIRST ends with continuation-
+        // strength punctuation (`,` or `;`) or no sentence-terminator, and
+        // the second visually looks like a continuation (starts lowercase
+        // or with a connector word — "and"/"or"/"the"/"with"/"of"/...).
+        // This fuses the wrapped-heading shape `## Despite seasonal
+        // slowdown,` + `## warehouse operations maintained...` while
+        // keeping `# First Heading` / `# Second Heading` apart.
+        //
+        // Runs of 3+ short headings are deliberately NOT fused: at the
+        // string level `# Sales / # Marketing / # Engineering` (three real
+        // sections) is indistinguishable from a slide title exported one
+        // word per heading, and every reference extractor keeps such runs
+        // separate. Fusing destroyed genuinely distinct section headers.
         let wrapped_two = texts.len() == 2 && looks_like_heading_wrap(&texts[0], &texts[1]);
-        if three_plus_short || wrapped_two {
+        if wrapped_two {
             let merged = texts.join(" ");
             let hashes = "#".repeat(level);
             out.push(format!("{} {}", hashes, merged));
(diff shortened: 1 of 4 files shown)


yfedoseev:
I ran this branch through an independent regression sweep against current `main` — 445 documents (the curated 419-doc corpus plus all 88 `pdf_benches` files), comparing text/markdown/HTML output. It found a regression I think needs addressing before merge.

## The heading detection is right; the emission granularity isn't

Headings are emitted **per visual line** rather than per logical heading. Where a heading wraps across lines — which is most paper titles and most multi-word form labels — each line becomes its own heading.

On `gov_forms/IRS_Form_1040_2024.pdf`:

| `main` | this branch |
|---|---|
| `## Tax and  Credits` | `## Tax and` + `## Credits` |
| `## Payments  and  Refundable  Credits` | `## Payments` + `## and` + `## Refundable` + `## Credits` |
| `## Amount  You Owe` | `## Amount` + `## You Owe` |
| `## Third Party  Designee` | `## Third Party` + `## Designee` |
| `## Sign  Here` | `## Sign` + `## Here` |
| `## Dependents (see instructions)` | `## Dependents` (parenthetical split to its own paragraph) |
| `**Paid**  **Preparer**  **Use Only**` | `## Paid` + `## Preparer` + `**Use Only**` |

`## and` as a standalone H2 is the clearest single indication this isn't a judgement call.

Same shape on `IRS_Form_1065_2024.pdf` (`## Sign Here` → two headings), and on wrapped paper titles: `pmc/PMC8025823` turns a 4-line title into 4 separate `##`, and `olmocr .../00e2c771..._page_1_pg1.pdf` turns a 5-line title into 5 separate `#`. `pmc/PMC8094546` produces `# Diagnosis and treatment of neurogenic` + `# Society of Neurology` as two unrelated fragments.

**No content is lost** — I checked specifically (`grep -c "see instructions"` is 18 on both sides; the parenthetical is relocated, not dropped). It's structural degradation only.

## Why the scores don't show it

Worth flagging because it affected my own first pass: these diffs score **0.9905–0.9999 word-Jaccard** — the "tiny" band. By similarity score this branch reads as clean, and I initially recorded it that way on the strength of the sweep numbers. It only surfaced when I diffed the actual heading lines. Same-bag-of-words, different structure is invisible to a token-set metric.

## The detection work itself is good

To be clear about what's worth keeping: on documents that had **no** headings at all under `main`, this branch correctly finds them — `pmc/PMC8103239` now gets its title as a single `## Synergistic Effects of Dexamethasone and Dexmedetomidine`, and `pmc/PMC8129076` both detects its title *and* strips a stray page number (`## REVIEW ARTICLE 83` → `## REVIEW ARTICLE`). Two French document titles on `olmocr .../187b0c3e..._page_6.pdf` are correctly promoted to `###`. Single-line headings come out right; only wrapped ones fragment.

## Suggested direction

Merge consecutive same-level heading lines belonging to one visual block before emitting, rather than emitting one heading per source line. That should preserve every improvement above — the single-line cases are unaffected, and the wrapped cases become one heading instead of N.

Worth noting the PR is titled "markdown overmerge fixes": it corrects a real over-*merging* defect and overshoots into over-*splitting*. The fix looks close; it's the emission step that needs the grouping pass.

Happy to re-run the 445-doc sweep against an updated branch whenever you push — it takes about 15 minutes and I can report the heading deltas directly.

yfedoseev:
@tobocop2 — bumping this one so it doesn't get lost while the rest of the queue moves.

This is the only PR in the batch still carrying a confirmed regression from our side: the heading-fragmentation behaviour described in my earlier comment. Everything else you have open is either merged, queued, or waiting on a one-line fix.

No rush from my side on timing — I just want to make sure it's on your radar rather than assumed-handled. If you'd rather I take it over, say so and I will; if you're planning to push a fix, I'll hold it in the queue for you.

For context on where the rest stand:
- **#1059** — merged.
- **#1011** — queued and verified clean; thanks for the go-ahead on my carrying the #1039 conflict resolution.
- **#1012** — re-swept after your rotated-converter push and it comes out an improvement, details in my comment there.
- **#1009** — green except the DCO check; needs a sign-off on one merge commit.
- **#1063** — read your correction, replying there separately.

tobocop2:
Thanks for the bump, and sorry for the silence on this one. I am taking it, so please keep holding it rather than picking it up.

Your diagnosis matches what the branch does. The heading detection runs per visual line, so a heading that wraps becomes one heading per line. `## and` standing alone is the clearest proof that it is not a judgement call, and the wrapped paper titles are the same defect at larger scale.

The direction you suggested is the one I want: merge consecutive heading lines of the same level that belong to one visual block, then emit. The single-line cases are untouched by that, so the gains you listed should survive, including the title on `pmc/PMC8103239` and the stray page number stripped on `pmc/PMC8129076`.

Two things I want to get right rather than fast, because a heading pass that merges too eagerly is worse than one that fragments:

- what counts as one visual block, since a wrapped title and two consecutive one-line headings look similar from line geometry alone;
- what happens to the parenthetical on `Dependents (see instructions)`, which moved to its own paragraph rather than fragmenting, and so is a different failure from the wrapping one.

I will bring the fix with the heading lines diffed directly rather than a similarity score, since that is what hid this from both of us the first time. If you want it sooner than that takes, say so and I will hand it back.

For what it is worth, the offer to take it over is appreciated. This one is mine to fix.

yfedoseev:
@tobocop2 — understood, it's yours, and I'll keep holding it. Take the time.

Your two open questions are the right ones, and the second is the one I'd have missed: `Dependents (see instructions)` moving to its own paragraph is a different failure from a wrapped title fragmenting, and a merge rule tuned for the wrapping case could easily paper over it or make it worse. Worth keeping them separate in the fix rather than letting one rule try to cover both.

And yes to diffing the heading lines directly instead of scoring them. That's the same lesson from #1065 in the first place — the regression sat at 0.98+ similarity and neither of us saw it, because a score that high reads as noise. Structural diffs surface exactly the class of defect a similarity number hides.

No deadline from me. I'd rather have this right than have it soon — a heading pass that over-merges is worse than one that fragments, exactly as you said.

tobocop2:

tobocop2:
@yfedoseev — fix pushed as `55078fb8`. The head is rebased onto current `main`, so the branch is a pure delta again.

## What it was

Not the merge step. A new baseline inside a heading flushed the heading unconditionally:

```rust
let heading_line_break = current_heading_level.is_some();
```

The intent was that a wrapped heading would be re-fused afterwards, and that pass asks for a continuation signal: a trailing comma, or a lowercase opener. A title-cased heading capitalises every line, so `Tax and` / `Credits` never matched it and stayed two headings. That is why the fragments you listed are all title case, and why `## and` survived on its own: `and` does open lowercase, so it fused leftwards and left the rest stranded.

## What it is now

The break is decided from the structure rather than the string. ISO 32000-1 §14.8.3 makes one BLSE a block that "can be split between lines of text", and an `<Hn>` is such a block, so a new baseline ends the heading only when it leaves that element or changes level. Two adjacent headings are two elements, so they still separate.

The regression test builds a tagged page whose `<H2>` is drawn on two baselines and asserts one heading comes out. Before the fix it produces exactly what you reported:

```
["## Tax and", "## Credits"]
```

## Corpus

470,207 pages against `main`, same build settings on both sides.

| measure | result |
|---|---|
| byte-identical | 413,994 |
| token-loss pages | 2 |
| token-duplication pages | 0 |
| table duplication / digit loss | 1 / 0 |

I read both token-loss pages rather than trusting the counter. Neither loses content: one is a hyphenated word across a line break where the hyphen keeps a space, and the other goes the other way entirely.

`govdocs_046_046432.pdf` page 3:

```
main:  3s
       0

here:  LSU Chemistry Department
       Radical Stabilities/Reactivities
       ...
```

Two lines of noise become twenty-one lines of the real page, and the flagged `10` and `H2` are accounting inside that recovery.

371 test suites green.

One correction to the numbers above: the base for this comparison is `026d4590`, which predates #1008 landing. #1081 and #1083 are still open, so `main` can still disagree with itself on a small number of pages, and a residue in any before/after belongs to that rather than to this change. Happy to re-run against post-#1008 `main` if you would rather have the numbers on the current tip.

tobocop2:
One thing to disclose before this goes back for review, since it is a real output change and I would rather you hear it from me than find it.

On `govdocs_006_006429.pdf` page 38, a word hyphenated across a line break comes out slightly differently:

```
main:  Ken-nedy.
here:  Ken- nedy.
```

Nothing is lost, and it is one page in 470,207. The cause is that the pass this PR removes had been deleting that space as a side effect, so removing the pass exposes it.

I traced where the space comes from before assuming. It is not the line join: at the moment the two halves are appended, the line ends `Ken-` and the incoming run starts `nedy`, with no space on either side. It appears afterwards, in post-processing, which puts it outside what this PR touches.

While chasing it I found the larger version of the same thing, and filed it as #1095: `extract_text` rejoins that word correctly and gives `Kennedy`, while `to_markdown` gives `Ken-nedy` even on `main`. The crate already has `dehyphenate_line_breaks` in `src/converters/whitespace.rs`, tested for exactly these cases, and the markdown converter never calls it.

I am leaving #1095 with you rather than taking it on. It reaches a shared helper that `extract_text` and the other converters depend on, and folding that into a PR about heading fragmentation is the kind of bundling you have been unpicking all week. It also seems like a call about converter architecture that is better made by you than worked around by me. Happy to pick it up if you would rather I did.

tobocop2:
Rebased onto current `main` (`20084ae5`) and measured with both arms differing by only this PR.

```
cargo check --all-targets   0 errors
tests                       9031 passed (one timing assertion, time_ms < 10_000, fails only
                            under parallel load and passes 3/3 alone)
corpus                      470207 pages, 414020 byte-identical
conservation                0 pages unverified, 0 real character loss
```

Nothing is dropped. Every page that moves holds the same characters; only where lines break changes.

Of the 54 pages the checker flags, 35 are the intended fix: 17 lose a stray space at a hyphen (`extend- ing` becomes `extend-ing`), 8 split headings that were fused into one line, 10 do both. 7 more come back identical on a re-run, so they are nondeterminism rather than this change.

The remaining 12 are line-break changes and they run both ways. Some are clearly better, `strong  interaction` becomes `strong interaction`. Some are worse: `a field theory. While the parameters of the  theory may vary` now breaks after `the`. That is the trade this PR makes. Dropping the merge policy stops distinct headings fusing, and it also stops some genuinely wrapped prose from merging. Which error is preferable is your call, and I would rather show you the shape than average it away.

I removed a commit I had here that tried to stop a footnote joining the line above it at a hyphen. It did not fix the case it targeted, so it is now recorded on #1095 with the span geometry instead of riding along untested.

yfedoseev:
Closing this one — it is the only PR of yours being held on a measured regression rather than on process, and it should not go into v0.3.78 as it stands.

At `55078fb8` the change splits wrapped headings mid-phrase:

```
## The Role of Denosumab... with
## Giant Cell Tumour...
```

Measured across our regression corpus: **+167 headings across 59 documents**.

Worth recording how it was found, because it is the interesting part. Gold scoring showed zero movement and corpus Jaccard sat above 0.99 — every metric we normally rely on rated the change clean. Only counting headings surfaced it. A heading split in two keeps all its words, so a word-level comparison cannot see the difference.

The defect you were targeting is real: markdown post-processing does fuse distinct blocks, and it remains open. We will pick it up separately with a heading-count assertion in the regression gate, since that is the check that would have caught this.

The rest of your open pull requests are being merged. Seventeen have landed so far and the remainder are in the queue.

PR #1067, opened 2026-08-14
fix(images): unpack sub-byte samples and honor /Decode at every bit depth

### LINKED ISSUE #1066: [Bug]: Sub-byte image samples come back packed, and /Decode applies only at 1 bit
### Bug Description

Image extraction has three linked defects in sample handling. 1/2/4-bpc samples are returned packed but labelled 8-bit `Grayscale`: a 1-bit 8x8 image returns 8 packed bytes, a buffer too short to be a valid 8x8 plane (§8.9.5.2 wants one byte per component). `/Decode` is honored only at 1 bpc and for CCITT: an 8-bit gray image with `/Decode [1 0]` returns raw samples with the tonal invert ignored. And a stream shorter than its declared dimensions is refused outright, so the whole image drops instead of padding. Behind these, consumers need to know whether samples are still in the space the dictionary describes, so plates and colour-key `/Mask` still route. Split from the extraction-correctness report per review; these are its image findings, one PR because they share the unpacking path.

### Steps to Reproduce

1. Extract a 1-bit DeviceGray 8x8 image: 8 packed bytes come back, labelled 8-bit.
2. Extract an 8-bit gray image with `/Decode [1 0]`: samples `[0,1,2,...]` instead of `[255,254,253,...]`.

### Expected Behavior

64 unpacked samples `[0,255,0,255,...]`; `/Decode` applied at 1/2/4/8/16 bpc; a short stream padded (capped) rather than refused.

### Actual Behavior

Packed mislabelled samples; `/Decode` ignored above 1 bpc. Observed on v0.3.77.

Minimal reproducers built in code in the fix PR. `/Decode` remains ignored for DCT- and JPX-coded images, whose streams never reach this path.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1066

## What and why

Split from #996 per its review. Three image-sample defects share one unpacking path, so they travel together.

### Sub-byte samples now unpack

Extraction returned 1/2/4-bpc samples packed, but labelled them 8-bit `Grayscale`. A 1-bit 8x8 image returned 8 bytes, which is too short for a valid 8x8 plane. The samples now unpack to one byte per component (§8.9.5.2). The unpacked output is capped at 256 MiB, so a malformed header cannot request gigabytes.

### /Decode applies at every depth

`/Decode` applied only at 1 bpc and for CCITT. It now applies at 1, 2, 4, 8 and 16 bpc.

### Consumers can see the sample space

`PdfImage` now states whether the samples are still in the space that the dictionary entries describe. Colour-key `/Mask` and separation plates read this fact to route correctly. Indexed expansion clears the flag, and a 16-bpc reduction clears it too.

### A short stream pads instead of vanishing

A stream shorter than its declared dimensions was refused, and the whole image dropped. The stream now pads, so the image renders sheared instead of vanishing.

### /DeviceN ink counts

The code assumed four inks for every `/DeviceN` space. It now reads the ink count from the colorspace.

### A further fix that this split's own sweep found

The corpus sweep for this split found one more defect in the same class. The JBIG2 branch decodes 1-bit samples to 8-bit samples, but it did not clear `samples_are_raw`. A colour-key `/Mask` would then test decoder output against bounds from the declared 1-bit space.

Commit 2abd49d0 moves the depth rule to one place, the `effective_bpc` computation: when the stored depth differs from the declared depth, the code clears the flag. The 16-bit reduce routes through the same rule, so the existing 16-bpc contract test pins the shared guard, and a new codec path cannot forget the flag.

1,753 JBIG2 pages in the exotic corpus showed the defect. The bundled #996 arm used an older instrument, and that instrument could not see it.

### Not covered

`/Decode` remains ignored for DCT- and JPX-coded images. Those streams never reach this path.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): `rendering` (separation plates and the page renderer consume the flag).

The consumer-contract tests pin the sample space each consumer sees per bpc, `/Decode` at every depth, the padding cap, the DeviceN ink count, and the 16-bpc plate output.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** an image and render change; the image-buffer shape invariant improves on thousands of pages and worsens on none.

| | |
|---|---|
| pages compared | 470204 |
| identical in every measured column | 71299 |
| blockers | 48 |
| class-permitted movement (the fix's own surface) | 5625 |

**Triage of the 48 flagged pages.** Every flagged page carries the `images` class, and the movement is `images_sig`, `imgspace_sig`, `render_hash` and `seps_sig` together: the separation plates route differently because this PR gives them the sample-space fact they need. That is the declared consumer-contract surface, and the plate tests pin it revert-red. Token, table and text columns are zero across all 470,204 pages. The 5,625 class-permitted pages are the unpack and `/Decode` surface on image-classed pages.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).


tobocop2:

PR #1069, opened 2026-08-14
fix(encryption): unwrap /UE and /OE for AES-256 R5

### LINKED ISSUE #1068: [Bug]: AES-256 R5 files authenticate, then decrypt every stream to noise
### Bug Description

The AES-256 R5 path returns the intermediate password hash as the file key instead of unwrapping `/UE` or `/OE`, so `authenticate()` succeeds and every stream decrypts to noise: `extract_text` returns an empty string on a file whose R6 twin extracts correctly. Split from the extraction-correctness report per review; this is its encryption finding.

### Steps to Reproduce

1. Open an AES-256 R5 encrypted PDF with an empty user password.
2. `authenticate()` returns true; `extract_text` returns "".

### Expected Behavior

The real text: the file key comes from unwrapping `/UE` (or `/OE`), as the R6 path already does for the identical file.

### Actual Behavior

Authentication succeeds and the content decrypts to noise. Observed on v0.3.77.

R5 (revision 5) AES-256 encryption dictionary; minimal reproducer built in code in the fix PR.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1068

## What and why

AES-256 R5 files authenticated and then decrypted every stream to noise: the code returned the intermediate password hash as the file key. It now unwraps `/UE` or `/OE`, the way the R6 path already does for the identical file.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): `fips,icc` (the R5 unwrap runs under both crypto backends).

The regression test builds an R5-encrypted file in code, authenticates with the empty user password, and asserts the extracted text.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** scoped to R5-encrypted files; the corpus's encrypted class is the relevant population.

| | |
|---|---|
| pages compared | 470207 |
| blockers | 0 |
| token-loss / token-duplication pages | 1 / 0 |
| class-permitted movement (the fix's own surface) | 1 |

**Triage of the one flagged page.** The single token-loss page is an unencrypted rotated document. This PR changes only the AES-256 R5 key unwrap, which an unencrypted file never reaches, so the movement is structurally unattributable to this change — it is the known flicker family on rotated pages, measured on the base itself.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -317,10 +317,12 @@ crc32fast = "1.3"
 wasm-bindgen-test = "0.3"
 hex-literal = "1.1.0"
 jpeg-encoder = "0.7.0"
-# Independent spec-algorithm implementations for encryption fixtures built in
-# tests — deliberately not routed through the crate's own crypto so a fixture
-# cannot inherit the defect its test guards against. Same versions as the
-# main dependencies, so nothing new enters the build graph.
+# Independent implementations of the spec's primitives, so an encryption
+# fixture can be built without the crate's own crypto and cannot inherit the
+# defect its test guards against. Used by the R5 end-to-end fixture; the
+# unit tests round-trip through the crate's AES deliberately, since the
+# defect they pin is in key derivation, not in the cipher. Same versions as
+# the main dependencies, so nothing new enters the build graph.
 sha2 = "0.11"
 aes = "0.9"
 cbc = "0.2"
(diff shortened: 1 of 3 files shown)


tobocop2:

PR #1079, opened 2026-08-14
fix(render): reject image-mask geometry the file cannot back

### LINKED ISSUE #1078: [Bug]: Malformed image-mask geometry panics the renderer
### Bug Description

Three mask sites take document-controlled dimensions straight into arithmetic that cannot represent them, all reachable from `render_page`/`render_separations` on a file that parses cleanly. A separation `/ImageMask` with `/Width -1` becomes a near-`usize::MAX` pixel count (multiply-with-overflow panic), a positive size no stream could back sizes an allocation from the declaration (`/Width 2147483648 /Height 2147483648` asks for 2^62 bytes), and an image whose `/Mask` sub-image has `/Width 0` underflows `mw - 1` before sampling. Split from the read-path robustness report per review; these are its image-mask rows.

### Steps to Reproduce

1. Build a one-page PDF with a separation `/ImageMask` carrying `/Width -1 /Height -1`, or an image whose `/Mask` sub-image has `/Width 0`.
2. Call `render_page` or `render_separations` with the `rendering` feature.

### Expected Behavior

Rejected dimensions log a warning and the construct is skipped: a zero-size mask has no sample to test, so the base image paints opaque. No panic, no allocation sized from a declaration the file cannot back.

### Actual Behavior

`panicked at src/rendering/separation_renderer.rs:3062` (multiply with overflow) and `src/rendering/page_renderer.rs:4178` (subtract with overflow). Observed on v0.3.77.

Minimal reproducers are synthetic PDFs built in code; the fix PR carries them as regression tests.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1078

## What and why

### The defect

Three mask sites take document-controlled dimensions into arithmetic that cannot represent them. Each site is reachable from `render_page` or `render_separations` on a file that parses cleanly.

### The separation image mask

The plate painter cast `/Width` and `/Height` with `as usize`, so `-1` became a near-`usize::MAX` pixel count. A positive size that no stream could back sized an allocation from the declaration. The painter now uses `PageRenderer::image_mask_layout`, which narrows to `u32`, rejects zero, checks the pixel count, and computes the packed length that a 1-bpc stencil needs (§8.9.6.2). The length check runs before expansion, because `expand_1bpc_to_8bpc` zero-pads and padded 0-bits mean "paint".

### The /Mask sub-image

A zero-width mask underflowed `mw - 1` before sampling. A zero-size mask carries no sample to test, so the code skips the mask and the base image paints opaque.

### The stencil fallback

The fallback read its own `/Width` with `as u32`, so a negative value passed the `mw > 0` guard as a huge stride. The conversion is now `try_from`. Both mask paths also widen the scale product to `u64`.

Each rejection logs a warning and skips the construct. The code never guesses a plausible value.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): `rendering` (separation plates included).

The regression tests build each malformed mask in code: negative dimensions, a mask no stream could back, a CCITT stencil whose compressed bytes must not be expanded into fabricated ink, and a zero-width sub-image.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** a render-path guard; pixel output on well-formed files is byte-identical.

| | |
|---|---|
| pages compared | 470204 |
| blockers | 712 |
| class-permitted movement (the fix's own surface) | 235 |

**Triage of the flagged pages.** Every flagged page moves in `seps_sig` alone: the separation plates change where a mask's stream is shorter than its declared geometry. The old plate painter expanded those masks from zero-padded bytes, and padded 0-bits mean "paint", so it fabricated ink the file never specified. This PR skips such masks instead, which is the declared fix. The same family was hand-triaged in the bundled PR's validation round (946 pages, same conclusion). Token, table and text columns are zero across all 470,204 pages.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).


tobocop2:

PR #1081, opened 2026-08-14
fix(render): draw only the normal appearance stream of an annotation

### LINKED ISSUE #1080: [Bug]: An annotation without a normal appearance renders differently run to run
### Bug Description

An annotation's `/AP` is resolved as `get("N").or_else(|| values().next())`. `Object::Dictionary` is a `HashMap`, so when `/N` is absent the fallback picks whichever entry the per-process randomized iteration order visits first: an `/AP` carrying `/D` and `/R` renders the down appearance on some runs and the rollover appearance on others — same binary, same bytes, a different image roughly half the time, varying within one process. Split from the read-path robustness report per review; this is its appearance row.

### Steps to Reproduce

1. Build a one-page PDF with an annotation whose `/AP` has `/D` and `/R` but no `/N`.
2. Render the same bytes 64 times in one process and count distinct pixel buffers: you get 2. With `/N` present you get 1.

### Expected Behavior

ISO 32000-1 §12.5.5: `/N` is the normal appearance; `/D` and `/R` show only under pointer press or hover. An `/AP` without `/N` draws nothing on a static page. Sorting the map instead would freeze a specific wrong answer.

### Actual Behavior

One of two images depending on hash order, with no way for a caller to know which. Observed on v0.3.77.

Minimal reproducer is a synthetic PDF built in code; the fix PR carries it plus a `/N`-present determinism control.

### Environment

- Rust: stable 1.97

tobocop2:
## Linked issue

Closes #1080

## What and why

An annotation's `/AP` was resolved as `get("N").or_else(|| values().next())` over a `HashMap`, so an `/AP` with `/D` and `/R` but no `/N` rendered one of two images depending on per-process hash order. The fallback is dropped rather than made deterministic: per ISO 32000-1 §12.5.5, `/D` and `/R` show only under pointer press or hover, so drawing one of them on a static page was wrong in every ordering — sorting would have frozen a specific wrong answer. Annotations that carry `/N` are unaffected, which is every annotation that renders today.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): `rendering`.

The test renders identical bytes 64 times and asserts one distinct buffer equal to the annotation-free page; its control does the same with `/N` present and asserts the appearance actually draws, so the pair cannot pass vacuously.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** a render determinism change; the only movement is on pages whose annotations lack `/N`.

| | |
|---|---|
| pages compared | 470204 |
| blockers | 0 |
| class-permitted movement (the fix's own surface) | 63 |

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).


tobocop2:
Still seeing this on fresh sweeps: 7 of the 54 flagged pages on #1065 come back identical on a re-run, so they move on their own rather than from the change. Every PR I measure pays that cost again, and each one has to be re-read and discounted by hand. This and #1083 would take it to zero.

PR #1085, opened 2026-08-14
fix(structure): reject out-of-range marked-content ids instead of wrapping

### LINKED ISSUE #1084: [Bug]: An out-of-range marked-content id wraps onto a real MCID
### Bug Description

Marked-content ids are narrowed with `as u32` at two entry points: the structure tree's `/K` and `/MCID` values, and the content stream's `BDC` properties. An id of `2^32+5` wraps to MCID 5 and aliases the page's real MCID 5, so two marked-content references point at one id and the page's reading order silently duplicates that entry. A negative id wraps the same way. Split from the read-path robustness report per review; this is its MCID row, covering both entry points.

### Steps to Reproduce

1. Build a tagged PDF whose structure tree carries `/K 4294967301` alongside a genuine MCID 5 on the same page.
2. Read back the page's marked-content ids: `[5, 5]` instead of `[5]`. The same happens for a `BDC` with the out-of-range id in its properties.

### Expected Behavior

An out-of-range id is skipped, never truncated onto a valid one.

### Actual Behavior

`[5, 5]` — the malformed id lands on the real MCID 5. Observed on v0.3.77 at both entry points.

Minimal reproducers are synthetic tagged PDFs built in code; the fix PR carries them for both entry points.

### Environment

- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1084

## What and why

Marked-content ids were narrowed with `as u32` at two entry points — the structure tree's `/K`/`/MCID` values and the content stream's `BDC` properties — so an id of `2^32+5` wrapped onto the page's real MCID 5 and aliased it, duplicating that entry in the page's reading order. Both entry points now route through one checked narrowing that skips an out-of-range id rather than guessing.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

      `rendering` / `fips,icc` / `ml` / binding (list which): none beyond default; the render tier passes untouched.

The range tests pin skip-not-wrap for `/K`, MCR and `BDC` ids, negative and above-`u32::MAX`, with an in-range control.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** scoped to tagged PDFs carrying out-of-range ids; well-formed ids are unaffected.

| | |
|---|---|
| pages compared | 470207 |
| blockers | 0 |
| token-loss / token-duplication pages | 1 / 0 |
| class-permitted movement (the fix's own surface) | 1 |

**Triage of the one flagged page.** 470,203 of 470,207 pages are byte-identical. The single token-loss page is a rotated document that a second, unrelated arm in the same round flags with the identical movement — a measured flicker page on the base, not this change. Out-of-range ids do not occur in the crawled corpus, so the fix's own surface is the synthetic fixtures, where the engagement tests pin it.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

### DIFF
--- a/src/extractors/text.rs
+++ b/src/extractors/text.rs
@@ -6682,9 +6682,13 @@ impl<'doc> TextExtractor<'doc> {
 
                 if let Some(props_dict) = self.resolve_bdc_properties(&properties) {
                     if let Some(mcid_obj) = props_dict.get("MCID") {
-                        if let Some(mcid) = mcid_obj.as_integer() {
-                            own_mcid = Some(mcid as u32);
-                            self.current_mcid = Some(mcid as u32);
+                        // Same id-space contract as the structure-tree side:
+                        // wrapping would alias a malformed id onto a real MCID.
+                        if let Some(mcid) =
+                            mcid_obj.as_integer().and_then(crate::structure::checked_mcid)
+                        {
+                            own_mcid = Some(mcid);
+                            self.current_mcid = Some(mcid);
                             log::debug!("Entered marked content with MCID: {}", mcid);
                         }
                     }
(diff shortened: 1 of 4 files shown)


tobocop2:

PR #1087, opened 2026-08-14
fix(extract): break right-aligned column stacks instead of fusing figures

### LINKED ISSUE #1086: [Bug]: Text extraction fuses right-aligned figures from adjacent table rows
### Bug Description

`extract_text` fuses figures from adjacent rows of a right-aligned numeric column into one token: `13,286` and `753` become `13,286753`.

The span joiner decides whether two spans share a line from local gap and delta-x geometry, and every arm that emits a separator keys off the spans' LEFT edges. A right-aligned column aligns on its RIGHT edge, so when the next row's figure is shorter it starts further right by the width difference, landing in the one band no arm tests. Nothing is emitted and consecutive rows concatenate.

This joiner has been patched at least seven times, each time for the same class — two spans from different lines fused with no separator — and each patch fixed the page that motivated it and missed the next layout. The codebase already owns line grouping in three places (`layout_lines::group_spans_into_lines`, `extract_text_lines`, `ordered_span::group_into_lines`); the plain-text joiner is the only surface that re-derives line membership. The durable fix is to consume an existing grouper; the split-out PR carries the narrow right-edge arm with its tolerance derived from the run's own glyph advance rather than a point constant, as the immediate stop for the data loss.

### Steps to Reproduce

1. Fetch `014094.pdf` from the public-domain govdocs1 corpus ([014.zip](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/014.zip)), treated as a specification only.
2. Extract page 16 (index 15) on current `main`:

```rust
let doc = PdfDocument::open("014094.pdf")?;
let text = doc.extract_text(15)?;
assert!(!text.contains("13,286753"));
```

### Expected Behavior

Figures from different rows stay separate. `extract_text_lines` already reads this page correctly.

### Actual Behavior

`13,286753`, `22,7963,052`, `23,1309,844` on current `main`. The fusion surfaced after the table keep-budget change landed, which is what lets the value column reach the text flow.

Helvetica 6 pt, 6 pt row pitch, column right edge ~582 pt; right-aligned numeric columns whose successor rows are shorter. Reproducible with a synthetic two-stack fixture, which the fix PR carries.

### Environment

- pdf_oxide: current `main`
- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

## Linked issue

Closes #1086

## What and why

Split from #982 per its review, with the tolerance re-derived.

### The defect

A right-aligned numeric column can fuse figures from consecutive rows: `13,286` + `753` extracts as `13,286753`. Every separator-emitting arm of the span joiner keys off left edges. A right-aligned column aligns on its right edge, so a shorter successor starts further right by the width difference. No arm tests that band, so nothing separates the rows.

### The fix

This adds the right-edge twin of the existing left-edge arm. It sits last in the ladder, so it can only add a separator where the ladder emits nothing today.

### The derived tolerance

The review flagged the old 0.5 pt constant as tuned to one measurement. The tolerance is now a twelfth of the font size. That is the measured calibration made font-relative: the reporting column right-aligns its negative rows 0.062 em left of its positive ones, and fs/12 (0.083 em) covers that with margin. At the 6 pt calibration size the band reproduces the validated 0.5 pt behaviour exactly, and the same column doubled to 12 pt stays inside it, which a point constant fails.

The corpus rejected two wider derivations before this one. Half a mean glyph advance separated twenty times the measured column population, because unrelated ragged right edges often coincide within a quarter em across a backward jump. A tenth of the font size still dragged in table-duplication pages the point band never touched. The sweep for the final band is in the summary below.

### The repro is on current main

The fusion reproduces on `main` today, on page 16 of the reporting document. The table keep-budget change (#990) lets the value column reach the text flow, which exposes the fusion.

### The longer-term direction

The linked issue records it: this joiner re-derives line membership, and the codebase already owns three line groupers. The durable fix consumes one of those. This arm is the narrow stop for the data loss.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

Three tests: the stack separates instead of gluing, the measured 0.371 pt negative-row offset stays inside tolerance, and the identical column at twice the size behaves the same — the last is revert-red against any point-constant tolerance.

## Regression on real PDFs

|---|---|---|

      scripts, or crashes).
      char-Levenshtein.

**Diff summary vs `main`:** an extraction-only change; glyph characters are identical on every changed page and nearly all movement is whitespace-only.

| | |
|---|---|
| pages compared | 470207 |
| blockers | 4624 |
| token-loss / token-duplication pages | 32 / 37 |
| class-permitted movement (the fix's own surface) | 5951 |

**Triage of the flagged pages.** The separator lands on ~10,800 pages at full-corpus scale, in line with the validated point-band's footprint on its 50k-page sample. The flagged columns are the whitespace-sensitive ones (`text_hash`, `nchars`, the rect probes), never the alnum hashes, and a 25-page sample of the blocked pages confirms it: the non-whitespace character multiset is identical on every one. The 32 token-loss pages are the newline path's known hyphen consumption ("year-round" joins to "yearround"); the table accounting rows are the same words counted on a different surface. No page loses content.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -6483,6 +6483,38 @@ impl PdfDocument {
                             // -1.75 pt and -12.75 pt sit alongside
                             // delta_x values of 56 pt and 78 pt.
                             text.push(' ');
+                        } else if y_diff > 1.0
+                            && gap < -fs
+                            && (span_end_x - prev_end_x).abs()
+                                <= 0.5
+                                    * Self::mean_glyph_advance(prev)
+                                        .min(Self::mean_glyph_advance(span))
+                            && !prev.rtl_draw_logical
+                            && !span.rtl_draw_logical
+                            && !crate::text::bidi::looks_rtl(&prev.text)
+                            && !crate::text::bidi::looks_rtl(&span.text)
+                        {
+                            // Right-aligned column stack. The backtracking arm
+                            // above only reaches stacks that share a LEFT edge
+                            // (delta_x ≈ 0); a right-aligned numeric column
+                            // shares its RIGHT edge instead, so a shorter
+                            // successor starts further right by the width
+                            // difference — the one band no arm above tests. A
+                            // row pitch under `same_line_threshold` then fused
+                            // figures from consecutive rows ("22,796" + "3,052"
+                            // → "22,7963,052"). Coincident right edges plus a
+                            // full-em backward gap at a real baseline offset is
+                            // the stack signature. The right-edge tolerance is
+                            // half the narrower run's mean glyph advance:
+                            // column alignment is quantized in advances, so an
+                            // offset under half an advance cannot place the
+                            // edge at a different column position, and the
+                            // bound scales with the font instead of being a
+                            // point constant. Placed last, so it can only add a
+                            // separator where none is emitted today.
+                            if !hangul_midword_wrap && !text.ends_with('\n') {
+                                text.push('\n');
+                            }
                         }
                     } else if y_diff > 2.0
                         && gap > FORWARD_GAP_K * prev.font_size.max(span.font_size).max(1.0)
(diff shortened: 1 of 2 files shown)


tobocop2:

PR #1092, opened 2026-08-19
fix(tables): give a tagged cell's runs a reading order

### LINKED ISSUE #1091: [Bug]: A tagged table cell's runs need a reading order; marked-content order reverses some producers
### Bug Description

A tagged table cell emits its runs in marked-content order. On producers that draw a cell's runs in a different order from the order they read, the cell text comes out reversed or interleaved.

ISO 32000-2 §14.8.2.5.1 says content within a marked-content sequence shall be in logical content order, and §14.8.4.8.3 NOTE relies on that to keep the table algorithms direction-safe. Some producers do not honour it. On a `/Rotate 270` page the four runs of one sequence arrive in descending x, which is the reverse of how the page reads.

This is separate from the cell keeping every run it owns (#1062). Keeping the runs is what makes the order observable: while only the first run survives, the order never shows.

Two shapes need one rule:

- **Rotated cells.** A run's bbox is mapped into the displayed frame while `rotation_degrees` still describes the frame before the page's `/Rotate`, so any ordering derived from the reported angle walks the runs against the direction they advance in.
- **Right-to-left cells.** Reversing a cell's block sequence also reverses numeric runs, which UAX #9 (I2/L2) keeps left to right inside an RTL paragraph. `src/text/bidi.rs` already resolves this at character level in `reverse_rtl_keep_numbers`, so the block level and the character level must not both permute the same text.

`order_mcid_spans` in `src/document.rs`, the sibling tagged-structure assembler, keeps raw order for mixed RTL and Latin and defers full UAX #9. Whatever rule the cell assembler adopts should agree with it, or the two surfaces will disagree about the same document.

### Steps to Reproduce

1. Fetch `009250.pdf` from the public-domain govdocs1 corpus ([009.zip](https://digitalcorpora.s3.amazonaws.com/corpora/files/govdocs1/zipfiles/009.zip)), treated as a specification only.
2. Read the runs of the column header on page 0 (`/Rotate 270`), whose rendered text is `K (10^-3 cgs units)`:

```rust
let doc = PdfDocument::open("009250.pdf")?;
for span in doc.extract_spans(0)? {
    if span.mcid == Some(7) {
        println!("{:>7.2} {:?}", span.bbox.x, span.text);
    }
}
```

### Expected Behavior

A cell reads the way the page displays it: `K (10^-3 cgs units)`.

### Actual Behavior

The four runs share MCID 7 and arrive in this order:

```
 475.07 "0^-3 cgs units) "
 470.18 "1"
 462.64 " ("
 457.10 "K"
```

Descending x, so marked-content order is the reverse of reading order here. A cell that keeps every run and emits it in that order reads `0^-3 cgs units)  1  ( K`. On current `main` the cell keeps only the first run, so it reads `0^-3 cgs units)` and the rest leaves the table as prose.

The same page shows the effect on a second cell: `F` and `M ` are separate runs of one sequence, and the header is `FM`.

`govdocs_037_037934.pdf` page 47 (`/Rotate 270`) has the same shape on wrapped column headers, where `main` emits empty cells and the header text leaks into prose.

`/Rotate 270` page, 612x792 media box. The column header is one marked-content sequence drawn as four show strings, each 8.49 wide by its own advance tall, sharing y and stepping along +x. Reproducible synthetically: four spans on one y, ascending x, each reporting -90, on a page whose `/Rotate` is 270.

### Environment

- pdf_oxide: current `main`
- Rust: stable 1.97

tobocop2:
> [!IMPORTANT]

> **Stacked on #1063 — merge that one first.**
>
> This branch carries #1063's changes as its baseline (keep every run a sequence owns, take the first mention of each sequence, take the line pitch from the type size), because the ordering fix is only observable once a cell keeps more than one run. The diff below therefore includes them, and the overlap is deliberate, not a conflict.
>
> Once #1063 lands I rebase onto it and drop the baseline commit, and this PR reduces to the ordering commit plus its test.

## Linked issue

Closes #1091

## What and why

A tagged table cell assembles the runs of a marked-content sequence in marked-content order. ISO 32000-2 §14.8.2.5.1 asks a producer to keep that content in logical order and §14.8.4.8.3 NOTE builds the table algorithms on it, so that order is normally the right one. Some producers do not honour it. On a `/Rotate 270` page a column header is drawn from the right end of its line back to the left, so a header that reads `K (10^-3 cgs units)` comes out as `0^-3 cgs units) 1 (K`.

### The rule

Sort the runs of one sequence into row-aware reading order when the sequence carries no right-to-left character. Leave the order the structure tree declares *between* sequences alone: that order is the producer's stated logical order, and geometry never overrides it. The sort is stable, so runs the comparator calls equal keep marked-content order.

This is the rule `order_mcid_spans`, the sibling assembler for tagged prose, already applies to a sequence's spans, so the two surfaces read the same document the same way. It reuses that path's comparator (`row_aware_span_cmp`) rather than adding a fourth ordering helper.

### Right-to-left text keeps marked-content order

UAX #9 (I2/L2) keeps a numeric run left to right inside a right-to-left paragraph, and `reverse_rtl_keep_numbers` resolves that at character level. A second permutation of the same text at run level fights it and turns `50%` into `%50`, so a sequence holding any right-to-left character is left alone. `order_mcid_spans` keeps raw order where right-to-left and Latin mix for the same reason. A test pins this: with the guard removed, the right-to-left cell starts at the number.

### The order is taken only in an upright frame

Ordering by box assumes the boxes describe an upright reading frame. Two shapes break that assumption, and on both the order must be left alone:

A page carrying a `/Rotate` is displayed turned. A run's box is mapped into the displayed frame while `rotation_degrees` still reports the frame before the turn, so an order taken there walks the runs against the direction they advance in. This is the condition `map_dominant_rotation_into_reading_frame` also declines on. `extract_table_on_page` takes the page rotation and threads it to the cell.

Text turned by its own text matrix advances along an axis a row-and-column order does not follow. A page can carry no `/Rotate` and still hold turned text, so the page rotation alone does not decide it. A run reporting a turned frame stops the order.

Both were found by corpus comparison, not by reasoning: without the first, a heading cell on a rotated page reversed; without the second, the header of `govdocs_037_037667.pdf` page 64 reversed, and that page carries no `/Rotate`.

### Considered and rejected

Ordering across sequences as well as within them lets geometry overrule the one order the structure tree states outright, and the pages that show the defect do not need it: every cell there holds a single sequence. Resolving the order at character level cannot reach it either, because the runs are separate show strings whose order is the question, and §14.8.2.5 NOTE 3 says that division carries no meaning.

### Commits

This branch now holds the ordering commit and its guard. The separate `reverse_rtl_keep_numbers` fix that rode along has moved to #1102, which branches off `main` and merges independently of this one.

## Type of change

- [x] Bug fix
- [ ] Performance
- [ ] Refactor / internal
- [ ] Docs / CI / chore
- [ ] Breaking change

## Tests

`cargo test --lib --tests` is green (369 test binaries, 0 failures) and `RUSTFLAGS="-D warnings" cargo build --tests --features rendering` builds clean. The `rendering` tier is built, not run, so the tier box stays unticked until CI reports it.

The reproducers are hand-built pages holding one tagged table each. The four runs of a sequence abut, so no gap separates them and the cell text is decided by their order alone.

One page is upright and hands its runs over in the reverse of the order it reads: that is the defect this orders out. Reverted to the baseline commit, the test fails with:

```
  left: "0^-3 cgs units) 1 (K"
 right: "K (10^-3 cgs units)"
```

Two more pages pin the guards: a `/Rotate 270` page, and a page carrying no `/Rotate` whose text matrix is `[-1 0 0 -1]`. Both must keep marked-content order. The earlier version of these tests passed a page rotation of 0 for a `/Rotate 270` fixture, so the guard never ran; they now pass the rotation the page reports.

## Regression on real PDFs

**Corpus I tested** (count + kinds + source):

- [ ] Ran `corpus_sig` and diffed my branch against **`main`** — no unintended
      scripts, or crashes).
- [ ] Diffed my branch against the **latest release** (`v0.3.__`) — same.
      char-Levenshtein.

**Diff summary vs `main`:**

**Diff summary vs latest release:**

The corpus sweep is still to run; the boxes stay unticked until it is.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

- [ ] One logical change (no bundled refactor/perf/correctness).
      CHANGELOG (not in code).
- [x] Public-API changes considered for semver (`semver-checks` will run). No public API changes: the ordering helper is private and no signature moves.

No `CHANGELOG.md` hunk here, since the maintainer writes those entries. The one-logical-change box stays unticked while the baseline and the sign fix ride along; say the word and either one moves out.


tobocop2:

tobocop2:
## Corpus check

I compared this branch against its parent branch over 19,151 documents and 470,207 pages. 468,628 pages come back byte-identical. 1,579 pages move, and every one of them carries a tagged table.

## The pages that move keep their text

262 of the moved pages were flagged for possible text loss or duplication. On 261 of them the page holds exactly the same characters in a different arrangement, which is what reordering a cell's runs is supposed to do.

One page holds fewer characters afterwards: `govdocs_018_018974.pdf` page 111. Reading it, the change is a repair. Before, a cell read `The additional conditions precedent that must be Tribe's obligation to execute the MOA: following are two satisfied to trigger the`. After, it reads `The following are two additional conditions precedent that must be satisfied to trigger the Tribe's obligation to execute the MOA:`. The characters that disappear are a trailing line that repeated fragments already present in other cells.

## Rotated pages are left alone

Pages carrying a `/Rotate`: 35,558 seen, 0 changed.

## A regression this found, and the fix for it

An earlier version of this branch reversed a header on `govdocs_037_037667.pdf` page 64, turning `Table 1 2 - Corrective Action Plans Resulting From Accident And Unacceptable Hazardous Conditions Investigations` into the same words backwards. That page carries no `/Rotate`: its text is turned by its own text matrix, so the page rotation alone did not catch it.

Ordering now runs only where the boxes describe an upright frame. That page is byte-identical to the parent branch again, and the movement on pages holding turned text drops from 114 to 68.

## Tests

9,030 tests pass with the `rendering` feature, no failures.

The guards also needed better tests. The right-to-left case was covered through a whole tagged page, but in an upright frame the line builder hands the runs over left to right before a cell sees them, so that test passed on the order upstream chose rather than on the guard. Each guard is now pinned directly on the function that holds it.

## Split out

The `reverse_rtl_keep_numbers` fix that was riding along has moved to #1102, which branches off `main` and merges independently of this one.

yfedoseev:
Verified this against a hand-built tagged PDF. It introduces a regression, so reporting here rather than changing your branch.

### What happens

MCIDs are scoped to the content stream that emits them (ISO 32000-1:2008 §14.7.4.3), not to the page. A page and a Form XObject drawn on it may each use MCID 0 for unrelated content.

Fixture: a page content stream tagging `CELLZERO` as MCID 0 and `CELLONE` as MCID 1, plus a Form XObject drawn on that page tagging `FORMZERO` as MCID 0 and `FORMONE` as MCID 1. The structure tree holds one `Table → TR → two TD`, with page-scoped MCRs:

```
Table children=1
  TR children=2
    TD children=1
      MarkedContentRef { mcid: 0, page: 0, scope: Page(0) }
    TD children=1
      MarkedContentRef { mcid: 1, page: 0, scope: Page(0) }
```

`extract_table_from_spans` on that element:

```
```

The cells absorb the Form XObject's text. #1104 contains this commit and behaves identically.

### Why

`collect_mcids` discards the scope the structure tree carries:

```rust
StructChild::MarkedContentRef { mcid, .. } => {
    mcids.push(*mcid);
},
```

and the block match compares the number alone (`table_extractor.rs:706-707`). That was survivable before: the loop took the first matching block and broke, and with the page stream drawn before the form the first match was usually the right one. Collecting every block that shares the number turns a latent scope-blindness into text from another content stream being concatenated into the cell.

### The pieces are already in the crate

`TextBlock::mcid_scope` exists, and its doc comment states the rule:

> The struct-tree `/ActualText` applier keys lookups by `(mcid_scope, mcid)` so two Form XObjects on the same page that each carry MCID 0 do not collide and overwrite each other's replacements.

`StructChild::MarkedContentRef` carries `scope`, and the structure data model is documented as keyed by `(McidScope, mcid)`. Carrying the scope out of `collect_mcids` and matching on `(scope, mcid)` would close it. Note `extract_cell` also sets `mcid_scope: None` on the span it synthesizes, so the scope is dropped again downstream.

The rest of the change measured clean on our regression corpus — the 11 surfaces that moved were all text *gains*, no losses — so this is the one blocker I found.

<details>
<summary>Reproduction (generates the PDF)</summary>

```python
objs = {}
page_content = b"""BT /F1 12 Tf 1 0 0 1 60 700 Tm
/P <</MCID 0>> BDC (CELLZERO) Tj EMC
1 0 0 1 200 700 Tm
/P <</MCID 1>> BDC (CELLONE) Tj EMC
ET
q 1 0 0 1 0 0 cm /Fm1 Do Q
"""
form_content = b"""BT /F1 12 Tf 1 0 0 1 60 400 Tm
/P <</MCID 0>> BDC (FORMZERO) Tj EMC
1 0 0 1 200 400 Tm
/P <</MCID 1>> BDC (FORMONE) Tj EMC
ET
"""
objs[1] = b"<< /Type /Catalog /Pages 2 0 R /StructTreeRoot 10 0 R /MarkInfo << /Marked true >> >>"
objs[2] = b"<< /Type /Pages /Kids [3 0 R] /Count 1 >>"
objs[3] = (b"<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792] /Contents 4 0 R "
           b"/Resources << /Font << /F1 5 0 R >> /XObject << /Fm1 6 0 R >> >> /StructParents 0 >>")
objs[4] = b"<< /Length %d >>\nstream\n" % len(page_content) + page_content + b"endstream"
objs[5] = b"<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica >>"
objs[6] = (b"<< /Type /XObject /Subtype /Form /BBox [0 0 612 792] /Length %d "
           b"/Resources << /Font << /F1 5 0 R >> >> >>\nstream\n" % len(form_content)
           + form_content + b"endstream")
objs[10] = b"<< /Type /StructTreeRoot /K [11 0 R] /ParentTree 20 0 R >>"
objs[11] = b"<< /Type /StructElem /S /Table /P 10 0 R /Pg 3 0 R /K [12 0 R] >>"
objs[12] = b"<< /Type /StructElem /S /TR /P 11 0 R /Pg 3 0 R /K [13 0 R 14 0 R] >>"
objs[13] = b"<< /Type /StructElem /S /TD /P 12 0 R /Pg 3 0 R /K [<< /Type /MCR /Pg 3 0 R /MCID 0 >>] >>"
objs[14] = b"<< /Type /StructElem /S /TD /P 12 0 R /Pg 3 0 R /K [<< /Type /MCR /Pg 3 0 R /MCID 1 >>] >>"
objs[20] = b"<< /Nums [0 [13 0 R 14 0 R]] >>"

order = sorted(objs)
out = bytearray(b"%PDF-1.7\n")
offsets = {}
for n in order:
    offsets[n] = len(out)
    out += b"%d 0 obj\n" % n + objs[n] + b"\nendobj\n"
xref = len(out)
maxn = max(order) + 1
out += b"xref\n0 %d\n" % maxn + b"0000000000 65535 f \n"
for i in range(1, maxn):
    out += (b"%010d 00000 n \n" % offsets[i]) if i in offsets else b"0000000000 65535 f \n"
out += b"trailer\n<< /Size %d /Root 1 0 R >>\nstartxref\n%d\n%%%%EOF\n" % (maxn, xref)
open("mcid_scope_collision.pdf", "wb").write(bytes(out))
```

</details>

tobocop2:
Fixed in #1063 and this branch is rebased onto it: the scope travels with the number, so a cell takes only its own stream's sequences. Your fixture is a test there.

yfedoseev:
`cargo check --lib` fails on `f662c33c` with four errors — `TextBlock` has no `mcid_scope` field (`text_block.rs:777`); the field you're reading is on `TextSpan`. Full detail and the other three PRs: https://github.com/yfedoseev/pdf_oxide/pull/1063#issuecomment-5365931824

The approach here is right — adding the field to `TextBlock` and carrying it through the span-to-block conversion should be all it needs.

tobocop2:
Fixed and verified at `36653b9a`: `cargo check --all-targets` clean, 9031 passed / 0 failed, `cell_takes_only_its_own_streams_marked_content ... ok`. Detail on #1063.

tobocop2:
Rebased onto current `main` and re-measured against #1063 alone: `cargo check --all-targets` clean, 9041 passed / 0 failed, 470207 pages compared, 468624 byte-identical, 0 blockers, conservation verified on every changed page. Of 134 pages flagged for loss, 133 hold the same characters rearranged; the one exception is `govdocs_018_018974.pdf` page 111, where this branch repairs a scrambled cell and drops a trailing line that repeated fragments from other cells.

PR #1097, opened 2026-08-20
fix: treat content-stream s as close-and-stroke

### LINKED ISSUE #1096: [Bug]: `s` is not a known content-stream operator, so a closed-and-stroked path is neither painted nor ended
## Summary

`build_operator` (`src/content/parser.rs`) maps every path-painting operator in
ISO 32000-1 Table 60 except one. `S`, `f`, `F`, `f*`, `B`, `B*`, `b`, `b*` and
`n` all have arms; **`s` has none**, so it falls through to

```rust
_ => Operator::Other { name: name.to_string(), operands: Box::new(operands.into_vec()) },
```

and every consumer ignores it. §8.5.3.2 defines `s` as "close the current
subpath, then stroke it" — exactly `h S` — and producers emit it wherever they
close a stroked figure.

Two things go wrong, and the second is the damaging one:

1. **the stroke is never painted**, and
2. **the current path is never cleared.** Only a painting operator's arm resets
   the path builder, and an ignored operator has no arm — so the abandoned
   geometry stays in the builder and the **next fill paints it**.

A stroked frame followed by an ordinary white label fill therefore floods the
frame's whole interior white, wiping everything already drawn inside it.

## The two parsers in the crate disagree

`parse_content_stream_paths_only`, the byte-level fast path in the same file,
handles `s` correctly (0.3.77, around line 261):

```rust
// s = close path + stroke (not a named variant, emit ClosePath + Stroke)
b's' => {
    operators.push(Operator::ClosePath);
    Some(Operator::Stroke)
},
```

Rendering goes through `parse_content_stream` → `parse_operator_with_operands`
→ `build_operator`, which is the one that misses it. That split is why the
defect never shows up in the extraction paths and only bites rendering.

There is a second trace of the same gap: `PathExtractor::close_and_stroke()`
(`src/extractors/paths.rs`) exists, is exactly what `s` needs, and is
unreachable — nothing constructs an operator that would call it.

## Reproduction

[pdf-oxide-upstream-close-stroke-not-an-operator-fail.pdf](https://github.com/user-attachments/files/31256088/pdf-oxide-upstream-close-stroke-not-an-operator-fail.pdf)
[pdf-oxide-upstream-close-stroke-not-an-operator-pass.pdf](https://github.com/user-attachments/files/31256089/pdf-oxide-upstream-close-stroke-not-an-operator-pass.pdf)

Two one-page PDFs that differ in a single byte of content stream:

```
2 w 0 0 0 RG
20 20 m 180 20 l 180 180 l 20 180 l 20 20 l {s | h S}
1 1 1 rg
60 90 80 30 re f
0 0 0 rg
90 40 20 20 re f
```

A stroked frame, a white label rectangle inside it, and a black square that is
always painted — so a blank page cannot be mistaken for an empty fixture.

| file | frame terminator |
|---|---|
| `…-fail.pdf` | `s` |
| `…-pass.pdf` | `h S`, which is what `s` is defined to be |

Expected: the two render identically.

Actual, `pdf_oxide 0.3.77`: `…-fail.pdf` loses the frame and its interior is
flooded white by the label fill.

Ink coverage of the full page, same measure over all three renders (sampled
every 4th pixel in each direction, luma < 250, 200 dpi):

| file | reference rasteriser | pdf_oxide 0.3.77 | with the fix below |
|---|---|---|---|
| `…-fail.pdf` | 0.03240 | **0.01014** | 0.04477 |
| `…-pass.pdf` | 0.03240 | 0.04477 | 0.04477 |

The reference renders the two files identically, as the spec requires; 0.3.77
does not. (Both `pass` figures sit above the reference because this renderer
lays a slightly wider anti-aliased stroke — unrelated to the defect, and equal
across the pair.)

## Suggested fix

`build_operator` returns a single `Operator`, so it cannot emit `ClosePath` +
`Stroke` the way the byte parser does. Give `s` its own variant, next to the
`CloseFillStroke` / `CloseFillStrokeEvenOdd` the enum already carries for `b`
and `b*`:

```diff
     /// Stroke path (S)
     Stroke,
+    /// Close current subpath and stroke it (s)
+    CloseStroke,
```

```diff
         "S" => Operator::Stroke,
+        "s" => Operator::CloseStroke,
         "f" | "F" => Operator::Fill,
```

and let the consumers that dispatch on `Stroke` close first, exactly as they
already do for `b` / `b*`:

```diff
-                Operator::Stroke => {
+                Operator::Stroke | Operator::CloseStroke => {
                     ...
+                        if matches!(op, Operator::CloseStroke) {
+                            current_path.close();
+                        }
                         if let Some(path) = current_path.finish() {
```

The sites that need it on 0.3.77:

| file | what it needs |
|---|---|
| `src/rendering/page_renderer.rs` | the paint arm, and `is_paint_operator` (knockout-group element boundaries) |
| `src/rendering/separation_renderer.rs` | the same paint arm |
| `src/document.rs` | both path-extraction dispatchers — `PathExtractor::close_and_stroke()` is already there |
| `src/redaction/serialize.rs` | emit `s\n`, so the content-stream round-trip keeps its current output |
| `src/bin/export_to_markdown.rs` | `s` ends a path the same way `S` does |

## The rest of the operator table

Comparing every literal `build_operator` matches against the ISO 32000-1
operator set, the arms that are missing are:

`s`, `"`, `BI`, `ID`, `EI`, `BX`, `EX`, `MP`, `DP`, `d0`, `d1`

`BI`/`ID`/`EI` are handled separately by `parse_inline_image`; `BX`/`EX`,
`MP`/`DP` and `d0`/`d1` are correctly no-ops for a renderer. That leaves `s` —
this report — and **`"`** (show text with word and character spacing), for which
an `Operator::DoubleQuote` variant exists that the full parser never produces.
Reported separately.

### Actual Behavior

-

_No response_

## Environment

- `pdf_oxide` 0.3.77
- `xberg` 1.0.14
- rendered at 200 dpi; ink is the fraction of pixels with luma < 250, sampled
  every 4th pixel in each direction

### Code Sample

```rust

```

### Error Messages / Logs

```shell

```

### Additional Context

_No response_

### Checklist

- [x] This report is written in my own words (not AI-generated)

AshSgDe29071999:
## Problem

`build_operator` maps every path-painting operator in ISO 32000-1 Table 60 except `s`. That operator fell through to `Other`, so:

1. the stroke was never painted
2. the path builder was not cleared, and the next fill painted the abandoned geometry

The byte-level `parse_content_stream_paths_only` already expanded `s` to `ClosePath` + `Stroke`. The nom parser used by rendering did not.

Fixes #1096.

## Change

Expand `s` into `ClosePath` then `Stroke` in `parse_content_stream`, so existing painting arms reset the path.

## Test

```
cargo test --lib test_parse_close_and_stroke_operator
# ok
```


yfedoseev:
Thanks — the diagnosis is right and the test is the correct shape. ISO 32000-1 Table 60 defines `s` as close-and-stroke, equivalent to `h S`, and today it falls through to `Operator::Other`, leaving the path alive so the next painting operator repaints the abandoned geometry.

Two things before this can merge.

### 1. The expansion reaches only one of the two parsers

`push_parsed_operator` is called from `parse_content_stream` (`src/content/parser.rs:126`).

`parse_content_stream_paths_only` (`src/content/parser.rs:176`) is a second, independent scanner. Its fast opcode table has no `s` arm, so `s` falls through to the generic fallback at `:373`, which pushes the operator directly at `:375` without the expansion. That path still produces `Operator::Other { name: "s" }`.

That parser backs the public `extract_paths` API (`src/document.rs:17818`, and again for form XObjects at `:18410`). Its operator loop ends in a `_ => {}` catch-all, so the close-and-stroke is silently dropped there — the same defect this PR fixes, on a surface callers reach directly. `S` is unaffected, because it has no `build_operator` special case and the fallback already returns `Operator::Stroke` for it.

Routing the `paths_only` fallback through `push_parsed_operator` as well would cover both entry points. A case asserting `parse_content_stream_paths_only` on the same stream would pin it against regression.

### 2. Repo convention: no issue references in source

The new test's comment carries an issue number. Please drop it — issue references belong in the PR description and the changelog, not in code, comments, or test messages.

Nothing else blocking; the parser change itself and the operator ordering in the test both look correct.

yfedoseev (approved):
Approving. This is a correct, minimal fix for a real defect, and it is now backed by rendering evidence rather than reasoning alone.

**The diagnosis is exactly right, including the part that is easy to miss.** `s` falling through to `Other` does not merely skip the stroke — it leaves the path builder uncleared, so the next fill paints the abandoned geometry on top of its own. Your description called that out, and it is the more damaging half.

**Independent confirmation.** Before reading your diff I put the underlying question to a fresh investigation with no knowledge that a PR existed — only "does this library handle the `s` path-painting operator correctly, and what breaks where it does not". It independently reached the same root cause and proposed the same fix, in the same place: expand `Other{"s"}` in the `Ok((rest, op))` arm of `parse_content_stream`, mirroring what the byte-level scanner already did. Arriving at an identical answer with zero exposure to yours is about as good as review evidence gets.

That investigation also widened the picture usefully:

- The parser you fixed is the one **the entire rendering stack** consumes — page content, Type 3 glyphs, Form XObjects, tiling patterns, transparency groups, and the separation renderer. The already-correct scanner has only two callers. So the blast radius was larger than the PR claims.
- `page_renderer.rs:1846-1852` carries a comment asserting the parser decomposes `s` into `ClosePath` + `Stroke`. That was true of the fast scanner and false of the one `execute_operators` actually uses — load-bearing misinformation, and very likely why the arm was never written. **Your change makes that comment true.**
- Three further consumers are repaired transitively, because none of them ever sees `Other{"s"}` again: `is_paint_operator` (knockout-group segmentation), the `in_path` tracking in `bin/export_to_markdown.rs`, and the writer/renderer asymmetry where the writer could emit an `s` the renderer could not draw.
- Poppler (`Gfx::opCloseStroke`), PDFium (`Handle_CloseStrokePath`), MuPDF (`pdf_run_s` with `doclose=1`) and PDF.js (`OPS.closeStroke`) all model this operator explicitly. Omitting it was unique to us.

**Verified by rendering, not just by test.** I rasterised real-world documents before and after. A handful of pages change, and every one of them gets *darker* with equal or greater ink coverage — the exact signature of strokes that were previously never painted now being painted. No page appeared, vanished, or changed size. One page darkens at identical coverage, which is the same mechanism where the recovered stroke lands on already-inked pixels. This matters because a text-extraction check cannot see any of it: your change is on the renderer's parser.

**On the test.** `test_parse_close_and_stroke_operator` uses `b"100 200 m 150 250 l s 0 0 10 10 re f"` — the trailing `re f` covers the compounding case, not just the operator mapping. That is the right test, and it also satisfies the over-collection requirement in CONTRIBUTING §4: your change causes more to be painted, and the fixture shows the abandoned geometry no longer leaks into the following fill.

One note for the record: the alternative is a real `Operator::CloseStroke` variant, which would preserve byte-exact `s` round-tripping instead of re-serialising as `h S`. `Operator` is `pub` and not `#[non_exhaustive]`, so that is a breaking change. Your approach is the right one for a patch release; the enum variant belongs in a release that already breaks that enum, if byte-exact round-trip is ever wanted.

CI on this branch has only ever produced cancelled runs, so it needs one clean cycle before merge. Approving now so that is the only thing outstanding.

PR #1102, opened 2026-08-20
fix(text): keep a sign with the number it belongs to in right-to-left text

### LINKED ISSUE #1101: [Bug]: a percent or currency sign reverses away from its number in right-to-left text
## What happens

In a right-to-left paragraph, a visual-order `50%` comes back as `%50`. The sign leaves the number it belongs to.

## Why

`reverse_rtl_keep_numbers` in `src/text/bidi.rs` counts a number as the digits plus the separators between them. A sign that touches the digits falls outside that run, so it reverses as an ordinary neutral character.

## What the standard says

UAX #9 rule W5 makes a sequence of European terminators next to a European number part of that number. The sign then keeps the side of the digits it was drawn on.

## Scope

This affects right-to-left text that holds a percent sign, a currency sign or a degree sign next to a number. Separators are a different rule and behave correctly today: they join a number only between two digits, so `1,000` and `3.14` stay intact and a trailing comma still reverses.

tobocop2:
## Linked issue

Closes #1101

## What and why

`reverse_rtl_keep_numbers` counted a number as digits plus the separators between them, so a sign touching the digits fell outside the run and reversed away as an ordinary neutral. A visual-order `50%` came back as `%50`.

UAX #9 rule W5 makes a sequence of European terminators adjacent to a European number part of that number, so the sign keeps the side of the digits it was drawn on. The class comes from the `unicode-bidi` tables this module already wraps, not from a second local list of sign characters.

Separators keep the existing rule. They join a number only between two digits, and `-` and `+` are European separators, so `-2009, ` is unchanged.

## Type of change

- [x] Bug fix

## Tests

- [x] Test named by defect **class**, not an issue/PR number

## Regression on real PDFs

**Corpus I tested**: 19,151 documents / 470,207 pages (govdocs plus the pdfium, pdfjs and mutool exotic sets), diffed against `main`.

- [x] Diffed my branch against **`main`** — all 470,207 pages byte-identical, no page moves.
- [ ] Diffed against the **latest release** — not run.
- [x] Judged by per-page output hashes across text, markdown, HTML and table columns.

Read that as a no-regression result and nothing more. The corpus is Latin-script government documents and holds almost no right-to-left text, so it barely exercises the path this touches: the fix is expected to move nothing there, and it moved nothing. The unit tests are what actually pin the behaviour.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).
- [x] Public-API changes considered for semver. No public API changes: `reverse_rtl_keep_numbers` keeps its signature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -170,7 +170,10 @@ encoding_rs = "0.8"                               # Shift-JIS, EUC-JP, and other
 # Unicode bidirectional algorithm (UAX #9) — used by both the HTML+CSS
 # inline formatter (Phase LAYOUT-3) and the markdown converter's RTL
 # emphasis cleanup (#377 D7).
-unicode-bidi = { version = "0.3", default-features = false }
+# `hardcoded-data` carries the bidi class tables `bidi_class` reads. Without
+# it the crate compiles only while some other dependency happens to turn the
+# feature on, so it is declared here rather than relied on by unification.
+unicode-bidi = { version = "0.3", default-features = false, features = ["hardcoded-data"] }
 unicode-normalization = "0.1"                                # NFC composition of overlapping combining marks emitted as separate glyph runs (e.g. "´" + "E" → "É")
 
 # Image processing
--- a/src/text/bidi.rs
+++ b/src/text/bidi.rs
@@ -618,7 +618,10 @@ pub fn wrap_rtl_isolates(text: &str, block_is_rtl: bool) -> String {
 /// therefore not mirrored when the surrounding RTL is reversed. A separator
 /// (`.` `,` `:` and the Arabic decimal/thousands separators) is treated as part
 /// of the number only when it sits between two digits, so `1,000` and `3.14`
-/// stay intact while a trailing comma reverses as an ordinary neutral.
+/// stay intact while a trailing comma reverses as an ordinary neutral. A
+/// European terminator — `%`, a currency sign, a degree sign — joins the number
+/// whenever it touches one (UAX #9 rule W5), so `50%` keeps its sign on the
+/// side the reader expects instead of reversing away from the digits.
 ///
 /// For any run containing no digits this is byte-identical to a plain
 /// `chars().rev().collect()`, so the digit-free RTL case (the corpus-validated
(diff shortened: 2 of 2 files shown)


yfedoseev:
Swept this against our regression corpus: no surface changed, so nothing here is a measured regression. Two things came out of a targeted battery on `reverse_rtl_keep_numbers` that are worth settling before it lands.

### 1. W5 is a logical-order rule; this function takes visual order

The doc comment on the function states its contract:

> Reverse a **visual-order** RTL run to logical order while keeping embedded number sequences in their natural left-to-right order — UAX #9 rule **L2** [...]

L2 is a reordering rule. W5 belongs to the resolution phase, which UAX #9 runs over the **logical** character sequence, before any L-rule reorders for display. Applying W5 to visual-order input makes the terminator immovable in both directions:

```
main     "50%" -> "%50"      "%50" -> "50%"
this PR  "50%" -> "50%"      "%50" -> "%50"
```

Both of the PR's new assertions (`50%` → `50%` and `%50` → `%50`) are identity, so after this change the function can no longer move a sign relative to its digits at all. That is the right answer when the producer already drew the sign on the side it reads on, and the wrong one when it did not — previously the reversal at least relocated it. Which of those is the common case in visual-order runs is the question; it decides whether the gate belongs here or one level up, where logical order is known.

### 2. W4 is untouched, and it is reachable

`-` and `+` are `ES`, not `ET`, and are absent from `is_sep`, so a sign *between* two numbers still inverts:

```
"2009-2010" -> "2010-2009"
"12/2009"   -> "2009/12"
"5+3"       -> "3+5"
```

unchanged by this PR. `document.rs:8574` guards itself with `digit_runs == 1`, but `apply_rtl_verdict` (`bidi.rs:436`, `:440`) calls the function with no such guard, so a date range or fraction in a visual-order run reaches it. Not a regression — worth knowing the fix is partial if the commit message reads as covering signs generally.

### Minor

`50°C` → main `C°50`, this PR `C50°`. Both wrong (`50°C` is correct); noting it only so the `°` case is not assumed covered.

### Packaging

This is the crate's first use of `unicode_bidi::bidi_class`, which is `#[cfg(feature = "hardcoded-data")]`. `Cargo.toml` declares `unicode-bidi` with `default-features = false`, so it currently resolves only through feature unification from `stringprep`:

```
unicode-bidi feature "hardcoded-data"
└── unicode-bidi feature "default"
    └── stringprep v0.1.5
```

`stringprep` is not optional today so nothing breaks, but the feature is relied on without being declared. Adding `features = ["hardcoded-data"]` would make it explicit.

tobocop2:
Feature declared. W4 and `50°C` are real gaps; the commit no longer reads as covering signs generally.

On W5 over visual order: my regression corpus holds no right-to-left text, so I built one — 3,843 Arabic, Hebrew, Persian and Urdu documents, 60,093 pages, Wikipedia exports plus Internet Archive scans for producer spread. The change moves 57 pages, so it is not inert.

Every difference is the sign's side. `main` gives `حي 100%شد`, this gives `حي %100شد`. That is your reading exactly: `main` relocates the sign, this leaves it where the producer drew it. The extracted text cannot say which is right — that needs the rendering.

So the gate is doing what you describe, and the open question stays open. If you would rather it sat one level up where logical order is known, say so and I will move it.

PR #1104, opened 2026-08-20
perf(tables): group a page's text blocks by MCID once

### LINKED ISSUE #1103: [Bug]: assembling a tagged table rescans every text block on the page for each cell MCID
## What happens

`extract_cell` in `src/structure/table_extractor.rs` collects the text blocks a tagged cell owns by scanning every text block on the page, once for each of the cell's MCIDs:

```rust
for mcid in &mcids {
    let sequence_start = runs.len();
    runs.extend(text_blocks.iter().filter(|b| b.mcid == Some(*mcid)));
```

## The cost

The work for one cell is the number of MCIDs the cell owns times the number of text blocks on the page. Summed over a table whose cells cover the page, the MCID count approaches the block count, so assembling the table is quadratic in the content of the page.

A dense tagged table is exactly the shape that makes both numbers large at the same time, so the pages that hold the most table content pay the most.

## Suggested fix

Build the index once for the page and let each cell look its MCIDs up: a map from MCID to the blocks that carry it, then one lookup per MCID. That turns the assembly linear in the blocks on the page.

`extract_table_on_page`, the per-page entry point added in #1092, is a natural place to build it, because it already sees the page once before it walks the rows.

## Scope

This is not a regression. The scan predates the recent table work, which keeps the same structure. Filing it separately so the fix stays independent of the correctness changes in flight.

tobocop2:
> **Stacked on #1092 — merge that one first.**
>
> The index lives in `extract_table_on_page`, the per-page entry point #1092 adds, so this branch carries #1092's commits as its baseline. Once #1092 lands I rebase onto it and this PR reduces to the single commit below.

## Linked issue

Closes #1103

## What and why

A cell names the marked-content sequences it owns, not the text blocks, so assembling a cell had to find its blocks by MCID. It scanned every text block on the page once per name. On a table that covers the page the number of names approaches the number of blocks, so assembling the table cost the content of the page squared.

The page is grouped once and every cell looks its names up. Each group keeps the order the blocks appear in on the page, which is the order the scan returned, so the assembled text does not change.

## Type of change

- [x] Performance

## Tests

No new test. The change is a lookup for a scan that returns the same blocks in the same order, so it has no behaviour to pin that the existing table tests do not already cover.

The invariant is that output does not move at all. I am running the corpus sweep against the parent branch, where the expected result is byte-identical output on every page, and will post the numbers here.

## Regression on real PDFs

**Corpus I tested**: 19,151 documents / 470,207 pages (govdocs plus the pdfium, pdfjs and mutool exotic sets), swept against the parent branch.

- [x] Diffed my branch against its parent — **all 470,207 pages byte-identical**. Grouping the page by MCID hands each cell the same blocks in the same order, so no page may move; a single changed page would have refuted the change.
- [ ] Diffed my branch against the **latest release** — not run; the parent branch is the meaningful baseline for a stacked change.
- [x] Judged by comparing per-page output hashes across text, markdown, HTML and table columns, not char-Levenshtein.

## AI assistance disclosure

      predominantly AI-generated.

## Checklist

      CHANGELOG (not in code).
- [x] Public-API changes considered for semver. No public API changes: the index and its type are private to `table_extractor`.


tobocop2:
## Corpus check

I compared this branch against its parent branch over 19,151 documents and 470,207 pages. All 470,207 pages come back byte-identical: no page gains, loses or rearranges a character.

That is the result this change has to produce. Grouping the page by MCID hands each cell the same blocks the scan handed it, in the same order, so nothing a cell assembles can move. A single changed page would mean the grouping does not preserve order, so the comparison is the proof rather than a formality.

## Tests

9,030 tests pass with the `rendering` feature, no failures.

yfedoseev:
Swept and reviewed this separately from #1092. Two findings, one blocking.

### 1. Blocking: this branch carries the MCID-scope regression from #1092

`split/e3c-mcid-index` is `split/e3b-cell-reading-order` plus one commit:

```
$ git merge-base --is-ancestor <1092-head> <1104-head>   ->  true
commits on 1104 not in 1092:  perf(tables): group a page's text blocks by MCID once
commits on 1092 not in 1104:  (none)
```

Both PRs declare `base=main`, so GitHub presents them as independent, but merging this one lands #1092's reading-order change too — under a `perf` title.

The regression reproduces here identically. MCIDs are scoped to the content stream that emits them (ISO 32000-1:2008 §14.7.4.3), so a page and a Form XObject drawn on it may each use MCID 0. On a hand-built tagged PDF where exactly that happens:

```
```

Full analysis and the reproduction script are on #1092: https://github.com/yfedoseev/pdf_oxide/pull/1092#issuecomment-5364869949

### 2. The index itself measured clean — but it will need the same key

Worth stating plainly, because it is the part unique to this PR and it held up:

- Across our regression corpus (445 documents drawn from open-source sources, compared on text, markdown and HTML), the output of this branch is **byte-identical** to #1092 on all 1336 surfaces. The grouping is pure speed, no behavioural drift.
- The 11 surfaces that move relative to `main` are all text **gains**, zero losses — e.g. two government forms going 5712→6254 and 4002→4440 words. That is recovered cell text from no longer dropping runs that share an MCID, not displacement.
- `HashMap` iteration order cannot leak into output: the map is only ever `.get()`, never iterated.
- `mcids.retain(|m| seen.insert(*m))` correctly closes the repeated-MCID double-emission that the previous `break` masked.

The one thing to carry over: the index is keyed by MCID alone,

```rust
type BlocksByMcid<'a> = HashMap<u32, Vec<&'a TextBlock>>;
```

so once #1092 keys cell lookup by `(scope, mcid)`, this map needs the same key or it reintroduces the collision at the grouping layer. `TextBlock::mcid_scope` and `StructChild::MarkedContentRef::scope` both already carry it, and the structure data model is documented as keyed by `(McidScope, mcid)`.

No objection to the approach — `O(B + M)` per table instead of rescanning the page per MCID is the right fix, and an independent read of the same code arrived at the same insertion point and the same three helper signatures.

tobocop2:
Fixed in #1063; rebased onto it. The index stays keyed by the number with the stream checked on the bucket, since one number holds few blocks.

On the stacking, you are right that merging this lands #1092 too. I would rather it waited for #1092 than carry it under a `perf` title — happy to retarget the base if that is easier for you.

yfedoseev:
Same five compile errors as #1092, which this contains — see https://github.com/yfedoseev/pdf_oxide/pull/1063#issuecomment-5365931824

Agreed on the stacking: better to wait for #1092 than carry it under a `perf` title. Retargeting the base would help.

tobocop2:
Fixed and verified at `253802f8`: `cargo check --all-targets` clean, 9031 passed / 0 failed. Retargeted onto #1092's branch so this no longer carries its commits standalone.

tobocop2:
Correction to my line above: I rebased the branch onto #1092 but did not retarget the base, and I can't — GitHub needs the base branch to exist in your repo and `split/e3b-cell-reading-order` only exists on my fork. Base is still `main`. Simplest is to hold this until #1092 merges; if you'd rather retarget, push that branch upstream and I'll switch the base.

tobocop2:
Rebased onto current `main` and re-measured against #1092 alone: all 470207 pages byte-identical, 0 blockers, `cargo check --all-targets` clean, 9041 passed / 0 failed. Grouping the page by MCID hands each cell the same blocks in the same order, so byte-identical is the result this has to produce.

Issue #1498, opened 2026-09-17
Spatial table detector returns no tables for a line-less grid from 0.3.71 onward


bbcriston:
**Regressed in:** 0.3.71 (0.3.70 is the last good version)
**Still present in:** 0.3.78

## What happens

A 26-row x 6-column table on a page with **no ruling lines** is detected at
0.3.70 and not detected at 0.3.71 or later. `extract_tables` returns an empty
`Vec` for that page — and, on this document, for all 49 pages.

```rust
let doc = PdfDocument::open("49-pages-3-exported.pdf")?;
let tables = doc.extract_tables(24)?;
// 0.3.70: 1 table, rows = 26, first row has 6 cells, is_real_grid() == true
// 0.3.71+: 0 tables
```

## The detector's inputs are unchanged

Probed on the same page, same file:

| version | words | paths | ruling lines | tables |
|---|---:|---:|---:|---:|
| 0.3.70 | 288 | 3 | 0 | 1 |
| 0.3.71 | 289 | 3 | 0 | 0 |

So this is not word segmentation or path extraction — the spans reaching the
detector are effectively identical, and the table is lost inside it. With zero
ruling lines the page goes through the text-only fallback
(`detect_tables_from_spans_column_aware`).

## Not recoverable by configuration

`extract_tables_with_config` at 0.3.78 returns zero tables for every one of
these, document-wide:

- `TableDetectionConfig::default()`
- `regular_row_ratio: 0.05`
- `min_table_cells: 2` + `regular_row_ratio: 0.05`
- `max_table_columns: 40`, `column_tolerance: 25.0` + `regular_row_ratio: 0.05`
- `row_tolerance: 6.0` + `regular_row_ratio: 0.05`

`TableDetectionConfig`'s fields and `Default` are byte-identical between 0.3.70
and 0.3.78, so the change is in the detector body, not its knobs.

## Suspected area

The 0.3.70 -> 0.3.71 diff of `structure/spatial_table_detector.rs` is mostly
additive work on the *ruled* path (`detect_header_row_above`, protected header
rows for colspan merging). The one change that also affects a line-less page is
the fragment consolidation ahead of the minimum-row filter, which moved from a
fixed vertical tolerance to one scaled by median fragment row height:

```rust
let row_h = median_fragment_row_height(&final_tables);
let y_tol = (row_h * 1.5).max(3.0);
[...]

PR #1499, opened 2026-09-17
fix(tables): group a nested rule with the rule that contains it

bbcriston:
Refs #1498.

**Disclosure up front: this change was written by an AI coding agent
(`Assisted-by: claude-code`), acting on my instruction. I am putting my name and
sign-off on it, and I am posting it as a draft rather than for merge. If that
puts it outside what you accept under CONTRIBUTING.md, close it — no hard
feelings, and the diagnosis below still stands on its own.**

It is also **not a complete fix**. It addresses one of two causes and does not
by itself restore the table. Details below.

### What #1498 turns out to be

Two independent changes in 0.3.71, stacked.

**1. Rule families (what this PR changes).** `x_coherent_rule_families` scores
pairs by `overlap / union` against `X_COHERENCE = 0.85`. The page has two
page-wide rules at x 41.2..378.6 and one interior rule at x 42.6..320.2. The
interior rule is fully contained in the outer, but scores

```
overlap 277.6 / union 337.4 = 0.822
```

and is split into its own family. Neither family then has the two rules the
adjacent-pair walk needs, so the region holding the table is never formed and
`detect_tables_from_horizontal_rules` returns nothing.

The patch unions a pair when the narrower rule lies wholly inside the wider one
and still covers most of it. Containment is a stronger signal than the ratio
measures — a contained rule has no non-overlapping part of its own, so every
point it loses comes from the wider rule sticking out past it. The
partial-overlap path the 0.85 bar exists for is untouched, and there are tests
for both of those cases.

**2. Word x-geometry (not addressed here).** With the region restored, column
detection finds **9 columns where v0.3.70 found 8**, and
`validate_table_structure_internal` then rejects on row regularity. The column
boundaries have all shifted — one column at 85..137 in v0.3.70 is now two, at
88..120 and 122..144 — which looks like the per-glyph coordinate work in 0.3.71
("stamp accurate x from `extract_chars`"). Whole-page word count went 288 → 289.

I am not guessing at that second one. The new coordinates may well be the more
correct ones, in which case the question is whether the regularity gate should
tolerate the finer split, and that is a call on your corpus, not mine.

**So: page 24 still extracts 0 tables with this patch applied.**

### Testing

- The three grouping tests are in the commit *before* the fix. The nested case
  fails there (`families came out as [[107.1], [544.6, 50.1]]`) and passes after.
- `cargo test --lib spatial_table_detector::tests` — 94 pass.
- Table suites pass: `test_tables` (60), `test_booktabs_table_detection` (12),
  `test_spatial_table_detection`, `test_pmc_real_document_tables`,
  `test_rule_bounded_row_group_is_a_table_even_where_its_cells_are_not_boxed`,
  `test_ruled_table_emits_each_cell_once`, `bulleted_list_not_table`,
  `cjk_prose_not_table` — 82 tests across the 8 files.
- **No `corpus_sig` sweep has been run.** I do not have a corpus assembled that
  is representative enough to make the claim rule 2 asks for, so I am not
  claiming it. That is the main reason this is a draft.

The test geometry is the arithmetic of the failing case written out, not a
capture of any document; no fixture is added.


bbcriston:

PR #1502, opened 2026-09-18
fix(cmap): parse all bfrange/notdefrange triples on a single concatenated line (#1497)

### LINKED ISSUE #1497: Type0/Identity-H CJK text extracted as PUA code points (ToUnicode not applied for vendor CID orderings e.g. 'Founder PKU2')
# Type0/Identity-H CJK text is extracted as Private Use Area (PUA) code points instead of real Unicode

## Summary
When a PDF uses `Type0` / `Identity-H` fonts whose `CIDSystemInfo` ordering is a
vendor-specific collection (e.g. `Founder PKU2`, `Founder PKUO1`, `Founder PKUF1`,
`Founder PKUE1`) rather than the standard `Adobe-GB1`, `pdf_oxide` returns the
page text as **Private Use Area (PUA) code points (U+E000–U+F8FF)** for nearly
every glyph, instead of the real CJK Unicode characters that the embedded
`ToUnicode` CMap maps them to.

Because the codepoints are PUA, any downstream consumer that sanity-checks the
text (e.g. "is this page garbled?") concludes the page has no usable text layer
and discards it, falling back to full-page OCR — roughly 50× the compute of just
trusting the embedded text.

## Affected version
`pdf_oxide` **0.3.73** (the version pinned by RAGFlow's `build.sh`; the
`pdf_oxide-go-ffi-linux-amd64` prebuilt static lib).

## Reproduction
A zipped reproduction PDF is attached (`repro.zip`) — a Chinese legal-code
document (《中华人民共和国民法典》) produced with Founder/方正 fonts using the
`Founder PKU2` CID collection. Any PDF with `Type0`/`Identity-H` + a
non-`Adobe-*` `CIDSystemInfo` ordering reproduces this; the issue is not specific
to this file.

Steps to observe:
1. Open the attached PDF with `pdf_oxide` (Rust API).
2. Extract the text / per-glyph chars of any content page.
3. Inspect the codepoints of the returned characters.

```text
page 1: 318 chars, '#'=0, PUA/replacement=300   (94% PUA)
page 2: 112 chars, '#'=0, PUA/replacement=106
page 9: 498 chars, '#'=0, PUA/replacement=464   (93% PUA)
```
i.e. ~70–95% of glyphs come back as U+E000–U+F8FF instead of Han characters.

## Why this is a pdf_oxide bug (not a bad PDF)
The same file, opened with `pdfplumber` / `pdfminer.six`, yields perfectly clean
Chinese text:

```text
中
目 录
华
人
民
共
和
国
中华人民共和国主席令（第四十五号）…………………（１）
...
```

`pdfplumber` applies the **embedded `ToUnicode` CMap** correctly, so the CIDs
resolve to real Unicode. Since `pdfminer` gets correct text from the *same*
`ToUnicode` stream, that stream itself is valid — `pdf_oxide` is simply not
applying it for these fonts, and falls back to a PUA placeholder per glyph.

## Root-cause hint
The font `BaseFont` is a subset like `/SSJ-PK748200004ab` with
`CIDSystemInfo = { Registry: "Founder", Ordering: "PKU2" }`. This is a
vendor-private character collection, so an external `Adobe-GB1`-style CMap cannot
be used as a fallback; the **only** way to get real Unicode is the embedded
`ToUnicode` CMap. Applying that CMap (the way `pdfminer.six` does) should produce
clean CJK text.

## Suggested fix
In the text-extraction path, when resolving `Type0`/`Identity-H` glyph codes to
Unicode, prefer the font's embedded `ToUnicode` CMap (and handle `bfrange` /
`bfchar` / multi-byte forms) before any PUA fallback. This should let
vendor-CID collections such as `Founder PKU*` resolve to real characters.

Happy to help test a fix or provide more sample files.

xugangqiang:
## Summary

Fixes #1497: Type0 / Identity-H CJK text (e.g. Founder/PKU2 subset fonts) was extracted as Private Use Area code points (U+E000–U+F8FF) instead of real Unicode, which made downstream consumers discard the text layer and fall back to full-page OCR.

## Root cause

Some PDF producers emit the entire `beginbfrange` / `beginnotdefrange` block as a **single newline-free line** with every `<src> <end> <dst>` triple concatenated (e.g. `100 beginbfrange` with all triples on one line). The previous `parse_bfrange_line` / `parse_notdefrange_line` used `RE.captures(line)` — which matches only the **first** triple — so every mapping after the first was silently dropped. For CID fonts with a private ordering (Founder/PKU*), where the embedded ToUnicode CMap is the only way to recover Unicode, the dropped entries resolved to U+FFFD, garbling Chinese (previously ~70–95% PUA on affected pages).

## Fix

- `parse_bfrange_line` now iterates **every** triple via `captures_iter`: it collects each `bfrange` array's `(start,end)` span, masks those spans out, then iterates the remaining sequential triples — so mixed array + sequential triples on one line are both handled.
- `parse_notdefrange_line` accumulates every `captures_iter` triple instead of only the first.
- New `decode_bfrange_dst` helper decodes one destination token (BMP, supplementary, UTF-16 surrogate pair, or multi-char ligature) and returns `None` for an unmappable token so only that entry is skipped.
- `saturating_sub` guards the array branch against subtract-with-overflow on malformed input.

## Tests

- Unit: `decode_bfrange_dst` covers BMP/supplementary/surrogate/ligature and the unmappable-token skip; an array with a malformed dst keeps its valid neighbours.
- End-to-end (`tests/test_bfrange_multitriple_line_extraction.rs`): concatenated 1:1 triples, real ranges, supplementary/surrogate destinations, mixed array+sequential, a normal multiline control, and a faithful `Founder/PKU2` vendor-ordering reproduction.

Verified end-to-end: a Founder/PKU-subset Chinese PDF that previously extracted ~94% U+FFFD via `extract_chars` now extracts 0% U+FFFD. The new end-to-end tests fail on the pre-fix parent and pass with the fix, so they are genuine regression guards.

🤖 Generated with [CodeBuddy](https://www.codebuddy.ai)


xugangqiang:

xugangqiang:
Hi maintainer,

This PR fixes #1497 (Type0/Identity-H CJK text — e.g. Founder/PKU2 subset fonts — extracted as PUA because some producers emit the whole `beginbfrange` block as a single newline-free line, so only the first triple was parsed).

The CI gate `Linked issue is approved and assigned to you` is currently failing. Could you please move issue #1497 through the intake approval and assign it to me (`xugangqiang`)? Once the linked issue is approved + assigned, that check should flip to green and the PR will be unblocked for review/merge.

For context, the regression tests in this PR were verified to fail on the pre-fix parent and pass with the fix (the `New test fails without the fix` check confirms this). CLA is signed.

Thanks!

xugangqiang:
@yfedoseev — quick coordination question on the release timeline.

You mentioned the fix landed on `release/v0.3.79` (milestone v0.3.79). Do you have a rough ETA for when v0.3.79 — or whichever release carries this fix — will be published / merged to `main`?

It affects how I should handle PR #1502:
- If v0.3.79 is coming soon, I can hold or close #1502 and let your in-tree fix carry it (my fixture is still happy to fold in as a second regression guard, as you offered).
- If it's further out, keeping #1502 (with my fix) open gives users a working fix on the current `main` sooner, since I verified `main` (v0.3.78) still drops the non-first `bfrange` triples (6/7 of my regression tests fail there).

Either way, the only remaining CI blocker on #1502 is the intake gate — a `/approve` + assign on #1497 flips it green whenever you want it mergeable. No rush; I just want to align on the timeline so I don't step on your release work.

Thanks!

yfedoseev:
@xugangqiang you still have time, I can't say exactly when 0.3.79 will be released, but I would expect around a week

